PR #2, opened 2020-06-11
feat: Add `caret` icons

jletey:
localhost_3000__query=caret (1)

Related: https://github.com/feathericons/feather/issues/600, https://github.com/feathericons/feather/pull/640

### DIFF
--- a/icons/caret-down.svg
+++ b/icons/caret-down.svg
@@ -1,0 +1,13 @@
+<svg
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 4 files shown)


locness3:
Not sure where this would be used

jletey:
> Not sure where this would be used

@locness3 Good point 🤔 Pretty much all of the use-cases that come to mind rn you can use the `chevron` icons 🤷‍♂️

ericfennis:
Nice I like them.
But @locness3 has a point, are those icons not the same use cases as the `chevron` icons ?

jletey:
Fair, I agree with you both - I'll go ahead and close this 😄

PR #3, opened 2020-06-11
feat: Add `sort` icons

jletey:
localhost_3000__query=sort

Related: https://github.com/feathericons/feather/issues/420, https://github.com/feathericons/feather/issues/202, https://github.com/feathericons/feather/issues/203, https://github.com/feathericons/feather/pull/668


locness3:
I think we should only keep with the sort-asc-2 and sort-desc-2 tho

jletey:
> I think we should only keep with the sort-asc-2 and sort-desc-2 tho

@locness3 Done! I'm not a huge fan of the `sort` icon, should I remove it?

locness3:
At a very small size it looks kinda weird

jletey:
I agree, I have removed it :)

locness3:
Well, nice, guess we can merge

PR #1, opened 2020-06-09
feat: Site

johnletey:
| Light Mode | Dark Mode |
| --- | --- |
| Screenshot 2020-06-10 at 11 42 03 | Screenshot 2020-06-10 at 11 42 08 |
| Screenshot 2020-06-10 at 11 42 16 | Screenshot 2020-06-10 at 11 42 24 |

A new site for this Feather Icons fork. Here are some highlights:

- Pull icon data from the `icons` directory.
  - Site code in main repository.
- Updated UI.
- Simplify site: Use [Next.js](https://nextjs.org/) with TypeScript.

PR #4, opened 2020-06-13
feat: Add `react` package

jletey:
_Based off of [`react-feather`](https://github.com/feathericons/react-feather)_

Icons are pulled from the main [`./icons`](https://github.com/jletey/feather/tree/master/icons) directory (this is where we differ from `react-feather`, which is dependent on the published JS library), and transformed into functions to be published.

I've also added Yarn Workspaces, which simplifies everything in general, as well as allows us to deploy #1.

I've moved the JS library code to a [folder](https://github.com/jletey/feather/tree/feat/react/packages/js). I'll refactor this in a later PR :)


locness3:
I'm not good with React and stuff, so merge if you wish. 👍

jletey:
> so merge if you wish

I'm gonna go ahead and do that 👍 There's probably some things missing for deploying the package, but that's for the future once we think of a name 😄

PR #12, opened 2020-06-17
refactor: Update license

jletey:
Whoops, we need to update the license 😅

Let me know what you think. I think the ISC license is better than the MIT license, but that's just my opinion.

[Rendered `LICENSE` file](https://github.com/featherity/featherity/blob/refactor/license/LICENSE)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,15 @@
-The MIT License (MIT)
+ISC License
 
-Copyright (c) 2013-2017 Cole Bemis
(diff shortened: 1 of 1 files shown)


locness3:
Erm, i don't see any difference

eonist:
@jletey Whatsup with the weird licence of lucide? from the terms: "A permissive license lets people do anything with your code with proper attribution". <- this line opens up for future lawsuits. Revert to MIT?

PR #18, opened 2020-06-23
feat: Add `contact` icon

lscheibel:
Hey :wave: was one project away of forking the repo myself. Thanks for doing this :)

Screenshot:

image

### DIFF
--- a/icons/contact.svg
+++ b/icons/contact.svg
@@ -1,0 +1,17 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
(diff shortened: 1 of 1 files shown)


locness3:
Unrelated, but I think the user icons might need a refresh :)

jletey:
@locness3 I completely agree, I'll open an issue with some thoughts 😁

jletey:
@locness3 I completely agree, I'll open an issue with some thoughts 😁

locness3:
yup

PR #19, opened 2020-06-23
feat: Add `alarm-clock` icon

lscheibel:
Screenshot:

image

### DIFF
--- a/icons/alarm-clock.svg
+++ b/icons/alarm-clock.svg
@@ -1,0 +1,18 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="12" cy="13" r="8"/>
+  <path d="M12 9v4l2 2"/>
+  <path d="M5 3L2 6"/>
+  <path d="M22 6l-3-3"/>
+  <path d="M6 19l-2 2"/>
+  <path d="M18 19l2 2"/>
+</svg>
--- a/icons/alarm-clock.svg
+++ b/icons/alarm-clock.svg
@@ -9,10 +9,10 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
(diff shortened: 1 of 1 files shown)


ericfennis:
Nice work @lscheibel !
How does it look if the feet are 45 degrees on the clock?

lscheibel:
Yeah you're right, the legs are way too straight. Here are some experiments:

100%
image

Closeup
image


For me it's a tie between B and D :sweat_smile:

ericfennis:
I agree. B or D on al large scale.
Personally I like the D variant, I think it fits better if you look to the "100%" view.

dgdavid:
I vote for C and D

lscheibel:
Alright, D it is :smile:

lscheibel:
New Screenshot
image

icons/alarm-clock.svg
```diff
@@ -0,0 +1,18 @@
...
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="12" cy="13" r="8"/>
+  <path d="M12 9v4l2 2"/>
+  <path d="M5 3L2 6"/>
+  <path d="M22 6l-3-3"/>
+  <path d="M6 19l-2 2"/>
+  <path d="M18 19l2 2"/>
```
jletey, line 17: Ah yeah I'm being that guy rn 👀 But if you could add some spacing that'd be great 🤣 Then we can merge this 🚀

```suggestion
  <circle cx="12" cy="13" r="8" />
  <path d="M12 9v4l2 2" />
  <path d="M5 3L2 6" />
  <path d="M22 6l-3-3" />
  <path d="M6 19l-2 2" />
  <path d="M18 19l2 2" />
```

jletey:
Ah so I found out that I can make those changes myself 😂 Anyways thanks for the great work @lscheibel 💪

lscheibel:
Haha no you're absolutely right, gonna be more careful in the future. The feather repo actually has a precommit hook for that. Don't know if I'm a fan of that but we could try that 😅

jletey:
Yeah I'm trying to refactor the build scripts which formatted the code - and let me tell you it's going so amazingly well 🙃

PR #31, opened 2020-06-27
feat: Add `switch-camera` icon

lscheibel:
Screenshot

image

https://github.com/feathericons/feather/issues/97

### DIFF
--- a/icons/switch-camera.svg
+++ b/icons/switch-camera.svg
@@ -1,0 +1,17 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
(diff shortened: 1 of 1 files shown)


ericfennis:
Maybe we can refactor the current `camera` icon with this "rectangle" body to give the camera 1 px safe zone (Current camera doesn't have the 1 px safe zone).

lscheibel:
image


Would submit PR with this one?

jletey:
@lscheibel That's perfect! Go for it!

ericfennis:
I think this is ready to merge right?

PR #39, opened 2020-06-28
refactor: Update `camera` and `camera-off` icons

lscheibel:
https://github.com/featherity/featherity/pull/31

image

right old, left new


jletey:
@lscheibel Also update the "slash" for `camera-off` so that's it's not touching the corners of the icons 😄

lscheibel:
I actually decided against that, for the sake of consistency. I think if we're gonna do it here, we'd have to do it everywhere, right? 😅

ericfennis:
> I actually decided against that, for the sake of consistency. I think if we're gonna do it here, we'd have to do it everywhere, right? 😅

Yeah I think we need to do it everywhere. To keep consistency in the 1px safezone

jletey:
@ericfennis I agree, we need to update all the icons with the slash.

lscheibel:
I would do that in another ticket though and leave these consistent with the others for now. I actually think that it's rather difficult to make all the lines have a 1px safe zone, since the icons they're on top also have to conform to that restriction.

jletey:
@lscheibel When adding the new unlink icons, @ericfennis made the slash have a 1px padding. I think you should do the same here in your PR.

I do agree that it would be hard to do if the icons themselves don't have a 1px padding. But in the future, I am planning on overhauling **all** icons that don't have this.

This will be one of the guidelines for all future icons to Featherity. This wasn't exactly enforced in Feather as much as it should've IMO.

lscheibel:
Okay, maybe we can make the slashes have the padding but I just wanna say, that I think the rule of same "visual-volume" is much more important than the safe-zone and I think we should be _very_ careful, when touching old icons to not remove the "feather"-look. Honestly after shrinking the camera I'm not sure that was the right move either but I'll trust you more than me 😄

An Icon that I think for example doesn't really fit the feather aesthetic is for example the "target" icon. Imo it is way to dense.

lscheibel:
Updated icon
image

jletey:
@lscheibel Thanks for updating the icon. I completely agree that we have to be careful not to lose that signature "feather" look!

I'm completely new at this tbh, @ericfennis do you have any thoughts?

ericfennis:
I think the feather look is still there, I think most of the people won't see it that this icon is updated.
But I agree some icon's can be tricky to fully implement the 1px safe-zone(For example the `cog` icon). 
Mostly all the icon's have the 1px safezone.
But I agree, keeping that feather look is important!

jletey:
Well I think we can maybe go ahead and merge this?

lscheibel:
I guess I'm just a little worried, that we're gonna create more icons that are too dense or overly complicated. I've found some of my icons that I did now, also fall in that "trap", so I might be a little overly eager trying to spot these things 😅  Anyway I'll trust you and I think we should merge these yes

PR #44, opened 2020-07-01
refactor: Update `phone` icons

ericfennis:
I added the 1px safezone for the phone icons.
I think the icon is "more correct" If you look the aligning with the phone.

The diagonal arrow are now also squared. Those match now with the arrows in the new `screen-sharing` icons.

image


ericfennis:
ignore the blue border, it's the mouse in figma :P

ericfennis:
@jletey I agree, thats why I think the diagonal arrows are looking better now. The arrow to right is now actually the same size as the diagonal ones, maybe it looks like it doesn't. 

Do you have an idea how I can improve those?

jletey:
Nope @ericfennis Was just a thought, I feel like we can ahead and merge these then.

PR #46, opened 2020-07-02
refactor: Update `git-commit` icon

jletey:
Screenshot 2020-07-02 at 09 41 12

- Added padding of 2px (same as the other git icons).
- Made the circle the same size as the other git icons.


jletey:
@locness3 Haha 😂 But still, we love hearing your opinion! And you have good comments and suggestions!

PR #49, opened 2020-07-05
feat: 'move-horizontal', 'move-vertical', 'move-diagonal', 'mouse-pointer-2'  

ericfennis:
image

### DIFF
--- a/icons/mouse-pointer-2.svg
+++ b/icons/mouse-pointer-2.svg
@@ -1,0 +1,13 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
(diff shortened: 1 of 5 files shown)


jletey:
I agree with @locness3 here, but it does look like the Figma cursor 🤔

ericfennis:
@jletey haha true. Figma uses this icon design to. It's actually a common cursor/mouse icon design in graphic design programs. Adobe Illustrator and After effects uses it also as the "selection-tool"

ericfennis:
I adjusted the move diagonal icons, the length of the line is now the same as the normal move icons.
image

locness3 (approved):
I'm not a designer

PR #56, opened 2020-07-24
feat: Add equal icons and refactor devide icon

ericfennis:
Adds two new icons:
- Equal
- Equal-off

I Adjusted also the devide icon, because it looked weird.

image


johnletey (approved):
@ericfennis LGTM

I'm not too certain about the `equal-off` icon. I would think of it more like this:

Screenshot 2020-09-03 at 12 32 35

ericfennis:
@johnletey initial that was my idea to make. But it fitted not well with the other icons.
image

johnletey:
Yeah I agree with that @ericfennis, looks a bit odd 😅

ericfennis:
@locness3 So you mean flip the icon horizontally?

locness3:
Basically yes

ericfennis:
I can't get it right with the slash, maybe just without the spacing? Or what do you think, top or bottom?
image

ericfennis:
According to the discord thread, we have chosen to implement the centered version. I also renamed the icon to `equal-not`, because thats name of the symbol.
image

PR #65, opened 2020-09-03
feat: Add clover icon

ericfennis:
image


delnyn:
Uhm... Now the identity has changed, can we close this?

ericfennis:
I think this icon can also be used to describe "luck"

PR #68, opened 2020-09-05
Workflow that will compile font files

yanquisalexander:
Hi :wave:!

I created a workflow that compiles the Featherity font in various formats, including TTF, Woff2, SVG Spritesheet and CSS for web use 🤗 

I have tried it and there is no apparent problem 😄

---

https://github.com/featherity/featherity/issues/54

### DIFF
--- a/.github/workflows/font.yml
+++ b/.github/workflows/font.yml
@@ -17,7 +17,6 @@ jobs:
       - name: Clone sfnt2woff-zopfli repo
         run: git clone https://github.com/bramstein/sfnt2woff-zopfli.git sfnt2woff-zopfli
           
-```suggestion
      - name: Compile program
         run: https://github.com/bramstein/sfnt2woff-zopfli.git
         with:
--- a/.github/workflows/font.yml
+++ b/.github/workflows/font.yml
@@ -28,8 +28,10 @@ jobs:
                 path: sfnt2woff-zopfli
           
           git clone --recursive https://github.com/google/woff2.git
-          cd woff2
-          sudo make clean all
+           - name: Compile woff2
+              run: sudo make clean all
+              with:
+                 path: woff2
            - name: Move and install dependency 
               run: sudo mv woff2_compress /usr/local/bin/ && sudo mv woff2_decompress /usr/local/bin/
               with:
--- a/.github/workflows/font.yml
+++ b/.github/workflows/font.yml
@@ -22,35 +22,36 @@ jobs:
         with:
             path: sfnt2woff-zopfli
       - name: Move and install dependency
-          run: sudo mv sfnt2woff-zopfli /usr/local/bin/sfnt2woff
-          with:
+        run: sudo mv sfnt2woff-zopfli /usr/local/bin/sfnt2woff
+        with:
                 path: sfnt2woff-zopfli
           
 
-           - name: Compile woff2
-              run: sudo make clean all && git clone --recursive https://github.com/google/woff2.git
-              with:
+       - name: Compile woff2
+         run: sudo make clean all && git clone --recursive https://github.com/google/woff2.git
+         with:
                  path: woff2
-           - name: Move and install dependency 
-              run: sudo mv woff2_compress /usr/local/bin/ && sudo mv woff2_decompress /usr/local/bin/
-              with:
+           
+       - name: Move and install dependency 
+         run: sudo mv woff2_compress /usr/local/bin/ && sudo mv woff2_decompress /usr/local/bin/
+         with:
                  path: woff2
           
-     - name: Install Font Custom dependency
-        run: sudo gem install fontcustom
+       - name: Install Font Custom dependency
+         run: sudo gem install fontcustom
           
           
-      - name: Build 'Featherity'
-        run: echo "Building Featherity font" && fontcustom compile ./icons -h -n Featherity -o build -F
+        - name: Build 'Featherity'
+          run: echo "Building Featherity font" && fontcustom compile ./icons -h -n Featherity -o build -F
         
         
-      - name: Zipping 'Featherity'
-        run: zip -r Featherity.zip build
+        - name: Zipping 'Featherity'
+          run: zip -r Featherity.zip build
         
         
-      - name: 'Upload to Artifacts'
-        uses: actions/upload-artifact@v1.0.0
-        with:
+        - name: 'Upload to Artifacts'
+          uses: actions/upload-artifact@v1.0.0
+          with:
           name: Featherity
           path: build
      
--- a/.github/workflows/font.yml
+++ b/.github/workflows/font.yml
@@ -10,48 +10,55 @@ jobs:
   Build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+    - uses: actions/checkout@v2
 
-      - name: Install Dependencies
-        run: sudo apt-get install zlib1g-dev fontforge woff2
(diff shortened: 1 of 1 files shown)


.github/workflows/font.yml
```diff
@@ -0,0 +1,43 @@
...
+          git clone https://github.com/bramstein/sfnt2woff-zopfli.git sfnt2woff-zopfli > /dev/null 2>&1
+          cd sfnt2woff-zopfli
+          make > /dev/null 2>&1
+          sudo mv sfnt2woff-zopfli /usr/local/bin/sfnt2woff
+          git clone --recursive https://github.com/google/woff2.git > /dev/null 2>&1
+          cd woff2
+          sudo make clean all > /dev/null 2>&1
+          sudo mv woff2_compress /usr/local/bin/ && sudo mv woff2_decompress /usr/local/bin/
+          sudo gem install fontcustom > /dev/null 2>&1
+          cd /home/runner/work/featherity/featherity
```
ericfennis, line 28: Is this not better to split those commands up in separate commands?
In case if one of the commands fail, we can directly see where goes wrong.

ericfennis (commented):
Hi @yanquisalexander,

Awesome work, looks really good!
What's the exact output of the zip file?

.github/workflows/font.yml
yanquisalexander: > Hi @yanquisalexander,
> 
> Awesome work, looks really good!
> What's the exact output of the zip file?

- Featherity.css
- Featherity.eot
- Featherity.svg
- Featherity.ttf
- Featherity.woff
- Featherity.woff2
- Featherity-preview.html

---

> Is this not better to split those commands up in separate steps?
> In case if one of the commands fail, we can directly see where goes wrong.


I will edit to show the logs, anyway if something fails the execution stops automatically and shows the error logs

.github/workflows/font.yml
```diff
@@ -0,0 +1,43 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Install 'FontCustom'
+        run: |
+          echo "Dependencies will be installed... Wait a moment"
+          sudo apt-get install zlib1g-dev fontforge woff2
+          git clone https://github.com/bramstein/sfnt2woff-zopfli.git sfnt2woff-zopfli
+          cd sfnt2woff-zopfli
+          make
+          sudo mv sfnt2woff-zopfli /usr/local/bin/sfnt2woff
```
ericfennis, line 22:
```suggestion
      - name: Move and install dependency
            run: sudo mv sfnt2woff-zopfli /usr/local/bin/sfnt2woff
            with:
                path: sfnt2woff-zopfli
          
```

.github/workflows/font.yml
```diff
@@ -0,0 +1,43 @@
...
+          git clone --recursive https://github.com/google/woff2.git
```
ericfennis, line 23: - name: Cloning Woff2 repo
        run: git clone --recursive https://github.com/google/woff2.git

.github/workflows/font.yml
```diff
@@ -0,0 +1,43 @@
...
+          cd woff2
+          sudo make clean all
```
ericfennis, line 25:
```suggestion
           - name: Compile woff2
              run: sudo make clean all
              with:
                 path: woff2
```

.github/workflows/font.yml
```diff
@@ -0,0 +1,43 @@
...
+          sudo mv woff2_compress /usr/local/bin/ && sudo mv woff2_decompress /usr/local/bin/
```
ericfennis, line 26:
```suggestion
           - name: Move and install dependency 
              run: sudo mv woff2_compress /usr/local/bin/ && sudo mv woff2_decompress /usr/local/bin/
              with:
                 path: woff2
          
```

.github/workflows/font.yml
```diff
@@ -0,0 +1,43 @@
...
+          sudo gem install fontcustom
```
ericfennis, line 27:
```suggestion
     - name: Install Font Custom dependency
        run: sudo gem install fontcustom
          
```

.github/workflows/font.yml
```diff
@@ -0,0 +1,43 @@
...
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  Build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
ericfennis, line 13: I think we can move this step after installing the dependencies, right before building the featherity font
```suggestion
      - uses: actions/checkout@v2
```

.github/workflows/font.yml
```diff
@@ -0,0 +1,43 @@
...
+          git clone https://github.com/bramstein/sfnt2woff-zopfli.git sfnt2woff-zopfli
+          cd sfnt2woff-zopfli
+          make
+          sudo mv sfnt2woff-zopfli /usr/local/bin/sfnt2woff
+          git clone --recursive https://github.com/google/woff2.git
+          cd woff2
+          sudo make clean all
+          sudo mv woff2_compress /usr/local/bin/ && sudo mv woff2_decompress /usr/local/bin/
+          sudo gem install fontcustom
+          cd /home/runner/work/featherity/featherity
```
ericfennis, line 28:
```suggestion

```

ericfennis (changes requested):
I meant more like this. 
For example if compiling woff2 fails, we get an email from github actions that it failed at that step.

yanquisalexander (commented):
I think this could add unnecessary extra time to compilation, besides as I said, on failure GitHub Actions will stop everything and show an error log

However I will accept the suggested changes

yanquisalexander:
I accepted the suggestions and the YAML got corrupted, I already fixed it 😅

yanquisalexander (commented):
:)

ericfennis (approved):
Awesome work @yanquisalexander! 
Curious how this work in action 🚀.

fabricedesre:
Is there a way to get the link to the artifact created for a given commit?

ericfennis:
@fabricedesre here is the [link](https://github.com/featherity/featherity/suites/1221644054/artifacts/18286530).
I didn't test it yet, so I'm curious if you see some improvements.

fabricedesre:
Overall it's working, but some icons are not rendered properly as part of the font. Here's the "phone-call" one for instance:

screenshot-featherity-phone-call

locness3:
Wdym, that's a feature (jk)

ericfennis:
Oh wow that's look like a mess.
I think the convertion to paths is not going right. Maybe @yanquisalexander knows what is going on?

yanquisalexander:
> Oh wow that's look like a mess.
> I think the convertion to paths is not going right. Maybe @yanquisalexander knows what is going on?

I'd dare say it's probably something in the SVG 

I have tried the same workflow with other sources (Bootstrap Icons to mention one) and this does not happen :thinking:

---

Although I think it would be useful to highlight this answer...

https://github.com/featherity/featherity/issues/54#issuecomment-663731375

ericfennis:
I see, the `phone-call` has merged path. I think we can fix that one soon.
For the other icons, I saw that the rounded joins are not possible font.
According to the fontcustom documentation we need to first covert stroke-paths to outlines.
So maybe we can use this library: [outline-stroke](https://github.com/elrumordelaluz/outline-stroke). That will convert all the icons to outline strokes.

PR #71, opened 2020-09-10
feat: Bot icon

mittalyashu:
Reference: https://github.com/feathericons/feather/issues/688


ericfennis (approved):
Awesome work, looks great!

PR #76, opened 2020-09-19
Add album icon

ericfennis:
image
100% 
image


locness3:
Feels a little off to me.

ericfennis:
@locness3 hmm ok.

locness3:
I'm just not too sure about the use case for this one.

ericfennis:
Sorry, maybe it's better that I explain more what I make. 😅
But for this one, I imagine it used in UIs for photo applications.

So here in example how it can look.
image

PR #82, opened 2020-10-04
README : Mention Featherity is independent + various changes

locness3:


johnletey (approved):
@locness3 Thanks for doing this - LGTM.

But, I think our rebranding soon will make it even clearer 😄

ericfennis:
Nice work @locness3 !

PR #85, opened 2020-10-06
refactor: use memoization and native NextJS router queries to reduce re-renders

aelfric:
Although the deployed site is fast on Netlify, on my machine in development mode, the search feature was extremely slow, with several seconds' delay between keystrokes.  This PR introduces memorization and debouncing to make the search much faster.  I also added two rudimentary test cases to make sure the site compiles and the search works.


ericfennis:
Nice work!

Issue #80, opened 2020-09-27
TTF is not compiling correctly, paths are mixed and a mess.


ericfennis:
According to thread in #68  the output of the icon-font is not matching the quality of the svgs.
![image](https://user-images.githubusercontent.com/11825403/94374973-7f31b280-0110-11eb-9274-bbdbbb9c7db7.png)

## The Problems:
- overlapping paths are inverting each other.
- paths not having the rounded joins
- paths not having the rounded caps

## Step to reproduce
Download artifact from this [link](https://github.com/featherity/featherity/suites/1202031238/artifacts/17790541).
Open the html file

### Actual behavior
Icons need to be the same as the svgs

## Possible solution
According to the fontcustom [documentation](https://github.com/FontCustom/fontcustom#svg-guidelines) we are not matching the SVG Guidelines.
![image](https://user-images.githubusercontent.com/11825403/94375079-49d99480-0111-11eb-9045-41996ed04cee.png)

Possible solution is converting the svgs stroke paths to outline paths before we build a font file.
I found this library [outline-stroke](https://github.com/elrumordelaluz/outline-stroke) transform paths, maybe we can use this for the solution.

ericfennis:
Fixed by #100

PR #98, opened 2020-10-15
FIX: 404 when installing dependencies using APT

yanquisalexander:
I added a step that updates APT repositories before installing dependencies

### DIFF
--- a/.github/workflows/font.yml
+++ b/.github/workflows/font.yml
@@ -8,14 +8,19 @@ jobs:
   Build:
     runs-on: ubuntu-latest
     steps:
+
     - name: Clone 'Lucide'
       uses: actions/checkout@v2
+    
+    - name: Update repos
+      run: sudo apt-get update
 
     - name: Install FontForge
       run: sudo apt-get install zlib1g-dev fontforge woff2
 
     - name: Clone sfnt2woff-zopfli repo
       run: git clone https://github.com/bramstein/sfnt2woff-zopfli.git sfnt2woff-zopfli
+
 
     - name: Install and move sfnt2woff-zopfli
       run: |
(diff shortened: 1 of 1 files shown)


ericfennis (approved):

ericfennis:
Nice, I have btw a fix for #80, the icons look way better now with the path to outline fix.

yanquisalexander:
> Nice, I have btw a fix for #80, the icons look way better now with the path to outline fix.

Is the solution already implemented or is it a possible solution?

I found this: https://github.com/elrumordelaluz/outline-stroke-cli

Convert multiple SVG files at once

I will try it

ericfennis:
Yeah I already builded it, can make an PR for it.
But I wasn't able to test is yet, not sure if the Action will work, but we can give it a try.

yeah that package is from the same gut that made the outline-stroke package, I used that one. the icons where looking great!

yanquisalexander:
> Yeah I already builded it, can make an PR for it.
> But I wasn't able to test is yet, not sure if the Action will work, but we can give it a try.
> 
> yeah that package is from the same gut that made the outline-stroke package, I used that one. the icons where looking great!

I just tried the CLI, maybe I was wrong about something, but look how the icons look:

Captura de pantalla (145)


Workflow file: https://github.com/yanquisalexander/lucide/blob/master/.github/workflows/font.yml

yanquisalexander:
https://github.com/elrumordelaluz/outline-stroke-cli/issues/3

ericfennis:
yeah those issues I tackled as wel, Is used exactly the solution in the issue you mentioned.
I pushed it to a separated branch: [](https://github.com/lucide-icons/lucide/tree/feature/ttf-fix)
Feel free to continue on this and check if everything is working, maybe you will find some improvements

yanquisalexander:
It's funny that I did a PR to master 20 minutes ago :joy:

ericfennis:
Oh noo, i see 😂😂

Issue #54, opened 2020-07-23
TTF for Featherity


yanquisalexander:
Hi!

Is there any chance that Featherity can be compiled to TTF?

Thanks :hugs:

mishkaio:
There may be a different way to do this but I can do this through an icon library tool I use called Nucleo. This includes eot, ttf, woff, woff2.
[Nucleo-Fonts.zip](https://github.com/featherity/featherity/files/4969349/Nucleo-Fonts.zip)

ericfennis:
@rfbezerra created a script for generating Font files, maybe we check it out.
https://github.com/feathericons/feather/pull/608

jletey:
@mishkaio That's really neat!

@ericfennis I don't have any of those tools installed locally, but we should be able to port it no problem 😄

rfbezerra:
Well, today I tried to build Featherity, to tell you if it works or not.
Maybe the tools have evolved since, so I'm able to build with 3 tools (without inkscape):
- fontforge
- sfnt2woff (npm)
- fontcustom

It was build with success, with only one icon problem (icon-phone-call).
The same process worked well with Feather.

If you want to test, after install these tools, do a simple:
`fontcustom compile ./icons -h -n Featherity -o build -F`

rfbezerra:
I made a docker image to simplicate necessary steps to build a TTF font.
If u want to check, it is this project [svg-to-ttf](https://github.com/rfbezerra/svg-to-ttf).
With that, its simple to build any ttf font from svg glyphs, e.g.:
```shell
docker run --rm -v "${PWD}":/fonts rfbezerra/svg-to-ttf -i ./icons -n Feather -u $(id -u)
```

ericfennis:
@rfbezerra Sorry for the late reply. I was on vacation.

Thanks for your work on this! Looks awesome.
I'm currently working on the NPM package for featherity
First I want to release the package, including the subpackages for Vue and React, after that I will pick this up!

yanquisalexander:
Just to know

Does the Workflow work correctly? 😅

ericfennis:
I forgot to check, but it is working correctly!!
Link of the [artifact](https://github.com/featherity/featherity/suites/1202031238/artifacts/17790541).
I didn't test it yet.

But I don't know how we can publish this nicely?
My suggestion, maybe we need to add this to the new package, then we can serve it through unpkg.

ericfennis:
Fixed by #100
Shortly I will create an solution for to download the icons

PR #100, opened 2020-10-16
Fix problems with SVGs

yanquisalexander:
I have changed part of the Workflow to run a Node script that solves the __peculiarities__ of SVGs files

### DIFF
--- a/.github/workflows/font.yml
+++ b/.github/workflows/font.yml
@@ -39,7 +39,7 @@ jobs:
       run: sudo gem install fontcustom
     
     - name: Install "outline-stroke-cli"
-      run: sudo npm i -g outline-stroke-cli
+      run: sudo npm i -g outline-stroke-cli --unsafe-perm
     
     - name: Convert SVG to Outlined with CLI
       run: outline-stroke ./icons -o converted_icons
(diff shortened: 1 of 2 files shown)


scripts/outline_svg.js
```diff
@@ -0,0 +1,55 @@
...
+init();
+
+function transformForward(node) {
+  if (node.name === "svg") {
+    return {
+      ...node,
+      attributes: {
+        ...node.attributes,
+        width: 800,
+        height: 800
```
ericfennis, line 39: I think it'd better to use round binairy numbers for better scale factor.
icon: 24px * 40 = 960

```suggestion
        width: 960,
        height: 960
```

.github/workflows/font.yml
```diff
@@ -2,6 +2,8 @@ name: Build Lucide
 on:
   push:
     branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
ericfennis, line 6: is there a reason why you want to build the font when a pull-request is opened?
I changed it, because it only build when a PR is merged to master.

ericfennis (commented):
Nice work, I have small things

yanquisalexander (commented):
:+1:

ericfennis:
Nice work, the icons looks way better know!
image

Shortly I will think of a solution for downloading the artifacts, so it will be easier to download the files.

ericfennis:
Closes: #54 
Closes: #80

Issue #83, opened 2020-10-06
Icon request: file-code


jerelmiller:
## Icon Request

* Icon name: file-code
* Use case: source code files
* Screenshots of similar icons:

<img width="123" alt="Screen Shot 2020-10-05 at 8 59 27 PM" src="https://user-images.githubusercontent.com/565661/95157372-acecac00-074d-11eb-83b3-a87bb9feffb1.png">

(source: https://fontawesome.com/icons?d=gallery&q=file-code)

Hey there 👋 

I'm working on the team in the process of migrating a website from one tech stack to another. In the process, we are migrating from font-awesome to feather icons. I see a lot of `file` type icons, but would also love one that looks similar to the screenshot above.

I'd PR this myself, but as an engineer and non-designer, I can't say I know the toolset well enough to know where to begin 😅 . Thanks for the help ahead of time!

ericfennis:
@jerelmiller Welcome to Lucide (before: featherity)!

Nice first issue, I think we can make this one and add it 😊.

johnletey:
Here are the icons from [Octicons](https://primer.style/octicons) @ericfennis, that I personally think look really nice!

<img width="183" alt="Screenshot 2020-10-06 at 20 42 27" src="https://user-images.githubusercontent.com/62398724/95252123-7463d580-0814-11eb-820a-8662612453cb.png"> <img width="182" alt="Screenshot 2020-10-06 at 20 42 50" src="https://user-images.githubusercontent.com/62398724/95252160-80e82e00-0814-11eb-82ae-0888276004bd.png">

ericfennis:
@johnletey Nice suggestions!

ericfennis:
![image](https://user-images.githubusercontent.com/11825403/95261774-42aa3900-082b-11eb-9bba-5402ffb03b43.png)

Issue #93, opened 2020-10-13
John Leteys awesome icons


ericfennis:
@johnletey Some open/old PRs from your old account in the feather repo have some awesome icons!
I think we can add them to Lucide. 

Here is a list of the icons I think are welcome to add to the Lucide repo!
- [x] Currency icons [pr](https://github.com/feathericons/feather/pull/670)
- [ ] code-2 icon [pr](https://github.com/feathericons/feather/pull/671)
- [x] file-check, file-times , folder .. [pr](https://github.com/feathericons/feather/pull/669)
- [x] webcam [pr](https://github.com/feathericons/feather/pull/687)

More PR from [John](https://github.com/feathericons/feather/pulls/jletey)

delnyn:
Also the `file-check, file-times , folder .. pr` is already resolved :)

locness3:
Checked it

PR #106, opened 2020-10-19
icons: File Icons

ericfennis:
Closes: #83 
Contributes to: #93

## File Icons
I refactored the file icons a bit, I noticed that the regular file was different from the others.
I also added the code & digit icons.
The top row are the icons we know, I see them as more icons for an action.
The bottom row are the icons that can be used as informative icons, like if there are "files added, files removed".

image


Andreto:
Maybe throw in some kind of zip file while you're at it.
image

ericfennis:
@Andreto nice idea, I will introduce them in a separate one!

PR #102, opened 2020-10-19
Add indentation icons in/out

Themistoklis:
Indent icon:
indent_grid

Outdent icon:
oudent_grid

PR #109, opened 2020-10-25
Refactor text-align icons

ericfennis:
According to the thread in #102, the align icons needs an redesign.

I inspired the icons from the figma icons. 
I think these icons are more simplified and fit better with the other icons as wel.
image

### DIFF
--- a/icons/align-justify.svg
+++ b/icons/align-justify.svg
@@ -9,7 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <line x1="21" y1="6" x2="3" y2="6" />
-  <line x1="21" y1="12" x2="3" y2="12" />
-  <line x1="17" y1="18" x2="3" y2="18" />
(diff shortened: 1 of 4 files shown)


icons/align-justify.svg
```diff
@@ -9,8 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <line x1="21" y1="10" x2="3" y2="10" />
```
locness3, line 12: This doesn't feel like "justify" to me.

ericfennis: other option:
image

I can also fill it up but then it will be the exact icon as the menu icon.

locness3: Think it's best to "fill it up"

ericfennis: Then it's the same as the menu-icon, I think we can better remove this one and add an tag for it.

ericfennis: I stretched it

locness3 (approved):

PR #111, opened 2020-10-26
Docs : Improve the design guide

locness3:
This is just a proposal.

- Use modal verbs instead of "has"
- Use "icons" instead of "each icon"
- change some other things as well


ericfennis (approved):
Nice, good work

Issue #110, opened 2020-10-26
Site : Allow changing the color of downloaded PNGs


locness3:
It's a bit unconvenient to download a PNG icon without being able to choose its color, you have to manually recolor the picture afterwards.
Please add a color picker, to change the color of the downloadable PNGs.

ericfennis:
Maybe we can add a global color adjuster, like the feather site.

aelfric:
Can I give this one a try?

locness3:
Sure!

PR #113, opened 2020-10-26
Add customization options to the icon display and download

aelfric:
As discussed in #110 , this adds a knob for changing the color of the icons output by the website.  I also went ahead and added sliders for stroke-width and icon size similar to what is on the Feather Icons website.

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -32,6 +32,7 @@
     "@types/react": "^16.9.35",
     "@types/react-dom": "^16.9.8",
     "babel-jest": "^26.5.2",
+    "cheerio": "^1.0.0-rc.3",
     "jest": "^26.5.2",
     "react-test-renderer": "^16.13.1",
     "typescript": "^3.9.5"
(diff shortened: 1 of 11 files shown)


locness3:
Love it! Maybe making the sliders value "float" to the right like this ?
Capture du 2020-10-27 10-41-26

locness3 (approved):

aelfric:
> Love it! Maybe making the sliders value "float" to the right like this ?
> Capture du 2020-10-27 10-41-26

Good idea - done.  

I also fixed the sliders - they are supposed to be filled in on the left of the thumb control, but something in Chakra was making the filled part have a zero height.  I put the values in manually for now, but I suspect this is a Chakra bug.  Will investigate a bit more later.

ericfennis:
@aelfric Nice work!

I noticed one thing, on default each icon have the stroke color on "currentColor", so it takes the font-color that is applied. 
Is it possible to keep that on default? unless you change it of course.

aelfric:
Yes. I can do that. I'll push a fix for it.

Andreto:
The rounded edges on the input look kinda odd.
image

Maybe put no border-radius on the bottom corners. I also think that the color field could have a border around it. Looks a lot better on the bright side.
image

Alternatively, some margin between the input and color field. 
image

Issue #122, opened 2020-11-01
Minor styling issues on the webpage.


Andreto:
### 1. Search input border-radius on dark mode.
**Issue**
The input field has a border-radius but the `chakra-input__group css-r27y85` element is poking out underneath. On light mode this is not an issue since the color of the element is the same as the body.
![image](https://user-images.githubusercontent.com/28541538/97800151-49ae4680-1c33-11eb-98f4-2a4defefea22.png)

**Solution**
Give the `chakra-input__group css-r27y85` element the same border-radius as the input.
or
Give the `chakra-input__group css-r27y85` element the same background-color as the body. (Maybe this messes up the color of the input.)

### 2. Color input box.
**Issue**
The color input elements have a border-radius on the bottom with no padding to the box below.
![image](https://user-images.githubusercontent.com/28541538/97551848-c7741700-19d3-11eb-85b7-47984666b0a5.png)

**Solution**
Maybe put no border-radius on the bottom corners. I also think that the color field could have a border around it. Looks a lot better on the bright side.
![image](https://user-images.githubusercontent.com/28541538/97551804-b88d6480-19d3-11eb-9765-31da27703ec5.png)

Alternatively, some margin between the input and color field. 
![image](https://user-images.githubusercontent.com/28541538/97552590-b972c600-19d4-11eb-9109-5ca6d65b44e7.png)

ericfennis:
@Andreto (1) Wow didn't saw that, really sharp 😄 
(2) Good point, I think the color picker can improved, I noticed it's really laggy if you drag the color picker and heu slider as wel, I think it fires too much events. @aelfric Do you have time to take a look to it?

aelfric:
Yes. I can take a look.

PR #123, opened 2020-11-02
Fix style issues and performance issues for Color Picker and Search Bar

aelfric:
This addresses #122 .

Note - the fix for the performance issue when dragging in the color picker was to downgrade the library I used for the color picker to a previous version.  See casesandberg/react-color#731 . This may not be the right permanent solution.  I'll poke around a bit more in the react-color documentation to see if there is a better approach, but this should work in the short term.


aelfric:
Looks like someone submitted a PR to `react-color` a few weeks ago that I think will fix the slow down issue but it hasn't been merged yet. I guess we'll just need to wait for that before bumping the version.

ericfennis:
Awesome @aelfric ! I works great!

ericfennis (approved):

ericfennis:
@aelfric Look like we have some conflicts in the lock file. Can you run `yarn` again to update it?

PR #129, opened 2020-11-10
Add pencil icon

ericfennis:
Adds pencil icon
image

### DIFF
--- a/icons/pencil.svg
+++ b/icons/pencil.svg
@@ -1,0 +1,14 @@
+<svg
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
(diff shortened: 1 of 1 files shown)


locness3:
edit-4 basically

ericfennis:
Yeah true, I a bit doubting what the right name is .. 🤔

ericfennis:
@locness3 Do you think we need to call it edit-4?

locness3:
No idea

delnyn:
Also shouldn't it be rounded? I personally don't like the shape of this icon. Maybe you could do something more like Octicon's pencil, but... 2px rounded? 
image

PR #132, opened 2020-11-12
Add icon podcast

sourcerer52:
For feather icon [Podcast](https://github.com/feathericons/feather/issues/98)

image


ericfennis:
@sourcerer52 Welcome to the Lucide community, thanks for opening this PR!

I like the concept, but I think some lines are a bit distorted. 
Can you try to make the waves perfectly round, like the [`radio`](https://lucide.netlify.app/icon/radio) icon ?

sourcerer52:
I see. How about this?

image

locness3:
Wouldn't the radio icon fit for podcast?

ericfennis:
@locness3 I agree, but I think it's nice to have alternative icon with a mic. So I think we need to rename this icon to `mic-radio` or something. 

@sourcerer52 This looks better. I noticed the line of the microphone holder is not perfectly round. Maybe it is a good idea to use the original mic icon and adjust it follow the guidelines.

sourcerer52:
@ericfennis As per your suggestion, I used the mic-radio icon for the podcast. Have a look, if a further change is required. 

image

ericfennis:
@sourcerer52 I think some elements in the icon are a bit distorted.
I hope this feedback maybe clears it more up.

1. There need to be 2px between elements. See the [Design Guide](https://github.com/lucide-icons/lucide/blob/master/docs/ICON_DESIGN_GUIDE.md)
2. The microphone its self look distorted, please try to make it perfectly round like the red stroke I drew.
3. The waves looks a bit off. I think they work better if you make them from the path of circle.

image

sourcerer52:
Apologies for the delay. I was wondering if we could add something like this, the Apple podcast icon?

image

ericfennis:
@sourcerer52 Sorry for the late reply. I think this fits better. I tried some as wel, but it fastly becomes a mess. Feel free to submit!

ericfennis (changes requested):

ericfennis (approved):

PR #134, opened 2020-11-14
Add hand icon

ericfennis:
I was quite hard, but this is what came up with.

image
100% image


Andreto:
My first impression was that it was very wide, but looking at it in 100% it looks a lot better.

100%
image

ericfennis:
@Andreto Yeah that's exactly how I saw it. I think this icon will work. 

I also made some other versions: `pointer` and `grab`
image

locness3:
What about the pointer and grab versions?

PR #151, opened 2020-12-02
feat: Add crown icon

ahtohbi4:
Icon `crown` was added

icon crown

### DIFF
--- a/tags.json
+++ b/tags.json
@@ -59,6 +59,7 @@
   "credit-card": ["purchase", "payment", "cc"],
   "crop": ["photo", "image"],
   "crosshair": ["aim", "target"],
+  "crown": ["crown"],
   "database": ["storage", "memory"],
   "delete": ["remove"],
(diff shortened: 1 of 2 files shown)


tags.json
```diff
@@ -59,6 +59,7 @@
   "credit-card": ["purchase", "payment", "cc"],
   "crop": ["photo", "image"],
   "crosshair": ["aim", "target"],
+  "crown": ["crown"],
```
ericfennis, line 62: Add a name as a tag is not necessary, search engine also search by name.
Here some other suggestions as tags

```suggestion
  "crown": ["king", "winner", "favourite"],
```

tags.json
```diff
@@ -59,6 +59,7 @@
   "credit-card": ["purchase", "payment", "cc"],
   "crop": ["photo", "image"],
   "crosshair": ["aim", "target"],
+  "crown": ["king", "winner", "favourite"],
```
locness3, line 62: award?

Issue #128, opened 2020-11-06
Reply, Reply All, and Forward


Andreto:
**What design direction should we go in for Reply, Reply All, and Forward?**
In the request issues on Feather, two example designs were given, similar to the icons in Outlook and Gmail.

**Requests from feather:** 
Reply: https://github.com/feathericons/feather/issues/380
Replay All: https://github.com/feathericons/feather/issues/381
Forward: https://github.com/feathericons/feather/issues/384

**Outlook (simple lines):**
![image](https://user-images.githubusercontent.com/28541538/98411182-e6c71000-2075-11eb-8574-2147aecc3ca2.png)
**Gmail (filled shapes):**
![image](https://user-images.githubusercontent.com/28541538/98411775-f6932400-2076-11eb-88d0-51aaf2e3faad.png)

I prefer the Gmail approach a bit more and I made icons for reply and reply all:
![image](https://user-images.githubusercontent.com/28541538/98414894-7c659e00-207c-11eb-8511-3dcdf72c39e3.png)
![image](https://user-images.githubusercontent.com/28541538/98414731-2f81c780-207c-11eb-9faf-16e916202cb3.png)

ericfennis:
I think I like it more like this way: 
Reply & reply-all : https://github.com/feathericons/feather/pull/574/files
Forward : https://github.com/feathericons/feather/pull/578

It matches way better with the rest of the arrow icons

Andreto:
I'm a bit skeptical about having the same icon for both reply and forward but in different directions. To me, it doesn't feel totally given that left should be reply and right should be forward. We already have normal arrows tough so maybe it's up to users to decide which icons to use.

ericfennis:
Yeah true, I'm curious what other people think about it.

delnyn:
I personally prefer the one-line icon. Also for the forward, the simple arrow (right or left whatever) works fine IMO (just like Outlook and Gmail).

ericfennis:
I agree with @delnyn, looking at gmail for example it. Here a screenshot. ![Screenshot_20201111-125900.jpg](https://user-images.githubusercontent.com/11825403/98809419-cb655780-241d-11eb-99ec-4cf02ef75e5e.jpg)

ericfennis:
@Andreto I think we can with both ideas. So designers have the choice to implement one. 
I think we can rename it to `arrow-big-corner-left`, `arrow-big-corner-left-double`. So we can create more big arrows as well.

[...]

PR #153, opened 2020-12-04
Icons: Big arrows

Andreto:
`arrow-big-right` image

`arrow-big-left` image

`arrow-big-up` image

`arrow-big-down` image

Part of #128;

### DIFF
--- a/icons/arrow-big-down.svg
+++ b/icons/arrow-big-down.svg
@@ -9,5 +9,5 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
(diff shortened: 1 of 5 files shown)


ericfennis:
@Andreto Nice work.
I think the icons look a bit better if the arrow length is a bit shorter.
So it also can be used for keyboard arrow-keys. 

image

I'm curious what you think.

ericfennis:
Any update on this @Andreto ?

Andreto:
@ericfennis  Totally agree with you. I think the -2px looks the best and I made a commit for it.

ericfennis (approved):

PR #159, opened 2020-12-21
feat: Add locate and locate-fixed icons

### LINKED ISSUE #156: Locate icon
## Icon Request

* Icon name: locate
* Use case: on map type applications, find the location of the user
* Screenshots of similar icons:  I actually already made one, and want to submit it but I'd like others' opinions on which one is best.  Personally I like the bottom right one the best, it sticks to the "simpler is better" theme that feather has going on and follows all the rules.

image

I can submit a PR if you agree that one is the best option!

csandman:
Add 2 new icons, `locate` and `locate-fixed`

closes #156

### DIFF
--- a/icons/locate-fixed.svg
+++ b/icons/locate-fixed.svg
@@ -1,0 +1,18 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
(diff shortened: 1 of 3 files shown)


ericfennis:
@csandman Looks great! 🚀

ericfennis (approved):

PR #160, opened 2020-12-22
Fix optimizeSvgs script

csandman:
- Switch `SVGO.optimize` function to returning a promise instead of accepting a callback - https://github.com/svg/svgo/releases/tag/v1.0.0
- Pin prettier to the last version before `1.18.0` which removed the self closing JSX brackets.  On a clean install without a `package-lock.json` a higher version would automatically be installed - https://prettier.io/blog/2019/06/06/1.18.0.html#stop-converting-empty-jsx-elements-to-self-closing-elements-6127-by-duailibe
  - Add `{ parser: babel }` option to `format` to avoid warnings on each iteration
- Fix some incorrect arguments for writing the SVGs to disk

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "htmlparser2": "^4.1.0",
     "jest": "^26.4.2",
     "minimist": "^1.2.5",
-    "prettier": "^1.8.2",
+    "prettier": "1.17.1",
     "rollup": "^2.7.3",
     "rollup-plugin-commonjs": "^10.1.0",
     "rollup-plugin-license": "^2.0.0",
--- a/scripts/helpers.js
+++ b/scripts/helpers.js
@@ -82,7 +82,7 @@ export const readSvg = (fileName, directory) => fs.readFileSync(path.join(direct
  * @param {string} content
  */
 export const writeSvgFile = (fileName, outputDirectory, content) =>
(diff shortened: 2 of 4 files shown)


ericfennis:
Nice work!
I will test it tomorrow.

ericfennis (approved):
Works great @csandman !!

ericfennis:
I saw that the current script is also did some optimizations for 137 icons.
I think we need to update them.

Maybe we can also create a github action for it, auto reviewing icons or something.

csandman:
> I saw that the current script is also did some optimizations for 137 icons.
> I think we need to update them.
> 
> Maybe we can also create a github action for it, auto reviewing icons or something.

I agree, these should be updated to keep the whole package consistent.  You could either add a Github Action or a pre-commit hook with something like [husky](https://github.com/typicode/husky) to ensure icons are optimized before a PR is submitted.  I have implemented the latter before and it is very straightforward so I could add it if you'd like!

ericfennis:
> I agree, these should be updated to keep the whole package consistent.  You could either add a Github Action or a pre-commit hook with something like [husky](https://github.com/typicode/husky) to ensure icons are optimized before a PR is submitted.  I have implemented the latter before and it is very straightforward so I could add it if you'd like!

 @csandman Yeah true and good idea,  adding husky is very easy to implement and contributes to the code quality.
I appreciate it, feel free to submit!

PR #166, opened 2021-01-05
Add hook to optimize staged SVGs before commit

csandman:
I added a pre-commit hook using husky and lint staged to run the `optimizeSvgs` script on only the staged SVGs when a contributor commits new icons.  They should probably do this before contributing anyway as it can sometimes change the way an icon looks, but this will aid in keeping the whole project consistent.

Mentioned here: https://github.com/lucide-icons/lucide/pull/160#issuecomment-753416087


ericfennis:
Nice work!

ericfennis (approved):

PR #167, opened 2021-01-05
Optimize all existing icons

csandman:
The results of running the `optimize` script on all existing icons.  In combination with #166 this will make all current and future icons consistent.

Mentioned here: https://github.com/lucide-icons/lucide/pull/160#issuecomment-752731198


ericfennis (approved):
LGTM

PR #204, opened 2021-02-03
add sidebar open & close icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-02-03 at 22 28 09@2x |


ericfennis:
Nice icon!

mittalyashu:
@ericfennis it doesn't look good in the sense that, it doesn't depicts the difference between sidebar and main.

| Preview |
| --- |
| CleanShot 2021-02-05 at 10 09 53@2x |

ericfennis:
@mittalyashu I think it will work just fine. 
To me the other icons,the arrow looks not aligned to something. And that looks weird to me.

ericfennis (approved):

ericfennis (changes requested):
@mittalyashu Can you try to make it consistent with the sidebar icon? 
image

PR #215, opened 2021-02-07
add fire icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-02-07 at 21 57 17@2x |


ericfennis:
@mittalyashu this icon already exists. The name is flame.

PR #217, opened 2021-02-07
add desktop icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-02-07 at 22 01 11@2x |


ericfennis:
@mittalyashu 
I'm convinced yet with the name.
Maybe rename this to `devices` ?

ericfennis (changes requested):
^

locness3:
`monitor-speaker`, with a lot of tags?

locness3:
Wait, doesn't really match both of these two icons

mittalyashu:
@ericfennis I think we should use `devices` as category name.

As for icon we can think of some other name?

ericfennis:
I thought of `devices` because that where spotify used them for. Otherwise I would say the same as @locness3. `speaker-desktop` or `desktop-speaker`

locness3:
We never used `desktop` anyhwere else; I'd use `monitor`.

ericfennis:
@locness3 You're right. `monitor-speaker`. You ok with that @mittalyashu ?

PR #218, opened 2021-02-07
add layout plus/minus icon

mittalyashu:
| `layout-plus` | `layout-minus` |
| --- | --- |
| CleanShot 2021-02-07 at 22 25 20@2x | CleanShot 2021-02-07 at 22 25 15@2x |

### DIFF
--- a/icons/layout-minus.svg
+++ b/icons/layout-minus.svg
@@ -1,0 +1,16 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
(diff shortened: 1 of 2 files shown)


mittalyashu:
Should I expand the `layout` icon by 1px too? To make it consistent?

ericfennis:
@mittalyashu I think we need to keep the icons in serie all consistent.
So this icon need to be changed or all the layout icons need to be changed.
Personally I prefer to keep the layout icons as it is, because it matches the other `square` icons.

ericfennis:
I close this one. 
This one is currently hard to fix.
I couldn't find any solution for know.

PR #223, opened 2021-02-08
Add contributors to icon overlay and add dot

### LINKED ISSUE #169: [website] modification status indicator for icons
since this is a fork of feather icons, would be good to have some sort of indication for icon status. 

ie. whether an icon is —
(a) unchanged from feather,
(b) modified from feather, or
(c) new icon not shipped in feather. 

This could either be a standalone docs page on GH or if integrated into https://lucide.netlify.app it could be in the form of a dot or triangle 

as there's currently no changelog (that I can see), the only way to find out if icons have changed or new from feather is to compare them or go through the commit history

pros for current feather users — 
able to do a quick glance on what the new sets are; easily allow designers/devs to make the judgement call on whether or not to switch. and if so, whether icons have been modified to avoid surprises

perhaps would look something like this — 
Frame 1
Frame 2

thoughts?

ericfennis:
Closes #169


site/src/components/IconList.tsx
```diff
@@ -63,6 +66,28 @@ const IconList = ({icons}) => {
...
+              { contributers?.length ? (
+                <Tooltip
+                  hasArrow
+                  label="This is new or modified icon"
+                  bg={colorMode === 'light' ? theme.colors.white : theme.colors.gray[700]}
+                  color={colorMode === 'dark' ? theme.colors.white : null}
+                >
+                  <Box
+                    {
+                      ...{
```
mittalyashu, line 78: @ericfennis are you sure this part is correct and not giving syntax error?

ericfennis: yeah this is prop destructing in react, so the object properties will be spreadout as component props

locness3:
It's contribut**o**rs, not contribut**e**rs.

ericfennis:
@locness3 Renamed everything!

PR #224, opened 2021-02-08
Update tags

ericfennis:


mittalyashu:
@ericfennis how about using mongoose for storage the tags?

Tags is a part of search (one of the main aspects) that way we (people who will have access to mongoose db) can add/remove tags at scale without the git workflow (commit > pr > merge).

and considering the future when we have more icons, with a flexible search in that case it will definitely help a lot.

ericfennis:
I'm open for alternatives. But I think it is nice that we can review changes.

PR #225, opened 2021-02-09
add option icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-02-09 at 18 32 21@2x |

_For mac users this icon will make sense_


ericfennis (approved):

locness3:
Rather `option-key`? With an `alt` tag?

PR #118, opened 2020-10-31
Icons: Tools and Construction

Andreto:
**So far:** Axe, Gavel, Hammer, Hardhat, Shovel
image
I would like to add a screwdriver but I couldn't really get it right.

**Satisfies Requests:**
Gavel: https://github.com/feathericons/feather/issues/59
Hammer: https://github.com/feathericons/feather/issues/717

PR #230, opened 2021-02-22
[fix] increase hammer icon size

mittalyashu:
| Before | After |
| --- | --- |
| CleanShot 2021-02-22 at 21 24 22@2x | CleanShot 2021-02-22 at 21 24 15@2x |


ericfennis:
@mittalyashu Can you explain why this refactor was needed? @Andreto created this icon in series in #118 with some other tool icons, so each tool have a consistent size and handle.

Andreto:
@mittalyashu The icon was created to be consistent with other icons with handles and especially [axe](https://lucide.dev/icon/axe) and [gavel](https://lucide.dev/icon/gavel). The handle could maybe be a bit longer but otherwise, I don't think that the sizing is a problem.

ericfennis:
I close this one for now, if you have a clear explanation we can consider opening this PR again.
You still open to comment here.

PR #236, opened 2021-02-25
add storage icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-02-25 at 16 00 12@2x |


locness3:
Not satisied with the name.

ericfennis:
@mittalyashu @locness3 
Is it not just another pie-chart icon? 
First I didn't know this icon design was associated with a storage.
Until I searched it, this is what I found https://icons8.com/icons/set/storage 
image


So basically it's another design of a `pie-chart`, I think the best name is then `pie-chart-2`.

mittalyashu:
As for the naming is concerned, it can be named anything.

But most of the time, we use different icon accommodation our use-case, even if the icon (might) mean something else.

ericfennis:
Correct, we should alway try naming icons as general as possible. 
And we should keep variants together. I see this one as a variant on the [`pie-chart`](https://lucide.dev/icon/pie-chart?search=pie&iconName=pie-chart) icon.

mittalyashu:
So, what should be the name of this icon?

ericfennis:
`pie-chart-2` please.

ericfennis (changes requested):
Can you match the shape of the icon with the `pie-chart` icon?
image

locness3:
> `pie-chart-2` please.

or `pie-chart-(out)line`?

locness3:
Feel free to reopen.

PR #237, opened 2021-02-25
add circle-check-2 icon

mittalyashu:
| zoom | 100% |
| --- | --- |
| CleanShot 2021-02-25 at 16 07 32@2x | CleanShot 2021-02-25 at 16 07 40@2x |

### DIFF
--- a/icons/check-circle-2.svg
+++ b/icons/check-circle-2.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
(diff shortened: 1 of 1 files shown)


ericfennis:
hmm looks not like the one you posted in discord. I think the check can bigger. on 100% scale the icons looks too small.
image

mittalyashu:
@ericfennis the reason I made the check smaller in size, as it looks like it will be touching the circle at 100% view.

Also, we try increasing the `stroke-width` it will look even more weird.

ericfennis (approved):

locness3 (changes requested):
All good except it should be `check-circle-2`.

mittalyashu:
Isn't the file name is already that `circle-check-2`?

locness3:
It should be the other way around. `check-circle-2`, not `circle-check-2`

ericfennis:
@locness3 It's `check-circle-2` !

PR #238, opened 2021-02-26
add selector icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-02-26 at 10 17 27@2x |


ericfennis:
@mittalyashu Can you  try to match it with the chevrons icons. And maybe we can call it `chevrons-up-down`.
image

locness3:
Needs renaming.

ericfennis:
I agree with @locness3. maybe rename it to: `chevrons-up-down`

Issue #244, opened 2021-02-27
Identical keys are being generated if SVG props are identical


FPDK:
## Prerequisites

* Version: 0.14.1
* Are you running from source/master: no
* Are you using a released build: yes

## Step to reproduce

Insert a UploadCloud icon in a React app
Watch console for error

## Any message or error

```
Warning: Encountered two children with the same key, `8cfiuk`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in svg (created by UploadCloud)

```

Looking at the generated upload-cloud.js file we can see that there are two duplicate keys with the name 8cfiuk

```
import createReactComponent from "../createReactComponent";
import defaultAttributes from "../defaultAttributes";
var UploadCloud = createReactComponent("UploadCloud", ["svg", defaultAttributes, [["polyline", {
  points: "16 16 12 12 8 16",
  key: "8cfiuk"
}], ["line", {
  x1: "12",
  y1: "12",
  x2: "12",
  y2: "21",
  key: "16to6l"
}], ["path", {
  d: "M20.39 18.39A5 5 0 0018 9h-1.26A8 8 0 103 16.3",
  key: "xftpgh"
}], ["polyline", {
  points: "16 16 12 12 8 16",
  key: "8cfiuk"
}]]]);
export default UploadCloud;
```

PR #245, opened 2021-02-27
Fix duplicate keys being generated if SVG attributes are identical

FPDK:
Fixes issue #244 by adding an extra attribute to the generated hash

### DIFF
--- a/scripts/render/renderIconNodes.js
+++ b/scripts/render/renderIconNodes.js
@@ -18,10 +18,11 @@ export default (iconsObject, options) => {
     const svgString = iconsObject[icon];
     const dom = parseDOM(svgString);
 
-    const children = dom.map(element => {
+    const children = dom.map((element, index) => {
       if (options.renderUniqueKey) {
         const hashSource = {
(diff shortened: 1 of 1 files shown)


ericfennis:
I think this fix is not needed. The hash is generated based on the svg elements, I checked the code of the [`cloud-upload.svg`](https://github.com/lucide-icons/lucide/blob/master/icons/upload-cloud.svg?short_path=0e2da28). And the icon contains double elements. 
So I think we need to fix this icon.

FPDK:
What if I change it so it throws on duplicate keys?
Then it could serve as a test, at least when _renderUniqueKey_ is on.

ericfennis:
@FPDK yeah exactly, we need to add check in the build.

ericfennis:
I added a check for the duplicated children in [a03ece6](https://github.com/lucide-icons/lucide/pull/272/commits/e8ca0a479ceacb27a5f1765a9f631045d77e8af2). So the build will fail if there are duplicated children.

PR #256, opened 2021-03-08
Add bug icon

ericfennis:
image

100% bug

### DIFF
--- a/icons/bug.svg
+++ b/icons/bug.svg
@@ -1,0 +1,22 @@
+
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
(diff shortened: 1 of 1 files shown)


mittalyashu:
Can you try in something like this style, instead of keeping the body in cylinder shape.

CleanShot 2021-03-10 at 20 39 42@2x

insect-icon-22

_the above screenshots are copyright to its respective owners_

ericfennis:
@mittalyashu Like this?
image

mittalyashu:
@ericfennis the first one, definitely looks like a bug but his head seems to be missing.

ericfennis:
I think from a 100% view the second design looks better. Also because we no really use oval shapes in our icons.
image

PR #258, opened 2021-03-09
add alphabet-a icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-03-09 at 10 40 36@2x |


locness3:
maybe just `a`?

mittalyashu:
I think adding "alphabet" before file name is more descriptive.

> An alphabet is a standardized set of basic written symbols or graphemes that represent the phonemes of certain spoken languages.

locness3:
This implies more alphabet icons then

mittalyashu:
okay

ericfennis:
We should not adding typography, see thread in #327.

Issue #119, opened 2020-10-31
Icon Requests from Fether


Andreto:
### To do
Id | Title | Engagement | Linked PRs
-|------------ | ------------- | --------
[#17](https://github.com/feathericons/feather/issues/17) | Wallet | 4 | [#658](https://github.com/feathericons/feather/pull/658)
[#21](https://github.com/feathericons/feather/issues/21) | newspaper | 0 |
[#27](https://github.com/feathericons/feather/issues/27) | gender | 15 | [#46](https://github.com/feathericons/feather/pull/46)
[#28](https://github.com/feathericons/feather/issues/28) | media-controls/repeat-1 | 6 | [#655](https://github.com/feathericons/feather/pull/655)
[#32](https://github.com/feathericons/feather/issues/32) | Food Icon | 6 |
[#53](https://github.com/feathericons/feather/issues/53) | factory | 0 |
[#56](https://github.com/feathericons/feather/issues/56) | forest | 6 |
[#59](https://github.com/feathericons/feather/issues/59) | Gavel | 0 |
[#61](https://github.com/feathericons/feather/issues/61) | Campsite | 4 | [#651](https://github.com/feathericons/feather/pull/651)
[#68](https://github.com/feathericons/feather/issues/68) | bed | 9 | [#662](https://github.com/feathericons/feather/pull/662)
[#69](https://github.com/feathericons/feather/issues/69) | sofa | 0 |
[#70](https://github.com/feathericons/feather/issues/70) | bath | 0 |
[#86](https://github.com/feathericons/feather/issues/86) | sum | 17 | [#660](https://github.com/feathericons/feather/pull/660)
[#98](https://github.com/feathericons/feather/issues/98) | podcast  | 0 |
[#100](https://github.com/feathericons/feather/issues/100) | money | 8 | [#670](https://github.com/feathericons/feather/pull/670)
[#113](https://github.com/feathericons/feather/issues/113) | Filled thermometer | 0 |
[#149](https://github.com/feathericons/feather/issues/149) | cake | 2 |
[#156](https://github.com/feathericons/feather/issues/156) | Filled circle | 2 |
[#157](https://github.com/feathericons/feather/issues/157) | RFI: Trophy | 10 |
[#170](https://github.com/feathericons/feather/issues/170) | Kola Nut  | 0 |
[#177](https://github.com/feathericons/feather/issues/177) | keyboard | 1 |
[...]

PR #261, opened 2021-03-12
add utensils icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-03-12 at 12 59 06@2x |

Reference https://github.com/lucide-icons/lucide/issues/119


locness3:
may want seperate icons with appropriate tags too

mittalyashu:
@locness3 sure

But then, there is a also very common use-case for using utensils as icon too.

locness3:
Yeah but both is good

ericfennis:
Hmm I don't know with this handles. Maybe we can try single line handles?
Maybe from this thread? https://github.com/feathericons/feather/issues/552

mittalyashu:
Oh! That's interesting. Will try that.

PR #265, opened 2021-03-16
fix: small error

0xflotus:


ericfennis (approved):
@0xflotus Welcome to the Lucide community!
Thanks for the fix!

PR #269, opened 2021-03-19
Generate index.d.ts when build

SMAH1:
For use in typescript project like angular.


ericfennis:
@SMAH1 Thanks for this. I think I will include this in #272

PR #281, opened 2021-04-06
Add lucide-angular again!

SMAH1:
After merge `buid d.ts`, `lucide-angular` is not availabe.
I request it again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ We're trying to expanding the icon set as much as possible while keeping it nice
   * [Web](#web)
   * [React](#react)
   * [Vue](#vue)
+  * [Angular](#angular)
   * [Figma](#figma)
 * [Contributing](#contributing)
 * [Community](#community)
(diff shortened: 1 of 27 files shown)


ericfennis:
Awesome work @SMAH1 !

I'm curious if it is also possible to just import icons as components instead of a global icon component. Or is that quite uncommon way of developing in angular, sorry I'm not very familiar with angular. If you don't know what I mean, look at the `lucide-vue` or `lucide-react` package.

ericfennis:
I think Kevin Kreuzer wrote a nice article about it:  https://medium.com/angular-in-depth/how-to-create-an-icon-library-in-angular-4f8863d95a#d31c
If you read from the `threeshaking` you see that he created a way to import each icon separately. What we also do for the other packages to encourage threeshaking.

SMAH1:
In angular, It is unusual to use ‍‍‍‍`import icons as components`. For example see [this](https://github.com/michaelbazos/angular-feather). `Component change` is done by changing the variable.
Angular use webpack and it eliminate `dead-code`. In `lucide-angular` Import icon before use it. Please see README of `lucide-angular`.
I hope I understood correctly.

ericfennis:
Ah ok understood. Let's merge it and create `beta` release, check if everything is working correctly.

PR #291, opened 2021-04-18
Lucide preact package

### LINKED ISSUE #274: Feature / Preact package
Add a package for https://preactjs.com/ along the line of `lucide-react`.

ericfennis:
closes #274

Currently still in development, need to be tested.
Also typescript types are not ready yet.

Released pre-release version on:
`0.15.1-beta.1`


ericfennis:
Just released `0.15.1-beta.1` with typescript support.

knoan:
In working order when consuming through vitejs.

knoan:
Still some issues with typescript definitions 

```


node_modules/lucide-preact/dist/lucide-preact.d.ts:4:11 - error TS2320: Interface 'SVGProps' cannot simultaneously extend types 'HTMLAttributes<EventTarget>' and 'SVGAttributes<SVGElement>'.
  Named property 'onWheelCapture' of types 'HTMLAttributes<EventTarget>' and 'SVGAttributes<SVGElement>' are not identical.

4 interface SVGProps extends JSX.HTMLAttributes, JSX.SVGAttributes {
            ~~~~~~~~

node_modules/lucide-preact/dist/lucide-preact.d.ts:7:31 - error TS2304: Cannot find name 'T'.

7   ref?: string | ((component: T) => any);
                                ~

node_modules/lucide-preact/dist/lucide-preact.d.ts:11:18 - error TS2430: Interface 'LucideProps' incorrectly extends interface 'Partial<SVGProps>'.
  Types of property 'size' are incompatible.
    Type 'string | number | undefined' is not assignable to type 'number | undefined'.
      Type 'string' is not assignable to type 'number | undefined'.

11 export interface LucideProps extends Partial<SVGProps> {
                    ~~~~~~~~~~~


```

ericfennis:
@knoan Thanks for testing it. 
I tested it in vite, weird I didn't see those errors. I will look in to i!

knoan:
> I tested it in vite, weird I didn't see those errors. I will look in to i!

Those error messages come from `tsc`: I'm currently using it to compile a library that's going to depend on lucide-preact.

Vite seems to compile the same code base with no problem, but as it's clearly stated in the docs it doesn't rely on tic and it doesn't perform type checking.

Looks like the current implementation is functionally correct, but still not correctly type checked according to tsc.

knoan:
> I fixed the types, can you check if everything is also working on your side?

Compiling with no issues. Looking forward to the first stable release ;-)

knoan:
Mmm… [`createPreactComponent.js `](https://github.com/ericfennis/lucide/blob/lucide-preact/packages/lucide-preact/src/createPreactComponent.js) still depends on `preact/compat`

```typescript
import { forwardRef } from 'preact/compat';
```
Beside negating the initial goals of the PR, the inclusion of `preact/compat` activates Preact's compatibility mode, wreaking havoc for instance on `onChange` event handlers (see https://preactjs.com/guide/v10/differences-to-react/#use-oninput-instead-of-onchange).

Unfortunately, no straightforward replacement is available (see https://preactjs.com/guide/v10/switching-to-preact/#forwardref).

ericfennis:
@knoan Hmm interesting. 
So if i'm correct, when adding this package to a preact project will break `onChange` events on elements if implemented?

If that is the case or not, do you think this ready to release?

knoan:
> @knoan Hmm interesting.
> So if i'm correct, when adding this package to a preact project will break `onChange` events on elements if implemented?

Correct.

> If that is the case or not, do you think this ready to release?

I'm afraid this is a blocking issue, as it would cause out of the blue and very difficult to diagnose bugs.

ericfennis:
@knoan Hmm interesting. 
So if i'm correct, when adding this package to a preact project will break `onChange` events on elements if implemented?

If that is the case or not, do you think this ready to release?

knoan:
Still minor typing issues when compiling with `tsc`:

```
node_modules/lucide-preact/dist/lucide-preact.d.ts:6:46 - error TS2314: Generic type 'RefObject' requires 1 type argument(s).

6   ref?: string | ((component: any) => any) | RefObject;
                                               ~~~~~~~~~
```

Having a look at it…

knoan:
Ok. The missing type parameter on `RefObject` in `scripts/buildTypes.js` could be fixed and simplified as:

```typescript
interface SVGProps<T extends EventTarget=SVGElement> extends Omit<JSX.SVGAttributes, "ref" | "size"> {
	key?: string | number;
	ref?: string | ((component: any) => any) | RefObject<T>;
}
```

However, I'm not sure to completely follow the logic here: `((component: any) => any)` seems to be already covered by the original definition of `ClassAttributes ` in `preact/src/index.ts.d`:

```typescript
type RefObject<T> = { current: T | null };
type RefCallback<T> = (instance: T | null) => void;
type Ref<T> = RefObject<T> | RefCallback<T>;

interface ClassAttributes<T> extends Attributes {
	ref?: Ref<T>;
}
```

Unless I'm missing something critical in the inner working of the library, `SVGProps` could be dropped, simplifying the 
definition of `LucideProps` as:

```typescript
export interface LucideProps extends Partial<Omit<JSX.SVGAttributes, "key" | "size">> {
	key?: string | number;
	size?: string | number;
	color?: string;
}
```

Along the same lines, unless we have some cogent reason to restrict keys to `string | number`, I'd preserve 
the original (expected) definition `type Key = string | number | any;`, further simplifying as:

```typescript
export interface LucideProps extends Partial<Omit<JSX.SVGAttributes, "size">> {
	size?: string | number;
	color?: string;
}
```

Will test the latest definition on a real app in the morning: back to you asap.

ericfennis:
@knoan thanks for looking with me. Good findings. I added the ObjectRef because I came in some problems if you pass the ref prop. But I will try to add you solution.

I'm not sure is the ClassAttributes are applied, I'm using functional components to render the svg. Or are they added in the `h()` function?

Thanks for the type refactoring, I will apply it.

Other problems I found are with the props. I removed the forwardRef function. But it didn't transforms props to attributes as kebabcase anymore. So when adding stroke width it rendered:
```html
<svg .. strokeWidth=4 ..>
```

I'm not sure if I'm doing something wrong, or thats the way how preact works.

knoan:
We could also try to steal a page from https://github.com/ForsakenHarmony/preact-feather/tree/master/src

ericfennis:
Hmm testing the package, but I got this error.
I tried to find, but still no luck.

@knoan Any ideas what I'm doing wrong?
image

knoan:
I seeing the same error under vite development server:

- apparently with no operational impact and only on the first loaded icon
-  again, apparently with no impact on the compiled version

Keeping an eye on it, but to this date unable to pinpoint the cause: back to you as soon as I have an update.

ericfennis:
Yeah true the package is just working, icons and props are working fine.
Maybe I can ask the preact community what I'm doing wrong.

ericfennis:
@knoan I fixed it, it's working now!
I'm going to merge it

knoan:
Fine. Will report from my side after upgrading.

knoan:
No more errors after upgrading to beta.7. Up and running 🥳

ericfennis:
@knoan Awesome!! 🔥🔥

PR #297, opened 2021-04-30
Add vibrate icon

ericfennis:
image


locness3:
Would've used a different name but ok

PR #301, opened 2021-05-01
add person icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-05-01 at 15 12 05@2x |

### DIFF
--- a/icons/person.svg
+++ b/icons/person.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="12" cy="5" r="3" />
+  <path d="M8 11h8v4l-2 2v5h-4v-5l-2-2v-4z" />
+</svg>


ericfennis:
Hmm, I'm not very fan of the design.
I just merged the `person-standing` (accessibility) icon. So what should this icon add?

mittalyashu:
Is the person's head too big?

locness3:
What's the use case?

mittalyashu:
Use case:

* Bathroom
* Person standing

In older version of Google maps there used to be a feature call Peek where you drag the person icon on the map.

mittalyashu:
@locness3 What do you think?

locness3:
Not sure we need that icon.

mittalyashu:
okay 🙂

locness3:
That's only my opinion.

delnyn:
do we need this @ericfennis @locness3 @mittalyashu?

ericfennis:
Hmm I don't know. I think I should match with this one: https://lucide.dev/icon/person-standing

mittalyashu:
This icon can also be used to define "male" gender 🤷‍♂️

image

delnyn:
no, because there should also be an icon for females, and maybe non-binary people.  
(adding a skirt to it doesn't make it a female,  defining a gender by clothes the gender is "supposed to wear" is kind of sexist, or at least very stereotypical)

mittalyashu:
Hmm... I see.

PR #305, opened 2021-05-01
add hard-drive crash icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-05-01 at 15 33 05@2x |


mittalyashu:
Any thoughts on this icon? @ericfennis

PR #308, opened 2021-05-01
Add timer-reset icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-05-01 at 15 47 46@2x |


locness3:
Doesn't really feel like resetting it, just like a slight variation.

mittalyashu:
There is a use-case for this, when the user hasn't started the timer at time this icon can be used.

locness3:
Then it needs a different name, or a stronger metaphor

ericfennis:
Maybe we can make it a bit more obvious.
The difference between the regular `timer` icon and this icon is too little for the average user.

Maybe we can go for something like this?

image

ericfennis:
@mittalyashu My proposal for the reset timer icon.

image


100%
Frame

PR #309, opened 2021-05-01
add timer off icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-05-01 at 15 50 24@2x |

### DIFF
--- a/icons/timer-off.svg
+++ b/icons/timer-off.svg
@@ -1,0 +1,16 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M10 2h4" />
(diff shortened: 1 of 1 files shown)


mittalyashu:
@ericfennis any changes require in this PR?

ericfennis:
@mittalyashu Sorry for the late reply.

Something bothers me that the gap along side the slash is on the bottom left instead of the right top we do on other `-off` icons. Did you try to make the gap top-right of the slash?

mittalyashu:
@ericfennis It looks more weird doing it from other side

CleanShot 2021-11-02 at 10 02 22@2x

ericfennis:
@mittalyashu hmmm, and if you remove the seconds hands. Like the alarm-off icon.

Can you post a 100% version maybe it looks ok.

mittalyashu:
| Full scale | 100% |
| --- | --- |
| CleanShot 2021-11-13 at 17 34 30@2x | CleanShot 2021-11-13 at 17 34 54@2x |

ericfennis:
@mittalyashu Hmm I think I'm ok with it. I see it is hard to design the slash with the gap on the other side. So I think we can do this one.

ericfennis (approved):

PR #310, opened 2021-05-01
add robot icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-05-01 at 16 00 03@2x |

### DIFF
--- a/icons/robot.svg
+++ b/icons/robot.svg
@@ -1,0 +1,15 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M2 21v-8C2 7.477 6.477 3 12 3s10 4.477 10 10v8" />
+  <path d="M9 18h6" />
+  <path d="M18 13l-2-2-2 2-2-2-2 2-2-2-2 2" />
+</svg>


locness3:
what's the difference with the `bot` icon?

mittalyashu:
The `bot` icon is different from this one.

ericfennis:
@mittalyashu Can you please give a better explanation.
What makes `bot` different from `robot`? 
Or you just want to submit an other variant from robot>

locness3:
> The `bot` icon is different from this one.

Yeah, i know 😁

mittalyashu:
The use case can be bit different based on the context, for instance.

Bot: Is more of related to programatic and automated
Robot: Can mean a machine which interact with human.

locness3:
:thinking:

mittalyashu:
@ericfennis @locness3 Is that convincing? https://github.com/lucide-icons/lucide/pull/310#issuecomment-833188701

locness3:
no

mittalyashu:
@ericfennis any thoughts?

locness3:
> Robot : Can mean a machine which interact with human.

From what I recall, much apps use a "bot" icon for this purpose.

mittalyashu:
then how about `bot-2`?

ericfennis:
@mittalyashu If I can be honest I don't really like the design. I think it's not fitting with the rest of the icons.
I don't know why. Maybe it's a bit too abstract.
But that's my opinion, if other people like it (give thumbs up for example), I maybe consider to approve it as well.

mittalyashu:
Okay :)

locness3:
This icon doesn't have much approval, so I' m closing this.

PR #311, opened 2021-05-01
add organisation icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-05-01 at 16 13 14@2x |


locness3:
what's the difference over `building` ?

mittalyashu:
The stroke at the back represents multiple buildings.

ericfennis:
Sorry @mittalyashu there is already the building icon representing an `organisation`, so I see this more as a duplicate.

PR #315, opened 2021-05-08
fix: smartphone outer line position

mittalyashu:


icons/smartphone.svg
```diff
@@ -9,6 +9,6 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect x="5" y="2" width="14" height="20" rx="2" ry="2" />
-  <line x1="12" y1="18" x2="12.01" y2="18" />
+  <path d="M17 2H7a2 2 0 00-2 2v16a2 2 0 002 2h10a2 2 0 002-2V4a2 2 0 00-2-2z" />
```
ericfennis, line 12: Can replace this rectangle shape with and `rect` element? 
We can win in some code size.

ericfennis (changes requested):
Can you also post an preview image to show you changes?

icons/smartphone.svg
mittalyashu: Sure

ericfennis (approved):

PR #324, opened 2021-05-15
add mobile power button

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-05-15 at 09 46 07@2x |

### DIFF
--- a/icons/mobile-power-button.svg
+++ b/icons/mobile-power-button.svg
@@ -1,0 +1,16 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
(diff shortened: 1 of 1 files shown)


locness3:
Hmm

mittalyashu:
Something wrong with the icon?

locness3:
I'll be the PR template :
- **What is the purpose of this icon?** : e.g. Shows that one can click it to... / Is used to denote or label...
- **Have you considered alternative possibilities** for its naming or design?

ericfennis:
I think this one is too specific and abstract.

mittalyashu:
My idea for designing this icon was creating a mobile power button.

iPhone power button, for instance.

locness3:
I understood that, but in what context should it be used?

mittalyashu:
IDK, if there will be any copyright issue from Apple

image

locness3:
Oh, makes more sense now. But I think this is a very specific use-case and it would not benefit the icon set as a whole.

mittalyashu:
You never know when and where people might use the icons.

locness3:
They can use another icon set as well.

mittalyashu:
Hmm... okay.

PR #327, opened 2021-05-15
add alphabet-b icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-05-15 at 11 47 44@2x |

### DIFF
--- a/icons/alphabet-b.svg
+++ b/icons/alphabet-b.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M7 3v18" />
+  <path d="M7 3h4c7.374 0 8.603 9 0 9m-4 9h4c8.603 0 7.374-9 0-9m-4 0h4" />
+</svg>


locness3:
@mittalyashu please use the pull request template...

mittalyashu:
Didn't knew we've PR template, I mostly create a PR directly from VS code.

Will do that next time.

MoltenCoffee:
The B feels a bit off for me. It might be the 'comic sans'-y playfullness of the rounded edges, but I think it's mostly the fact that both the upper and lower counter (the open areas inside) are the same size and shape. Most font families will have the upper counter be slightly smaller than the lower one. 
I'm sorry for nitpicking here, but I have the feeling this 'B' is also a lot more 'playful' than the even more geometric 'A' in #258.

In general though, I'm personally not entirely sure there's really a need for letters. In my experience fonts are really a skill in its own right, and require a lot of work to look natural. I'm also somewhat afraid of a slippery slope, would we need underscores, slashes, and punctuation as well? Cyrilic characters?

mikeott:
@MoltenCoffee I agree. The design language is not quite the same as the rest of the iconography.

@mittalyashu Adding a random letter suggests the rest of the English speaking alphabet should also be considered. And if so, what about numerals? And even then, why only English? This is a slippery slope and seems like a pretty good way to add unnecessary bloat.

An icon is a way to graphically represent something symbolic, a concept or an idea. Letters and numbers are what they are - just letters and numbers and don't require graphical representation to explain what they are, hence, can not be considered icons.

mittalyashu:
Hmm... @mikeott you do make a point.

Okay, closing this PR for now.

@ericfennis what should we do about "A" icon already added to the set.

ericfennis:
I agree with @MoltenCoffee and @mikeott. We should avoid adding typographic. Then we can better consider making a font 😄 . 

@mittalyashu Yeah close it as wel.

mikeott:
@mittalyashu If it helps, there is this: https://lucide.dev/icon/bold

mittalyashu:
> @mittalyashu If it helps, there is this: https://lucide.dev/icon/bold

Wait, when did we had that icon 😂

locness3:
lmao, perfect.

fbarikzehi:
I suggest adding a guideline like `Avoid adding typographic icons` to the PR checklist.

PR #328, opened 2021-05-16
Prompt for tags in PR template

locness3:


ericfennis (approved):
Yess nice addition. I will add the missing tags

PR #329, opened 2021-05-16
Update tags for the website

### LINKED ISSUE #314: Recently merged icons need tags again!
Also it should be added in the design guide, that one should submit icons with some tags.

ericfennis:
Closes #314

### DIFF
--- a/tags.json
+++ b/tags.json
@@ -13,7 +13,9 @@
   "alarm-check": [
     "done"
   ],
-  "alarm-clock": [],
+  "alarm-clock": [
+    "morning"
+  ],
   "alarm-minus": [
     "remove"
   ],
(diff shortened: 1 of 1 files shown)


tags.json
```diff
@@ -234,6 +275,11 @@
     "paint",
     "color"
   ],
+  "bug": [
```
locness3, line 278: `insect`?

tags.json
```diff
@@ -348,6 +439,10 @@
     "source",
     "programming"
   ],
+  "code-2": [
```
locness3, line 442: `html` and `xml`? as well for `code-1`

tags.json
```diff
@@ -480,8 +575,12 @@
     "money",
     "payment"
   ],
-  "download": [],
-  "download-cloud": [],
+  "download": [
```
locness3, line 578: *export?

tags.json
```diff
@@ -540,56 +643,78 @@
...
+    "document"
   ],
   "file-plus-2": [
     "add",
     "create",
-    "new"
+    "new",
+    "document"
   ],
   "file-search": [
```
locness3, line 692: `find` as well

locness3 (changes requested):
Some suggestions, didnt go through the whole thing

tags.json
ericfennis: https://en.wikipedia.org/wiki/Insect

ericfennis:
@locness3 updated!

tags.json
locness3: https://en.wikipedia.org/wiki/Bug?wprov=sfla1

> More broadly, any of many small, primarily terrestrial invertebrates, such as Insects

ericfennis: yeah true, but it's just an tag to help people search for icons. 
So any related words or part of something or synonyms are welcome.

locness3 (approved):
lgtm.

Issue #61, opened 2020-08-31
Add a little "context" to the website


locness3:
Maybe something like :
> Featherity is a fork of [Feather Icons](https://feathericons.com) made by the community. We try to expand the icon set as much as possible while keeping it nice-looking - [join us](https://github.com/featherity)!

Issue #116, opened 2020-10-30
Uv Index


cjh9:
Sun with the word UV written inside for use for UV index in weather map

Andreto:
@cjh9 Does any of these matches what you are looking for?
![image](https://user-images.githubusercontent.com/28541538/98152366-e4c44c00-1ed1-11eb-9e4f-c618d34f0a7a.png)

ericfennis:
Nice I like the right one. How does it look if the text: `uv` is centered?

Andreto:
@ericfennis
![image](https://user-images.githubusercontent.com/28541538/98443628-c5fdc980-210c-11eb-8a76-98ab2d3c45d8.png)

ericfennis:
hmm, not sure yet.
Can you try to make the spacing in the letter U to be 2 px?

Also, I think we need to refactor the sun icon, it doesn't have the 1px safezone.
Can you try to make one with this icon?
![image](https://user-images.githubusercontent.com/11825403/98445700-5f32dd00-2119-11eb-9cd6-68546aca067b.png)
```svg
<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
<path d="M12 16C14.2091 16 16 14.2091 16 12C16 9.79086 14.2091 8 12 8C9.79086 8 8 9.79086 8 12C8 14.2091 9.79086 16 12 16Z" stroke="black" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
<path d="M12 2V4" stroke="black" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
<path d="M12 20V22" stroke="black" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
<path d="M5.21997 5.21997L6.63997 6.63997" stroke="black" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
<path d="M17.36 17.36L18.78 18.78" stroke="black" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
<path d="M2 12H4" stroke="black" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
<path d="M20 12H22" stroke="black" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
<path d="M5.21997 18.78L6.63997 17.36" stroke="black" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
[...]

PR #331, opened 2021-05-17
Add tent icon

MoltenCoffee:
name | about | labels
-- | -- | --
New icon | Add a new icon to the library | 🎨 <icon

- **Name of the icon** : Campsite
- **Tags (alternative names for this icon)** (add them in tags.json) :  tent
- **What is the purpose of this icon?** : Depicting holidays or outdoor activities. Can be used in maps as well.
- **100% scale preview** : campsite
- **Have you considered alternative possibilities** for its naming or design? : It's roughly based on [#61 at feather](https://github.com/feathericons/feather/issues/61), but I feel this one is slightly more recognizable as a tent.

### DIFF
--- a/icons/campsite.svg
+++ b/icons/campsite.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M3 20h18M5 20l7-13 7 13M10 4l2 3 2-3" />
+  <path d="M9 20l3-5 3 5" />
+</svg>
--- a/icons/tent.svg
+++ b/icons/tent.svg
@@ -9,6 +9,9 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M3 20h18M5 20l7-13 7 13M10 4l2 3 2-3" />
-  <path d="M9 20l3-5 3 5" />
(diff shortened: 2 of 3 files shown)


ericfennis:
@MoltenCoffee Nice icon, thanks for the contribution!

Can you split up the path in separate shapes?
Because some design programs don't like combined/merged paths. 

If you need any help with that let me know!

MoltenCoffee:
Sure, of course! Which parts would need to be separate paths? I quickly 'colored' the icon below for reference, I hope you're able to see it.

Untitled

I'd assume it would have to be split in the following paths:

- red (ground)
- both greens (outside of the tent)
- blue and brown (the entrance)
- pink and yellow (the rods at the top) (these are at a slightly different angle compared to the tent itself)

MoltenCoffee:
Also, I named it "campsite" as that is how it is listed in #116, but I think "tent" with "campsite" as an alias is probably better. Do you agree?

ericfennis:
@MoltenCoffee Yeah `tent` is perfect. We prefer naming as is what it shows than what it means. Because icon can multiple meanings.

For the splitting paths, I think we can split it up in 4 paths.

1. red
2. Brown
3. Green#1 with extended line to pink
4. Green #2 with extended line to yellow

I think the change of the paths that have a different angle (pink and yellow) is not visible when seeing this icon on 100%.

MoltenCoffee:
I've replaced all paths, and like you said the difference in angles are negligible when viewed around 24px and way beyond. I've also renamed it to 'tent'.

mittalyashu:
@ericfennis Once you merge this PR, can you update icon status in this issue as well https://github.com/lucide-icons/lucide/issues/119

ericfennis:
@mittalyashu Will do!

PR #341, opened 2021-05-31
add reply-icons

Andreto:
#128


Andreto:
Tags could not be updated without conflict.

ericfennis (approved):
Awesome work, thank you!

PR #351, opened 2021-06-19
create wand icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-06-19 at 13 00 06@2x |

### DIFF
--- a/icons/wand.svg
+++ b/icons/wand.svg
@@ -1,0 +1,20 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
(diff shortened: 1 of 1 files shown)


ericfennis:
@mittalyashu Nice icon!
I think the horizontal right icon  is a bit shifted to the left.

locness3:
Lacks some spacing IMO, doesn't look very Lucidey. (please use the pull request template)

mittalyashu:
The size of the wand is 12px in width & height, i don't see it shifted.

mittalyashu:
Only at the top there is 1px spacing.

JanTrichter:
> 
> 
> The size of the wand is 12px in width & height, i don't see it shifted.

I think they mean this part of the wand which is different from the other parts.
grafik

ericfennis:
Exactly @JanTrichter

ericfennis:
my pruposal:
wand 24px

wand

PR #355, opened 2021-07-02
Build multiple font weights in action

### LINKED ISSUE #354: Request: Multiple font weights for web fonts
It would be nice to have thin/medium/thick versions of the webfonts available.

I think this could be done in outline_svg.js by changing the stroke width before calling outline stoke, and then generating the fonts for each width.

Billiam:
Fixes #354 

This PR creates a build artifact with fonts in several different weights. `fontcustom` doesn't support any kind of font weight combining that I see, so if you wanted to use multiple weights in the same project, you'd need to create a merged CSS file with the additional font faces you need.

The default weight (2) has no name suffix, while the others (1, 1.5, 2.5, and 3) use filenames that might match their expected css font-weights.

Shown: 200, 400 (the default), and 600 weight:
New Project

Artifact output: https://github.com/Billiam/lucide/suites/3139790111/artifacts/72098979

### DIFF
--- a/.github/workflows/font.yml
+++ b/.github/workflows/font.yml
@@ -50,10 +50,11 @@ jobs:
 
     - name: Build 'Lucide'
       run: |
+          mkdir build
           list=(_200 _300 "" _500 _600)
           for name in "${list[@]}"
           do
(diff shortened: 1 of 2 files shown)


ericfennis:
Nice work!

ericfennis (approved):

PR #356, opened 2021-07-12
Allow to use with React 17

Aviortheking:
The library works perfectly with React 17 so it will allow users to not use `npm --legacy-peer-deps`


ericfennis:
@Aviortheking Thanks, didn't know you need to do that on react 17. Thanks for the addition!

ericfennis (approved):

PR #357, opened 2021-07-14
Add info about laravel blade icons repackage

mallardduck:


ericfennis:
@mallardduck Sorry for the late reply. I was on vacation.

Thanks for this addition! 
Very nice to see other packages adopting Lucide!

locness3:
Isn't the Usage section in the readme otherwise outdated though?

mallardduck:
@locness3 - from what I can see the Usage section for this icon package is not outdated.
All of those use-case specific packages seem to exist and are ready for use by anyone.

Can you provide an example of where you feel something is outdated?

PR #368, opened 2021-08-04
Add alternative home icon

Tummerhore:
Preview:
Unbenannt

### DIFF
--- a/icons/home-2.svg
+++ b/icons/home-2.svg
@@ -1,0 +1,13 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M2 11l10.053-9.048L22.106 11"/>
+  <path d="M19 8.563V22h-4v-9H9v9H5V8.563"/>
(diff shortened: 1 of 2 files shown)


ericfennis:
@Tummerhore Welcome to the Lucide Community!

Thanks for opening this.
This icon is very similar to the current [`home`](https://lucide.dev/icon/home) icon.
Can you maybe explain why this is a good addition?

locness3:
Did you update the preview along with your commits?

Tummerhore:
> Did you update the preview along with your commits?
Yes. The preview schows the latest iteration.

> This icon is very similar to the current home icon.
Yeah, you're right. It's just personal taste. While the original icon is ok, I did not like it that much. Maybe some people share that opinion so I made an alternative icon. But since some other people might like the original icon more I did not want to change the original one.
I think it's not wrong to have more alternatives. (Other icon projects like font awesome for example also have a lot of alternatives for one icon.)

ericfennis:
@Tummerhore Good explanation! 
Personally I like the current home icon, but indeed maybe there are more people like to change this.
So for now I keep this one open and if there are more upvotes, we will consider merging this.

delnyn:
I personally think that adding this icon isn't a great idea - it's not really useful

ericfennis:
@delnyn You're right. I'm closing this one for now.

Issue #352, opened 2021-06-20
Illustrator Icon Design Template


JanTrichter:
Hey,
I just created a small Illustrator template which could help new designers to better and faster implement the [icon design guidelines](https://github.com/lucide-icons/lucide/blob/master/docs/ICON_DESIGN_GUIDE.md).

I uploaded it here: [https://icedrive.net/0/94vskayEo6](https://icedrive.net/0/94vskayEo6)

I adjusted the pen tool settings and line thickness to match the requirements of the icon design guide, added a small grid and marked the recommended margin in the document. I also added an icon which demonstrates how this template could be used.

Could you check if it fulfils all the requirements from the icon design guide and tell me if you have other feedback?

Have a great day 😎

ericfennis:
@JanTrichter Sorry for my late reply, I was off for a bit.

I checked the template, it looks great. All the the guidelines are there and nice you added the padding.
I wasn't sure about the icon you placed in the center looks a bit random but that's me.

To finish I think we can add this to the repo if it's ready, so people can download it to use it.
Overall thanks for doing this @JanTrichter , awesome work.

JanTrichter:
> I checked the template, it looks great. All the the guidelines are there and nice you added the padding.
> I wasn't sure about the icon you placed in the center looks a bit random but that's me.

Yeah, it is a bit random but it should demonstrate all the guidelines that you cannot see on the template itself (distance between lines, round corners, etc.)

delnyn:
So you could put this file in the repo and link it in the docs (e.g. docs/ICON_DESIGN_GUIDE.md) ?

JanTrichter:
> So you could put this file in the repo and link it in the docs (e.g. docs/ICON_DESIGN_GUIDE.md) ?

I have a lot of projects lately so that could take some time :(

delnyn:
No no I meant both of you (I could do it myself too)

JanTrichter:
> No no I meant both of you (I could do it myself too)

Got Ya!
Made a PR with the template and updated the according parts in the documentation! ( #377 )

delnyn:
Great! Thanks a lot

JanTrichter:
Done in #377

PR #377, opened 2021-08-09
Added the template and changed corresponding documentation

JanTrichter:
Added the illustrator template and created (`ILLUSTRATOR_GUIDE.md`) or changed (`CONTRIBUTING.md`) files to reflect changes in the documentation.

**Added:**

- added Adobe Illustrator template (`/docs/templates/illustrator-template.ai`)
- added documentation for the template (`/docs/ILLUSTRATOR_TEMPLATE.md`)

**Changed:**

- changed the `ICON_DESIGN_GUIDE.md` to have a reference to the Inkscape and Illustrator Guide as well as the Illustrator template

This fixes issue #352


docs/ILLUSTRATOR_GUIDE.md
```diff
@@ -0,0 +1,21 @@
...
+
+2. You can now remove the content from the example logo layer ("Draw") and start creating.
+
+3. Verify that you follow the [Icon Design Guidelines](ICON_DESIGN_GUIDE.md).
+
+4. When you are ready to export the file as an SVG make sure to check that you followed the guidelines and remove all unecessary layers (especially "Padding" and "Grid").
+
+5. Double check that the [code conventions and SVG global attributes](https://github.com/lucide-icons/lucide/blob/master/docs/ICON_DESIGN_GUIDE.md#code-conventions) are correct.
+
+6. Minify paths with [SVGOMG](https://jakearchibald.github.io/svgomg/).
```
ericfennis, line 21: Maybe nice to write also something about exporting designs to SVG.
Illustrator has an excellent exporting tool, called "Export as" in the file menu.
Here is an article about it: [Export as article - CSS Tricks](https://css-tricks.com/illustrator-to-svg/#the-export-as-method)

CONTRIBUTING.md
```diff
@@ -20,10 +20,11 @@ Guidelines for pull requests:
 
 ### Pull Requests Including Icons
 
-#### Guidelines
+#### Guidelines & Templates
 
 Please make sure you follow the icon guidelines, that should be followed to keep quality and consistency when making icons for Lucide
-Read it here: [ICON_GUIDELINES](docs/ICON_DESIGN_GUIDE.md)
+
+Read it here: [ICON_GUIDELINES](docs/ICON_DESIGN_GUIDE.md). You can find a template and instructions on how to implement those guidelines with Adobe Illustrator and Inkscape in the */docs* folder of the project.
```
locness3, line 27: Would've made a seperate section for instructions relating to particular software.

locness3 (changes requested):

docs/ILLUSTRATOR_GUIDE.md
JanTrichter: Does the minify option in the export menu do the same thing as SVGOMG? Would be important to know as that could avoid redundancies in the workflow.

What would you recommend as standard export settings?
I would say: 

- **Style**: `internal CSS` (easier to overwrite and remove)
- **Font**: probably doesn't matter but I would go with `Convert to Outlines` (font independent but larger)
- **Images**: don't know if we need to change that
- **Object-IDs**: I would go with `unique` to avoid conflicts
- **Decimal**: `2` is fine
- **minify**: `yes` cause why not. Seems to work properly
- **responsive**: `no` because width and height are required attributes according to the [IDG - Global Attributes](https://github.com/lucide-icons/lucide/blob/master/docs/ICON_DESIGN_GUIDE.md#code-conventions)

CONTRIBUTING.md
JanTrichter: Agree. 
Adjusted in the latest commit 👍🏽

docs/ILLUSTRATOR_GUIDE.md
ericfennis: Yeah, SVGOMG is the best, but illustrator can compress a lot as well. I would say do both

The preferences matching the best with lucides codestyle are: 
- **Style**: `Presentation Attributes`
- **Font**:  `Convert to Outlines`, 
- **Images**: no config, we don't accept images
- **Object-IDs**: `none` (if possible)
- **Decimal**: `2` or less depending if curves are used.
- **minify**: `yes` cause why not. Seems to work properly
- **responsive**: `no`

JanTrichter: image
so those are the best settings.

ericfennis (approved):
Awesome work @JanTrichter !
Look good to me.

PR #380, opened 2021-08-10
Created selection-flag start and end flag

tidoni:
I use it for selecting the starttime and endtime in an audio file.


ericfennis:
Nice icon!

ericfennis (changes requested):
Seems like there is a random square in the svg.
The rect element can be removed.
Can you also add a new line add the end of the files.

icons/selection-flag-end.svg
```diff
@@ -0,0 +1,14 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect width="100" height="100" x="50" y="50" />
+  <path d="M17 22 17 2 7 7 17 12"></path>
+</svg>
```
ericfennis, line 14:
```suggestion
  <path d="M17 22 17 2 7 7 17 12" />
</svg>

```

icons/selection-flag-start.svg
```diff
@@ -0,0 +1,14 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect width="100" height="100" x="50" y="50" />
+  <path d="M7 22 7 2 17 7 7 12"></path>
+</svg>
```
ericfennis, line 14:
```suggestion
  <path d="M7 22 7 2 17 7 7 12" />
</svg>

```

locness3 (changes requested):
What about `triangle-flag-{left,right}`?

ericfennis:
I agree with @locness3, with generalizing the name. We make it more multy-purpose. Fits well with our other names.
I like `flag-triangle-{left,right}` more, because `flag` is the icon and `triangle` is the variation.

PR #388, opened 2021-08-15
Fix typescript declarations for lucide-react

knoan:
- removed leftover references to `vue`
- removed redundant property declarations on `LucideProps`


ericfennis:
@knoan Thanks for this, I accidentally changed this type generator when building 'lucide-vue-next' .

PR #421, opened 2021-10-11
✅ Add tests for flutter package and Formatted files

### LINKED ISSUE #417: Write tests for Flutter package
To make sure the flutter package will work, we need to add tests for the package.
In every other package we added tests to make sure the packages will work.

prateekmedia:
- `pubspec.lock` is deleted as it is not needed for flutter packages.
- Add flutter tests for icons before publishing to release (Fix #417)
- Formatted `.dart` files and `.yaml` workflow files (most of the time quotes[ " " or ' ' ] are not needed in yaml so I removed them)


ericfennis:
@prateekmedia  Sorry wasn't able to test it yet, but I just did. It works perfectly. Thanks!

PR #423, opened 2021-10-13
Added chevron-first and chevron-last

dperezcabrera:


locness3:
I think I am okay with the icon names.

PR #428, opened 2021-10-23
Add `signal*` icons

### LINKED ISSUE #283: Icon Request: Signal level
Name : Signal 1,2,3,4, no-signal, signal-research
Is it possible to have something like that : 
signal-indicator

Thank you for your help

ericfennis:
closes #283
image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -350,6 +350,7 @@ jobs:
 
       - name: Test
         run: flutter test
+        working-directory: packages/lucide-flutter
 
       - name: Update yaml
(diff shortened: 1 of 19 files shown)


mittalyashu:
feedback: any better way to make it distinct from `bar-chart`

CleanShot 2021-11-02 at 10 03 45@2x

mittalyashu:
@ericfennis how about something like this?

CleanShot 2021-11-13 at 18 01 45@2x

ericfennis:
@mittalyashu 
I think it is too thick. The lucide style is, thin and simplistic. 
I don't really bother me that the icon is a bit similar with bar-chart. The signal icon is to be fair a sort of bar chart icon displaying the strength of the signal.

maxim-s-barabash:
I'm not sure if the icon "signal-zero" is correct

Screenshot from 2022-04-14 18-01-47

ericfennis:
@@maxim-s-barabash you right. This need a fix

Issue #427, opened 2021-10-20
Update dependencies for Angular v12


angelinuxx:
I would love to see updated dependencies for Angular v12 in lucide-angular package.

Thanks in advance

ericfennis:
Thanks I will update them soon!

ericfennis:
Fixed by #432

PR #432, opened 2021-11-10
Support Angular version 11 or higher

SMAH1:
I test in angular v11, v12 and v13.
#427


ericfennis (approved):
Awesome @SMAH1 Thanks!

PR #433, opened 2021-11-11
Use 'angular build workflow' in Windows

SMAH1:
I use windows and linux.
Befor, for compile lucide-angular  use linux only. Now I changed it for work in windows and linux .
I test it in windows and ubuntu.

### DIFF
--- a/packages/lucide-angular/package.json
+++ b/packages/lucide-angular/package.json
@@ -23,7 +23,7 @@
   ],
   "scripts": {
     "build": "yarn clean && yarn build:icons && yarn build:ng",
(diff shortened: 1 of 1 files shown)


ericfennis (commented):
Can we not better use rimraf for this?
https://github.com/isaacs/rimraf

SMAH1:
I see it when I decide use 'shelljs' or 'rimraf'.
I use 'shelljs' beacause: 

- More than star
- It has more up-to-date commit
- We use in other project (lucide-vue, ...) for other unix shell command

ericfennis (approved):
@SMAH1 Ah thanks for your explanation. Then it is good to go!

PR #436, opened 2021-11-13
Refactoring

SMAH1:
Some refactoring when build typing for of lucide

### DIFF
--- a/packages/lucide/scripts/buildTypes.js
+++ b/packages/lucide/scripts/buildTypes.js
@@ -10,9 +10,11 @@ const TYPES_FILE_NAME = 'lucide.d.ts';
 // Generates header of d.ts file include some types and functions
 const typeDefinitions = `\
 declare module 'lucide'
+
 export interface SVGProps extends Partial<SVGElement> ${JSON.stringify(defaultAttributes, null, 2)}
 
-export declare type IconNode = readonly [tag: string, attrs:SVGProps, children?:IconNode];
+export declare type IconNodeChild = readonly [string, object];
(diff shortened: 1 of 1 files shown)


packages/lucide/scripts/buildTypes.js
```diff
@@ -10,9 +10,11 @@ const TYPES_FILE_NAME = 'lucide.d.ts';
 // Generates header of d.ts file include some types and functions
 const typeDefinitions = `\
 declare module 'lucide'
+
 export interface SVGProps extends Partial<SVGElement> ${JSON.stringify(defaultAttributes, null, 2)}
 
-export declare type IconNode = readonly [tag: string, attrs:SVGProps, children?:IconNode];
+export declare type IconNodeChild = readonly [string, object];
```
ericfennis, line 16: These can be tags and atrts as well.
```suggestion
export declare type IconNodeChild = readonly [tag: string, attrs: SVGProps];
```

ericfennis (changes requested):
Awesome thanks for doing this!

packages/lucide/scripts/buildTypes.js
SMAH1: 'SVGProps' is type variable of 'defaultAttribute' .
Each icon has:

1. tag
2. defaultAttring
3. data

and data is:

1. tag
2. svg object data

so 'svg object data' is not 'SVGProps'.
I think use:

``` ts
export declare type IconNodeChild = readonly [tag: string, attrs: object];
```

ericfennis: Sorry my bad, I thought it was extending from Native SVG type

ericfennis (approved):

PR #437, opened 2021-11-13
icon: iMac

mittalyashu:
- **Name of the icon** : iMac
- **Tags (alternative names for this icon)** : _will do after the approval of icon_  

| Full scale | 100% |
| --- | --- |
| CleanShot 2021-11-13 at 17 32 56@2x | CleanShot 2021-11-13 at 17 45 06@2x |

### DIFF
--- a/icons/iMac.svg
+++ b/icons/iMac.svg
@@ -1,0 +1,13 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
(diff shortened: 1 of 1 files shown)


ericfennis:
Something else,  I'm not sure if we should make icons for specific brands/devices.
@locness3 What do you think?

locness3:
> Something else, I'm not sure if we should make icons for specific brands/devices.

Definitely NOT.

ericfennis:
@locness3 Allrighty.
@mittalyashu Sorry we close this one. We should not make icons for specific brands/devices.

mittalyashu:
Ignoring the name "iMac", it can be considered as another display icon.

PR #438, opened 2021-11-13
feat: pin icon

mittalyashu:
- **Name of the icon** : Pin
- [x] - Tags (alternative names for this icon)

| Full scale | 100% |
| --- | --- |
| CleanShot 2021-11-16 at 21 31 42@2x | CleanShot 2021-11-16 at 21 31 49@2x |


mittalyashu:
If anyone has any suggestion for un-pin icon, let me know. I came up with these two but didn't liked both 😞

CleanShot 2021-11-13 at 17 53 55@2x

ericfennis:
@mittalyashu Nice icon!! I like it.
I think the needle feels a bit too long. I see also that there is 2 pixels margin on the top and only 1 pixel at the bottom. Maybe decrease the needle one pixel on the bottom so the margins are equal top and bottom.

tags.json
```diff
@@ -1410,6 +1410,9 @@
     "statistics",
     "diagram"
   ],
+  "pin": [
+    "pin"
```
ericfennis, line 1414: An extra tag with the same content as the name is useless.

```suggestion
    "save",
    "map",
    "lock",
    "fix"
```

bduffany:
@mittalyashu I like the second unpin icon -- it seems to match the other "un-" / "-off" styles of the other icons, with the strikethrough connecting to the bottom-left of the icon, and the gap to the upper-right of the strikethrough.

mittalyashu:
I also don't like leaving this small part, which looks kind of weird.

CleanShot 2021-11-16 at 21 11 00@2x

mittalyashu:
> I think the needle feels a bit too long. I see also that there is 2 pixels margin on the top and only 1 pixel at the bottom. Maybe decrease the needle one pixel on the bottom so the margins are equal top and bottom. @ericfennis 

fixed! Thanks for pointing it out.

ericfennis:
@mittalyashu this one feels better!

mittalyashu:
LOL! I literally forget to `git push` the commit 🤣

icons/pin.svg
```diff
@@ -0,0 +1,13 @@
...
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M9 3v5l-2 4v2h10M9 3h6M9 3H7m10 11h-5v7m5-7v-2l-2-4V3m0 0h2" />
```
ericfennis, line 12: Can you split this path up in smaller parts. Otherwise design programs like figma don't like it.

ericfennis (changes requested):

icons/pin.svg
mittalyashu: sure

mittalyashu: like this? Please check the most recent commit.

ericfennis:
I think you need to split it once more. maybe the top stroke?
image

zamkevich:
> I also don't like leaving this small part, which looks kind of weird.
> 
> CleanShot 2021-11-16 at 21 11 00@2x

Left have a correct shadow

icons/pin.svg
```diff
@@ -0,0 +1,14 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M12 21v-7" />
+  <path d="M9 3v5l-2 4v2h10M9 3h6M9 3H7m10 11h-5m5 0v-2l-2-4V3m0 0h2" />
```
ericfennis, line 13: The fix

```suggestion
  <path d="M12 21v-7" />
  <path d="M9 3v5l-2 4v2h10m0 0h-5m5 0v-2l-2-4V3" />
  <path d="M7 3h10" />
```

ericfennis:
@mittalyashu Can you take a look, looks like you have some conflicts

locness3:
Would be great to have a tilted version as well.

mittalyashu:
> Would be great to have a tilted version as well.

@locness3 What do you mean by titled version?

ericfennis (approved):
Awesome!

ericfennis:
@mittalyashu I assume @locness3 thinks about this kind of icon.
image

PR #445, opened 2021-11-26
feat: add egg icon

mittalyashu:
| Zoom | 100% |
| --- | --- |
|  CleanShot 2021-11-26 at 17 41 18@2x |  CleanShot 2021-11-26 at 17 41 34@2x |

Credit goes to https://github.com/feathericons/feather/pull/83


ericfennis:
@mittalyashu hmm I'm not sure about the gloss line. I think without look a bit better with the others.

mittalyashu:
@ericfennis here's a comparison between egg with/without gloss line.

CleanShot 2021-12-26 at 21 13 26@2x

Andreto:
@mittalyashu  Isn't it a bit too pointy? Real life egg for reference:
image

ericfennis:
I prefer the egg without gloss, but that is my preference. What do you think?

mittalyashu:
@Andreto that's true, just by creating and outline shape it will only look like an oval shape without giving a clear visual representation of an egg.

@ericfennis Sure, we can remove the gloss. But the image @Andreto shared does have a gloss 😂

Andreto:
@mittalyashu I realize that your more exaggerated shape does look a lot better in true size.
image

@ericfennis Could the gloss perhaps work better if it was a bit smaller and actually followed the shape of the edge?
image

mittalyashu:
After comparing all four, I like the 4th one.

CleanShot 2021-12-29 at 22 15 42@2x

ericfennis:
@Andreto Yeah it is more about the gloss, we don't use it on other icons for example `camera` or `smartphone`. So that why I don't feel something for the gloss.

@mittalyashu I agree with you, 4th looks better!

mittalyashu:
@Andreto Since you've the source code of that 4th icon. You can override the changes to this PR or create a new one or can you share the source code?

Andreto:
@mittalyashu I was unable to push a commit to this pr. Could you perhaps turn on “Allow edits from maintainers.” ?

mittalyashu:
> I was unable to push a commit to this pr. Could you perhaps turn on “Allow edits from maintainers.” ?

@Andreto It's already enabled.

ericfennis (approved):

PR #448, opened 2021-11-27
feat: Add `clock-*` icons

wojtekmaj:
https://user-images.githubusercontent.com/5426427/143689259-76f54ca1-ca12-4077-9515-f05dc1e3221c.mov

image

### DIFF
--- a/icons/clock-4.svg
+++ b/icons/clock-4.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="12" cy="12" r="10" />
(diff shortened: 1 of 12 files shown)


icons/clock-4.svg
```diff
@@ -0,0 +1,14 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="12" cy="12" r="10" />
+  <polyline points="12 6 12 12 16 14" />
+</svg>
```
wojtekmaj, line 14: I'm thinking, this icon is identical to existing Clock icon, but it doesn't make a lot sense to have clock-1, 2, 3 and then 5, 6, 7.

ericfennis: hmm yeah that's true. But replacing `clock` with `clock-4` is not an option.

wojtekmaj: Let's keep it then?

ericfennis: yess I think that's better than removing one or the other

PR #455, opened 2021-12-15
added gizmo icon

pandrr:
Screenshot 2021-12-15 at 14 46 16

a 3d gizmo icon

### DIFF
--- a/icons/gizmo.svg
+++ b/icons/gizmo.svg
@@ -1,0 +1,16 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+<path class="st0" d="M20.7 19.6 12 14.5l-8.7 5.1M12 4.4v10.1" />
+	<circle class="st1" cx="12" cy="4.4" r="1.9" />
+	<circle class="st1" cx="3.3" cy="19.6" r="1.9" />
+	<circle class="st1" cx="20.7" cy="19.6" r="1.9" />
+</svg>


mittalyashu:
Google search results shows this on "gizmo" (sorry, poor joke) 😂

image

pandrr:
haha maybe try "3d cursor gizmo"

Screenshot 2021-12-29 at 21 33 50

this is how the icon looks in blender for example:
Screenshot 2021-12-29 at 21 34 51

mittalyashu:
@ericfennis instead of fill circle, do you think we should try with outline circle?

delnyn:
that's a great idea but the circles definitely break the guidelines

delnyn:
I agree,  
also circles would look too big

karsa-mistmere:
image
```svg
<svg
  width="24"
  height="24"
  viewBox="0 0 24 24"
  fill="none"
  stroke="currentColor"
  stroke-width="2"
  stroke-linecap="round"
  stroke-linejoin="round"
>
  <circle cx="5" cy="19" r="2"/>
  <path d="M5 17V3M3 5l2-2 2 2M7 19h14M19 17l2 2-2 2M6.5 17.5 17 7M14 7h3v3"/>
</svg>
```

lscheibel:
Is there really a need for a general "gizmo-3d" icon? In most 3d applications gizmos have one of three functionalities: move/translate, scale and rotate, each with their own respective icon usually just represented by their 2d equivelent.

The blender icon used as an example actually represents the relative coordinate system used when applying transforms and is unrelated to the gizmos themselves.

I prepared a rough sketch of the transform operations represented as gizmos but the upper two could also just be used for 3d coordinate systems.

image

delnyn:
that's an interesting proposition and the icons look stunning, however, the squares from the `gizmo-scale` icon look like they don't respect the guidelines

ericfennis:
@lscheibel Thanks a lot for checking this, use full information.
Also the icons looks good. 
Maybe Idea, try adding the cricle from @karsa-mistmere icons as center point?

ericfennis:
@lscheibel Can you open an PR for these icons?

lscheibel:
@ericfennis Just submitted them, sorry I'm a bit short on time again. Btw, feel free to just take any screenshot / svg from any issue and take it, I probably won't be that reactive here for the next weeks.

ericfennis:
Fixed by #705

PR #457, opened 2021-12-22
Icons: Carrot & Drumstick

Andreto:
image
image


ericfennis:
Nice I love the style and simplicity.
Can you give small description for what kind of use cases we can use these icons?

Andreto:
@ericfennis I created these for use on a website for my school's lunch menu. I wanted some icons to represent the ordinary meal and the vegetarian alternative.

ericfennis:
@Andreto Thanks for sharing, I think this is great use case. I can imagine that you can use this as well in restaurant websites or "menu" applications as well.

ericfennis (approved):

icons/drumstick.svg
```diff
@@ -0,0 +1,14 @@
+<svg
```
locness3, line 1: Is this supposed to be called a "drumstick"? @Andreto

Andreto: I think that's what it's called but of i have a better name it then we could change it.

PR #464, opened 2022-01-01
feat: add egg icon (updated design)

Andreto:
Some design changes for the egg icon, discussed in #445.


Andreto:
@mittalyashu @ericfennis

ericfennis:
@Andreto I merged this one in to #445 so we keep everything in one thread.

PR #470, opened 2022-01-11
refactor: running optimize script blindly (not literally)

mittalyashu:


ericfennis (approved):
Thanks, I tested some icons in figma. If they don't break. I changed one. But the rest is good. Tnx!

PR #472, opened 2022-01-12
Add React Native package

henrykuzmick:
This PR adds a package for React Native.

I hope someone from the maintainers could help me out, as I don't know how to deal with yarn workspaces or how to prepare this package for npm.

This is what running `npm run prepublish` does atm:
- Uses SVGR to convert existing icons to `react-native-svg` compatible .tsx files
- Converts them to .js files and creates typings from the existing .tsx files.

The users of the package should have `react-native-svg` installed.


ericfennis:
@henrykuzmick Thank you looks good. Didn't know about svgr. I works really nice.
I'm curious did you know that we use our own parser for svgs to components?

For this implementation I miss some things. we like to have a similar api for all the packages. 
The regular react packages has one extra prop: `size` this is bind on the width and the height.
Is it possible to add the 'size' prop as well to these components?

Here an example from the react package.
``` js
export default (iconName, iconNode) => {
  const Component = forwardRef(
    ({ color = 'currentColor', size = 24, strokeWidth = 2, ...rest }, ref) =>
      createElement(
        'svg',
        {
          ref,
          ...defaultAttributes,
          width: size,
          height: size,
          stroke: color,
          strokeWidth,
          className: `lucide lucide-${toKebabCase(iconName)}`,
          ...rest,
        },
        iconNode.map(([tag, attrs]) => createElement(tag, attrs)),
      ),
  );

  Component.propTypes = {
    color: PropTypes.string,
    size: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
    strokeWidth: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
  };

  Component.displayName = `${iconName}`;

  return Component;
};
```

And do we need to have support for commonJs as wel or are ES Modules fine for react native development?

ericfennis:
fixed by #687 with the same api as other packages as prefered.

PR #485, opened 2022-01-24
Allow children in React and Preact icons

### LINKED ISSUE #477: [Feature Request] Accessibility Improvement / Tooltip Support
I shifted to Lucide from Feather after having no way to add tooltips directly. A similar issue (https://github.com/feathericons/react-feather/issues/67) exists for Feather which references this project but I didn't see any request or any implementation here yet. The gist of this issue/request is the ability to add alternative text like this:

`<Icon alt="Navigate to Report" />`

This would allow browsers to display a tooltip when you hover the cursor over the icon like this:

image

The linked issue goes into more detail. The icons work really well as buttons and this addition could improve the accessibility of the websites using these icons directly.

MoltenCoffee:
This will allow adding children to the SVG icons in React and Preact, thus allowing important a11y features like `<title>` tags, and of lesser importance also allowing custom additions to icons.

Both the Preact and React version place new children _before_ existing elements within the SVG. This is to comply with [this caveat found here on MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) to be backward compatible with SVG 1.1 (if this is a non-issue and other arguments can be made for adding new elements _after_ existing children, this can be reversed).

To avoid errors with destructuring non-existent Arrays, the Preact version uses `Preact.toChildArray()` (available in the Preact version required by the package) and the React version `children || []`.

Edit: I see this has (recently) been mentioned in an issue before as well: fixes #477

### DIFF
--- a/packages/lucide-preact/src/createPreactComponent.js
+++ b/packages/lucide-preact/src/createPreactComponent.js
@@ -1,4 +1,4 @@
-import { h } from 'preact';
+import { h, toChildArray } from 'preact';
 import defaultAttributes from './defaultAttributes';
 
 /**
@@ -12,7 +12,7 @@ import defaultAttributes from './defaultAttributes';
 export const toKebabCase = string => string.replace(/([a-z0-9])([A-Z])/g, '$1-$2').toLowerCase();
 
 export default (iconName, iconNode) => {
-  const Component = ({ color = 'currentColor', size = 24, strokeWidth = 2, ...rest }) =>
+  const Component = ({ color = 'currentColor', size = 24, strokeWidth = 2, children, ...rest }) =>
     h(
       'svg',
       {
@@ -24,7 +24,7 @@ export default (iconName, iconNode) => {
         class: `lucide lucide-${toKebabCase(iconName)}`,
         ...rest,
       },
-      iconNode.map(([tag, attrs]) => h(tag, attrs)),
+      [...iconNode.map(([tag, attrs]) => h(tag, attrs)), ...toChildArray(children)],
     );
 
   Component.displayName = `${iconName}`;
--- a/packages/lucide-preact/src/createPreactComponent.js
+++ b/packages/lucide-preact/src/createPreactComponent.js
@@ -24,7 +24,7 @@ export default (iconName, iconNode) => {
         class: `lucide lucide-${toKebabCase(iconName)}`,
         ...rest,
       },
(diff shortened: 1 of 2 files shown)


packages/lucide-preact/src/createPreactComponent.js
```diff
@@ -24,7 +24,7 @@ export default (iconName, iconNode) => {
         class: `lucide lucide-${toKebabCase(iconName)}`,
         ...rest,
       },
-      iconNode.map(([tag, attrs]) => h(tag, attrs)),
+      [...toChildArray(children), ...iconNode.map(([tag, attrs]) => h(tag, attrs))],
```
ericfennis, line 27: Is it not better to spread the children after the iconChildren, so we can render things on top of the icon?
Like I discussed this feature in #477

ericfennis (commented):
@MoltenCoffee Awesome work!! Thanks a lot!

packages/lucide-preact/src/createPreactComponent.js
MoltenCoffee: That depends. As you can see from the commits, that was my initial idea as well. MDN however recommends putting `<title>` elements (which were my primary motivation to add this feature) at the top of `<svg>` elements, as this [is required by spec in SVG 1.1](https://www.w3.org/TR/SVG11/struct.html#TitleElement), otherwise browsers will not show tooltips:

```
In conforming SVG document fragments, any ‘title’ element should be the first child element of its parent. Note that those implementations that do use ‘title’ to display a tooltip often will only do so if the ‘title’ is indeed the first child element of its parent.
```

I do believe however, this is no longer the case in the newer iterations of SVG (but haven't been able to confirm this).

MoltenCoffee: Thanks for the extensive testing! If it's this widely supported, I think the benefits of icon customization options outweigh the backward compatibility with an old SVG standard - especially taking into account it's a new feature to this package so it doesn't break existing implementations depending on it.

I don't know if you're able to do so, otherwise I'll revert the latest commit so it's in the proposed order tomorrow (GMT+1).

ericfennis: @MoltenCoffee Thanks ! Tomorrow is fine, no hurry 😄

MoltenCoffee: Done!

ericfennis (approved):
Awesome work!

PR #494, opened 2022-02-01
Add Baseline icon

ericfennis:
image


Andreto:
@ericfennis Shouldn't this be named _underline_ to stick with the text-editor namings. If not then maybe add it as a tag?

ericfennis:
@Andreto I can add is as an tag. But we already have the [underline](https://lucide.dev/icon/underline) icon.
This icon mostly used for text color and "baseline" in graphic design applications.

PR #510, opened 2022-02-18
Add cookie icon

### LINKED ISSUE #502: Cookie icon
I think a cookie icon might be helpful when people are making cookie popups and cookie settings.

## Icon Request

* Icon name: cookie
* Use case: in cookie settings, popups, etc
* _Screenshots_ of similar icons: cookie icon - I made this earlier but it probably needs changing a bit (I can give code if you want)

it-is-not:
Closes #502

Cookie iconCookie icon

(hopefully I haven't broken anything)


ericfennis:
@it-is-not Awesome!
No you didn't break anything!

One thing I noticed, just pasted the icon in Figma and noticed that the chocolate chips looks a bit different. Probably because the circle is 1.5 pixel in width. Maybe we could use the circles from [`more-horizontal`](more-horizontal) or [`cloud-now`](https://lucide.dev/icon/cloud-snow).
Let me know what you think.

image

it-is-not:
Thank you :)

With the `more-horizontal` dots:
image

With the `cloud-snow` dots (I prefer this):
image

Also, it looks like the `palette` icon also has this issue as I used that as a template for this icon.

ericfennis:
@it-is-not I agree with you. Looks good to me!

ericfennis (approved):

PR #522, opened 2022-03-07
Add keyboard icon

it-is-not:

### DIFF
--- a/icons/keyboard.svg
+++ b/icons/keyboard.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect x="2" y="4" width="20" height="16" rx="2" ry="2" />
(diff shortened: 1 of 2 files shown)


tags.json
```diff
@@ -1313,6 +1313,7 @@
     "authentication",
     "secure"
   ],
+  "keyboard": [],
```
ericfennis, line 1316:
```suggestion
  "keyboard": [
    "layout"
    "spell",
    "settings",
    "mouse"
  ],
```

icons/keyboard.svg
```diff
@@ -0,0 +1,14 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect x="2" y="4" width="20" height="16" rx="2" ry="2" />
+  <path d="M6 8h.001m4 0h.001m4 0h.001m4 0h.001M8 12h.001m4 0h.001m4 0h.001M7 16h10" />
```
ericfennis, line 13: Can you maybe split this up in separate parts, some design programs don't like this combined paths.
If you need any help with that let me know.

ericfennis (approved):
Nice work, I like the icon
Left some comments for the code.

tags.json
it-is-not: thanks, I couldn't think ideas for tags earlier

PR #532, opened 2022-03-19
Add eraser icon

maxwellito:
- **Name of the icon** : eraser
- **Tags (alternative names for this icon)** (add them in tags.json) :  "pencil", "drawing", "undo", "delete", "clear"
- **What is the purpose of this icon?** :Eraser tool for graphical interfaces
- **100% scale preview** : _see below_
- **Have you considered alternative possibilities** for its naming or design? : not really

131222669-10fc86e9-a1da-4011-9cfd-c7cf4d8d46d0

### DIFF
--- a/icons/eraser.svg
+++ b/icons/eraser.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M20 20H7L3 16C2.5 15.5 2.5 14.5 3 14L13 4L20 11L11 20"/>
+  <path d="M6 11L13 18"/>
+</svg>
--- a/tags.json
+++ b/tags.json
@@ -907,6 +907,13 @@
     "calculate",
     "off"
   ],
+  "eraser": [
+    "pencil",
+    "drawing",
+    "undo",
+    "delete",
+    "clear"
+  ],
   "euro": [
     "currency",
     "money",


ericfennis (approved):
nice love it!

mittalyashu:
I was wondering how it would look if we split this part with 2px spacing? 🤷 

CleanShot 2022-03-27 at 21 32 46@2x

maxwellito:
Something like this?

Screenshot 2022-03-27 at 21 14 03


Or like this?

Screenshot 2022-03-27 at 21 10 49

I'm sorry, I'm not sure to understand the request.

ericfennis:
I think @mittalyashu want to mention the Guideline 8 from [Icon Guidelines](https://lucide.dev/docs/icon-design-guide#8-distinct-elements-must-have-2-pixels-of-spacing-between-each-other). I think it is ok to have them connected, but that my opinion. But if you separate them keep 2px between the elements.

maxwellito:
Oh right! Thanks for the explanation. How about the following?

Screenshot 2022-03-28 at 21 30 12

At 15 instead 16, I'm no longer respecting the rule.

Screenshot 2022-03-28 at 21 31 22

delnyn:
I think it looks better to have them connected

bornova:
Thank you for this new icon!  I could definitely use it in my application.  However, it seems to be smaller than other icons by couple pixels:
icons

maxwellito:
Yes, I noticed as well. Last night before bed I wanted to see my icon on
the lucide website and it didn't feel right. I used this icon for a
personal project and it was ok.

But next to other icons of the lucide library it feels a bit small. I will
make another PR today with a slightly bigger icon next to other Lucide
icons so we can compare.

Sorry about the troubles, I should have tested it before.

maxwellito:
Screenshot 2022-04-14 at 00 02 11


Do you think this one is enough?

Here is the new path in case : `M22 20H7L3 16C2.5 15.5 2.5 14.5 3 14L13.5 3.5L21.5 12L13.5 20`

bornova:
@maxwellito Do you still plan on implementing this larger eraser icon?  Thanks for the efforts!

maxwellito:
Sorry I forgot to follow up on it. I'm happy to implement it but I'm not sure about the protocol. Shall I open another PR?

karsa-mistmere:
@maxwellito I think we should also add the 2px rounding to the top, left and right corners as per the guidelines.

maxwellito:
Before
Screenshot 2022-06-04 at 11 42 09

After
Screenshot 2022-06-04 at 11 42 19

The changes are on this [branch](https://github.com/lucide-icons/lucide/compare/master...maxwellito:fix/eraser?expand=1&short_path=a31cc67#diff-a31cc67c661d03d5c805d7813ca3a0c009cf98bd87463bd4aac4e820e161f1b8)

ericfennis:
@maxwellito nice work, this looks better.
Feel free to open a PR.

PR #541, opened 2022-03-25
fix: shrink arrow size in sidebar icons

mittalyashu:
This PR shrinks the size of the arrow in both the sidebar icons as they were looking too big and aligned the arrows 1px to the other side.


mittalyashu:
🤦🏻 why a conflict now?

ericfennis (approved):
yeah this definitely looks better, nice improvement!

PR #544, opened 2022-03-28
feat: add 'a' plus/minus icons

mittalyashu:
Can be used for increase/decreasing font size.

| Preview |
| --- |
| CleanShot 2022-03-28 at 19 33 50@2x |

### DIFF
--- a/icons/a-minus.svg
+++ b/icons/a-minus.svg
@@ -1,0 +1,15 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M4 20l8-16 8 16" />
+  <path d="M7 15h10" />
+  <path d="M22 5h-6" />
+</svg>
--- a/icons/a-plus.svg
+++ b/icons/a-plus.svg
@@ -1,0 +1,16 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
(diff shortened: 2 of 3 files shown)


delnyn:
it looks kinda weird to me...  
also in which use case would this be useful?

mittalyashu:
> In which use case would this be useful?

If we want to add a button for increase or decreasing the font size.

delnyn:
Oh okay  
your opinion on this @ericfennis ?

mittalyashu:
I was thinking to make the "A" alphabet more smaller but we also haven’t setup any guidelines on english alphabets either.

ericfennis:
I think the plus and the minus look to close to the A. Maybe move the A a little bit to the left, creates more room for the minus and the plus

mittalyashu:
> I think the plus and the minus look to close to the A. Maybe move the A a little bit to the left, creates more room for the minus and the plus

@ericfennis I also agree, but then, the "A" alphabet would not be vertically centred. The only I can think of is to shrink the size bit more.

ericfennis:
I think the icon should be vertically centered but horizontally it is not really that problem. See for example the `user-plus` icon.

delnyn:
maybe you should narrow the A

SnoutBug:
I personally think that an arrow would be more fitting because you increase the font size rather than adding or subtracting a font size.

In my head plus or minus is for deleting or adding

Take a look at open office writer or microsoft word

ericfennis:
@mittalyashu Any updates on this?

mittalyashu:
Adding space between `A` and `+`/`-`.

| |
| --- |
| CleanShot 2022-05-15 at 12 59 13@2x  |

This is, adding space + shrink `A` alphabet.

| |
| --- |
| CleanShot 2022-05-15 at 12 59 43@2x |

ericfennis:
@mittalyashu Nice work, I think it is better to have the A on the same position in both icons.
I think I like the first one the most.

ericfennis:
any progress on this @mittalyashu ?

ericfennis:
Closing this due to inactivity

PR #545, opened 2022-03-28
feat: dim sun icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2022-03-28 at 19 34 25@2x |

### DIFF
--- a/tags.json
+++ b/tags.json
@@ -2286,6 +2286,10 @@
     "weather",
     "light"
   ],
+  "sun-dim": [
+    "dim sun",
+    "dim"
+  ],
   "sunrise": [
     "weather",
     "time",
--- a/tags.json
+++ b/tags.json
@@ -2287,7 +2287,7 @@
     "light"
   ],
   "sun-dim": [
-    "dim sun",
+    "brightness",
     "dim"
   ],
   "sunrise": [
--- a/tags.json
+++ b/tags.json
@@ -2288,7 +2288,9 @@
   ],
   "sun-dim": [
     "brightness",
-    "dim"
+    "dim",
+    "low",
+    "brightness low"
   ],
   "sunrise": [
     "weather",
--- a/icons/sun-dim.svg
+++ b/icons/sun-dim.svg
@@ -10,8 +10,12 @@
   stroke-linejoin="round"
 >
   <path d="M12 16a4 4 0 100-8 4 4 0 000 8z" />
-  <path d="M12 2v2" />
-  <path d="M12 20v2" />
-  <path d="M2 12h2" />
-  <path d="M20 12h2" />
+  <path d="M12 4h.01" />
+  <path d="M20 12h.01" />
+  <path d="M12 20h.01" />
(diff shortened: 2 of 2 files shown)


mittalyashu:
I also kind-of don't look like sun either 😅

fdev:
What is the use case for this icon? The most common reason for the sun to be "dim" is when there are clouds in front of it, and there is [already an icon](https://lucide.dev/icon/cloud-sun) for that.

delnyn:
yeah well I get the use case but would it be really useful?

mittalyashu:
Every icon can have it's uses, someway or other 🤷

ericfennis (approved):

tags.json
```diff
@@ -2286,6 +2286,10 @@
     "weather",
     "light"
   ],
+  "sun-dim": [
+    "dim sun",
+    "dim"
```
bduffany, line 2291: mind adding "brightness" to the tags here? maybe also add "brightness" to the regular "sun" icon as well

tags.json
```diff
@@ -2286,6 +2286,10 @@
     "weather",
     "light"
   ],
+  "sun-dim": [
+    "dim sun",
```
bduffany, line 2290:
```suggestion
    "brightness",
```

(remove "dim sun" tag as it's not really useful, and add "brightness")

mittalyashu: Thanks 🙂

karsa-mistmere:
How about something like this?
image

mittalyashu:
> How about something like this? image

OMG! That's a good one.

//cc @ericfennis

ericfennis:
@karsa-mistmere @mittalyashu I think this looks even better!

karsa-mistmere:
I think there's maybe even space for multiple dim levels (I've just checked and iOS uses the same concept for its brightness level indicator icon).
image

I don't have access to the PR code, so here's a Figma file: https://www.figma.com/file/iRBudsI3NG79KtEODp1V5Z/lucide%2Fbrightness-dim-(%23545)?node-id=0%3A1

delnyn:
`brightness-full` breaks guidelines, but everything else is great :0

karsa-mistmere:
> `brightness-full` breaks guidelines, but everything else is great :0

Yeah, at that point due to some other icons I'd been under the impression that this particular guideline could be broken on a case by case basis. Let's scratch full and stick with medium and low (and potentially night light).

delnyn:
yeah all the icons that break this rule are basically from the feather icon set and need to be refactored

karsa-mistmere:
> yeah all the icons that break this rule are basically from the feather icon set and need to be refactored

Yeah, I've been on it this last week. 😁

ericfennis (changes requested):
@mittalyashu I think the icons from @karsa-mistmere fit better. Do you want to change them?

mittalyashu:
@ericfennis just added now 🙂

ericfennis (approved):

PR #549, opened 2022-04-01
fix: Spacing in License page

fdev:
The license page has spaces missing where lines from the `LICENSE` file are joined.
I also changed the code a bit so the pipe symbol can be safely used in the license text :slightly_smiling_face: 

missing spaces


ericfennis:
Looks good! tnx

PR #550, opened 2022-04-01
fix: Remove console.log calls from site

fdev:
I noticed a bunch of `console.log` calls on the website, which I assume were used at one point to test the website code but should probably not still be there.


ericfennis (approved):
Tnx!

PR #555, opened 2022-04-11
flag-off / unflag

karsa-mistmere:
## Icon Request

* Icon name: flag-off
* Use case: Removing the flagged state from a previously flagged item
image

Proposed icon:
image


karsa-mistmere:
Done.

ericfennis:
@karsa-mistmere Awesome work, you rocking it!

PR #556, opened 2022-04-11
Hourglass icon

karsa-mistmere:
## Icon Request

* Icon name: hourglass
* Use case: Setting a timer, marking an item as timed.
image

image


ericfennis (approved):
@karsa-mistmere Awesome work!

PR #557, opened 2022-04-11
Fix misaligned lines in `crop`

karsa-mistmere:
The `crop` icon has a few misaligned lines which I believe are probably not intentional. This pull request fixes these (assumed) errors, see (old vs. new):
image


karsa-mistmere:
Sure, it's done.

ericfennis:
Awesome work!

PR #558, opened 2022-04-11
Added factory icon

karsa-mistmere:
## Icon Request

* Icon name: factory
* Use case: Used for representing industry, sectors, manufacture, businesses etc.
image

image

### DIFF
--- a/icons/factory.svg
+++ b/icons/factory.svg
@@ -9,5 +9,5 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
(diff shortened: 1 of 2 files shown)


karsa-mistmere:
https://github.com/lucide-icons/lucide/issues/119

ericfennis:
@karsa-mistmere I think the icon can be a bit more refined. Side by side with the other building icons, the "factory" looks a bit off.
image

karsa-mistmere:
@ericfennis I think you're right, I've increased the size to match the others, this is what they currently look like:
factory

ericfennis:
@karsa-mistmere Nice this is better.
One idea, I don't know if you agree with me.
My feeling the icon feels a bit empty, maybe we can try add some windows.

My take:
image

I'm curious what you think.

karsa-mistmere:
Oooh, love it!

ericfennis (approved):

PR #564, opened 2022-04-11
Added alternate line/bar chart icons

karsa-mistmere:
## Icon Request

* Icon name: bar-chart-*, line-chart
* Use case: Alternate style for line and bar chart icons.
image


ericfennis:
@karsa-mistmere Nice I thinks these looks great

PR #565, opened 2022-04-11
Fixes `award`

karsa-mistmere:
old vs. new:
image


ericfennis:
@karsa-mistmere Nice work!

PR #568, opened 2022-04-12
Fixes `battery-charging` & `cloud-lightning` + `battery-*`

karsa-mistmere:
This PR fixes guideline violations in the icon `battery-charging`:
image


ericfennis:
@karsa-mistmere Can you try to shrink the pink areas? Just to compare, not sure if it will look better.
image

karsa-mistmere:
How about this:

image
battery-charging-100

This makes the spark the exact same as in `cloud-lightning`

ericfennis (approved):
@karsa-mistmere Awesome work, well done!

karsa-mistmere:
Thank, but I'm not so sure anymore, maybe it's cloud-lightning that needs to be fixed after all; on a lo-dpi screen the bottom part of the lightning looks a bit wonky to me:
lightnings
(top row original PR, bottom row current proposal)

ericfennis:
@karsa-mistmere Looking at it closely, I think you're right. The width of the zap looks a bit too small.

ericfennis:
@karsa-mistmere I'm a bit lost with the current status, is the current zap updated with the zap from cloud-lighting. Do we need to update some things?

karsa-mistmere:
@ericfennis Um... I actually think it is `cloud-lightning` that should also be updated / tidied up a bit along with `battery-charging`, at least that was what I was trying to say in my last comment.
This is a bit of an exaggeration, but on 100% it looks to me like something like this is going on with the current icon:
image
image

ericfennis:
@karsa-mistmere Allright gotcha.

ericfennis (approved):

PR #570, opened 2022-04-12
Fixes `coins`

karsa-mistmere:
Fixes design guideline violations of `coins`:
image

### DIFF
--- a/icons/coins.svg
+++ b/icons/coins.svg
@@ -9,8 +9,8 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <circle cx="8" cy="8" r="7" />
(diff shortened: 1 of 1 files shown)


karsa-mistmere:
There's a minor guideline issue with the flags of the number 1s.
Two obvious solutions would be to:
1) have the flag at an angle
2) shrink them to fit
image

I'm personally not really a fan of either solution, but prefer the 1st out of the two.
We might also let this one slide to keep the general aesthetic of the original. 🙃

ericfennis:
I think I'm with you the first ones looks the best. I don't really mind if a micro bit will be in the "2px" spacing area, on 100% you won't see it.

ericfennis (approved):

PR #575, opened 2022-04-12
Fixes `inspect`

karsa-mistmere:
image


ericfennis:
The only change I see is the gap, that's a bit bigger. Is this change needed?
What do you think @mittalyashu ?

karsa-mistmere:
It's probably not critical, it's a very minor violation:
image

mittalyashu:
Yup, LGTM 🙂

karsa-mistmere:
@mittalyashu: But doesn't having a 2px spacing at 2x already ensure that? I mean if you have 2px at 2px, then you would have `2px - 2 * 0.5px = 1px` at 3 px (as long as corners have round, not miter joins, which of course they do).

lscheibel:
While we're at it: Do you guys think we could remove the "stem" of the pointer? In my opinion it only adds visual noise.

karsa-mistmere:
I think I agree with @lscheibel, the stem is maybe a bit too much, what about a complete redesign like this?
(This would not only fix the spacing issue and make the pointer pixel perfect, but also make it consistent in size and spacing with `square`).

image
inspect

karsa-mistmere:
@ericfennis Thanks, I've updated the PR.

ericfennis (approved):

mittalyashu:
LGTM, thank you guys.

mittalyashu:
Do you think we need to match the cursor icon [with this](https://lucide.dev/icon/mouse-pointer-2?search=cursor)?

karsa-mistmere:
I think aiming for pixel perfection and cleanliness is more important in this instance, generally speaking, the shape is not very different and is pretty much only adjusted for changes in density because of the size difference.

karsa-mistmere:
@mittalyashu For demonstration, here's the difference between `inspect`'s (green) and `mouse-pointer-2`'s (magenta) cursor:
image

PR #577, opened 2022-04-12
Fixes `mic` & `mic-off`

karsa-mistmere:
image


ericfennis (approved):
Nice change, I think it also looks better in Icon buttons!

Issue #580, opened 2022-04-13
flag-off.svg error


cyberalien:
Icon `flag-off.svg` is invalid.

Current content:
```xml
<svg
  xmlns="http://www.w3.org/2000/svg"
  width="24"
  height="24"
  viewBox="0 0 24 24"
  fill="none"
  stroke="currentColor"
  stroke-width="2"
  stroke-linecap="round"
  stroke-linejoin="round"
>
  <path d="M8 2c3 0 5 2 8 2s4-1 4-1v11"/>
  <path d="M4 22v-7m0 0s1-1 4-1 5 2 8 2M4 15V4"/>
  <path d="m2 2 20 20"
</svg>
```
Last `<path>` tag is not closed, resulting in invalid icon.

PR #581, opened 2022-04-13
Fixes flag-off

cyberalien:
Fixes https://github.com/lucide-icons/lucide/issues/580


icons/flag-off.svg
```diff
@@ -11,5 +11,5 @@
 >
   <path d="M8 2c3 0 5 2 8 2s4-1 4-1v11"/>
   <path d="M4 22v-7m0 0s1-1 4-1 5 2 8 2M4 15V4"/>
-  <path d="m2 2 20 20"
+  <path d="m2 2 20 20"/>
```
karsa-mistmere, line 14: This is my bad, also, we should probably use a `<line>` element as other `*-off` icons do instead of a `<path>`.
```suggestion
  <line x1="2" y1="2" x2="22" y2="22" />
```

cyberalien: Changed to line

ericfennis:
Awesome Tnx @cyberalien

PR #587, opened 2022-04-13
Fixes `twitter`

karsa-mistmere:
Although deprecated, I think this icon still needs to be fixed to comply with our guidelines and fix erroneous paths:
image


ericfennis (approved):
awesome!

PR #592, opened 2022-04-13
Fixes `user-check`, `user-minus`, `user-plus`, `user-x`, `users`

karsa-mistmere:
image

### DIFF
--- a/icons/user-check.svg
+++ b/icons/user-check.svg
@@ -9,7 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M16 21v-2a4 4 0 0 0-4-4H5a4 4 0 0 0-4 4v2" />
-  <circle cx="8.5" cy="7" r="4" />
-  <polyline points="17 11 19 13 23 9" />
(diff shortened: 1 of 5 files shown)


ericfennis:
nice clever solution to fix these. I'm curious what the other think.

bduffany:
what do you think about making the "user" icon have the same shoulder width as the other icons? just so the person part of the icon is the same across all icons

karsa-mistmere:
image

karsa-mistmere:
This is what that would look like:
image
image

I personally don't have strong feelings about either. Uniformity would be nice, but they're not the same width in the current iteration either.

ericfennis:
@karsa-mistmere I agree, on 100% scale the difference is hard to see. I think consistency is better, so same width as the other 'user' icons.

karsa-mistmere:
I've pushed the change, user has now the same base width as the other icons.

karsa-mistmere:
@ericfennis: is there anything else pending with this issue?

ericfennis:
Sorry for late reply, I kept it open for a time to get some more feedback from more people. But there is not a lot of response, so I think we can merge it.

ericfennis (approved):

PR #594, opened 2022-04-16
Fixes `image`, `image-minus`, `image-plus` & `image-off`

karsa-mistmere:
image

In order to get rid of all design guideline violations, the sun needed to be scaled up and the mountain's left slope needed to be shifted by one unit. In order to completely fulfill the guidelines and still keep the right slope of the mountain visible on `image-off` the mountain also had to have a 2px rounding (see below - red: original, green: fixed):
image

100% view:
original:
image
fixed:
image


karsa-mistmere:
100% view. Damn do we have these guidelines for a good reason.
image
image

karsa-mistmere:
image
Made `image-off` appear more like before, but in order to do so, I had to introduce a minuscule design violation (a tiny bit less padding than 2px between the sun + mountain and the plus sign and the mountain):
image

karsa-mistmere:
100% view comparison:
original:
image

fixed:
image

karsa-mistmere:
Come to think of it, this violation is also present in the previous version of the icon:
image

One solution would be to add the standard 2px rounding to the mountain shape, although I'm totally not sure if that's the best course of action:
image

ericfennis (approved):
Nice small improvements, but definitely looks better!

mittalyashu:
Looks great @karsa-mistmere 👍 

But the half-stroke in `image-off` icon looks bit off to me.

CleanShot 2022-04-24 at 17 07 00@2x

ericfennis:
Yeah this looks even better!

PR #625, opened 2022-04-22
Fixes `map`

karsa-mistmere:
image


ericfennis (approved):
Nice, first i looked a bit weird zoomed in, but the 100% version looks good!

PR #628, opened 2022-04-24
feat: add `night-light` & `sun-medium` icon + tags

mittalyashu:
| Preview |
| --- |
|  CleanShot 2022-04-24 at 16 59 18@2x | 

Credits goes to @karsa-mistmere 

https://github.com/lucide-icons/lucide/pull/545#issuecomment-1095338659


ericfennis:
Hmm I'm not sure about the `night-light` icon. I think it is a bit too dense in the circle itself. And what is exactly the use case. Do we have some similar icons or examples?

karsa-mistmere:
Some examples or (very) similar icons: A0AF7B15-8DD0-4285-8BDD-F6468A8A5C43
BD3599DB-A713-4884-84A2-CD3D6F0FC8BB

ericfennis:
@karsa-mistmere AllRighty thanks I get it!

ericfennis:
Do you think we need to change `night-light` to `sun-night-light` to group it with the sun icons?

karsa-mistmere:
Well, that's a tough nut. 🤔
I personally would've called the other icons `brightness-dim` and `brightness-medium` to begin with, but if the rest are grouped under sun, I guess it makes sense to also add the `sun-` prefix to `night-light` as it's not only a similar icon but a similar use case as well. It could also be `sun-moon` I guess.

ericfennis:
@mittalyashu `sun-moon` is great fit I think. Keeps them a bit grouped.

mittalyashu:
Updated 🙂

ericfennis (approved):

PR #629, opened 2022-04-24
Add movie icons

it-is-not:
movie icon
movie icon


it-is-not:
I also added movie-play and movie-pause
movie pause
movie play

karsa-mistmere (changes requested):

karsa-mistmere:
I can see legit use cases for the movie icon, not quite sure if we need the play/pause variants, can you provide a use case for these, @it-is-not?

karsa-mistmere:
Also, I think the gap between the parts is a bit narrow, and the top part is maybe a bit too dense, how about just two lines instead of three?
image
movie

it-is-not:
thank you @karsa-mistmere I'll try and sort those issues out later

karsa-mistmere:
We should probably try sticking to the grid as much as possible, straight horizontal/vertical lines on half pixels sometimes end up looking really funky on low dpi screens.

it-is-not:
agreed, I'll have another look

it-is-not:
here's the new icon with on top of a grid

image

it-is-not:
I'm not entirely sure what you mean, but I've redesigned the top so it's a rectangle, moved some of the lines to make it more evenly spread out and increased the top's rotation so it's further away from the bottom bit. Does this fix the issue?

image

ericfennis:
@it-is-not Nice Icon, love it!

ericfennis:
I think the name of the icon need to be changed to 'clapperboard', because that is what the icons represents.

mittalyashu:
According to what @it-is-not said https://github.com/lucide-icons/lucide/pull/629#issuecomment-1107889192

Those two icons might not match the lucide icon guidelines

it-is-not:
@ericfennis - I'll change this and add `movie` as a tag

it-is-not:
@mittalyashu - I've removed the play and pause icons since they don't really have a use case

ericfennis:
@it-is-not can you also remove the icons in the tags.json?

ericfennis (approved):

it-is-not:
thank you for your help @karsa-mistmere and @ericfennis and @mittalyashu 😄

Issue #330, opened 2021-05-17
Train icon ideas


MoltenCoffee:
## Icon Ideas

* Icon names: train, steam train

I tried my hand at creating a train icon, basing the design off of other transport icons like car, truck and bus. I was hoping for some feedback before I submit a PR:

Open cabin:
![train-01](https://user-images.githubusercontent.com/13321277/118499868-9b2da700-b727-11eb-8d3d-a3d1c9113c34.png)
Closed cabin, small front wheel:
![train-04](https://user-images.githubusercontent.com/13321277/118499874-9bc63d80-b727-11eb-8161-a97df3ee264d.png)
Closed cabin, same size wheels:
![train-05](https://user-images.githubusercontent.com/13321277/118499878-9bc63d80-b727-11eb-907b-2f2067b0dc81.png)

I get the feeling one might mistake it for a front loader or something. 
I think a more modern looking train would fit as well, but didn't really manage to come up with something that's unmistakenly a train whilst keeping the windowless style of the existing icons:

![train-02](https://user-images.githubusercontent.com/13321277/118499872-9b2da700-b727-11eb-95ac-24bd8bf98c4d.png)

Another approach, from a different angle is also possible, but is different from the existing transport icons:

![train-06](https://user-images.githubusercontent.com/13321277/118502792-522b2200-b72a-11eb-8438-eba2a8e361b2.png)
![train-07](https://user-images.githubusercontent.com/13321277/118502794-52c3b880-b72a-11eb-842b-5ab9634c6fc8.png)
![train-09](https://user-images.githubusercontent.com/13321277/118502796-52c3b880-b72a-11eb-8e24-d5deddf8027a.png)

Would love to hear your thoughts!

ericfennis:
@MoltenCoffee Nice work.

I think both types of icons for transportation icons are welcome.
I like the first one of the locomotives.
For the train I think a front view is very recognizable to describe transportation.
I like the way how material does it, the headlights and the track  make it really like a train. But the icon lines are very dense, maybe we can use some inspiration of this and make it more cleaner? 🤔 
![image](https://user-images.githubusercontent.com/11825403/118719474-d368de80-b828-11eb-97b2-318fe1e582ff.png)

MoltenCoffee:
I agree! The main problem I had was the rule about having 2 points distance between separate elements, like the train and the track. Is that more a guideline then a rule?

delnyn:
[...]

Issue #609, opened 2022-04-22
[Icon Requests from Feather] `walking`


karsa-mistmere:
As per https://github.com/lucide-icons/lucide/issues/119 and https://github.com/feathericons/feather/issues/471:

![image](https://user-images.githubusercontent.com/17746067/164645033-1075fa5d-8d89-4cef-8d0f-83940b130d85.png)
![walking](https://user-images.githubusercontent.com/17746067/164645044-ecc1adbb-a2f3-414c-a72a-32a8dea03b07.png)

ericfennis:
I think this one should be matching [`person-standing`](https://lucide.dev/icon/person-standing).

karsa-mistmere:
This redesign would make it have the same dimensions, but at the same time, it makes the walking person look comically short compared to the standing person (partly because of the slight guideline violations in case of the shoulders of `person-standing`):

![image](https://user-images.githubusercontent.com/17746067/164983131-aacb0738-7b81-44e8-8c05-f34903e22ad4.png)
![image](https://user-images.githubusercontent.com/17746067/164983137-25dcac29-f81f-491e-9c6f-56c5cebb5268.png)


Moving this icons shoulders 1px closer to its head kind of solves this issue, but also violates the design guidelines:
![image](https://user-images.githubusercontent.com/17746067/164983186-28d018d4-b631-40fd-b6f0-91d9b491ba8e.png)

Not sure, what the best approach would be. Maybe `person-standing` also needs a redesign.

karsa-mistmere:
@ericfennis Come to think of it, `walking` should definitely also match [`bike`](https://lucide.dev/icon/bike), since they're likely to be used together.
![image](https://user-images.githubusercontent.com/17746067/166923920-46895428-e61e-4657-b46a-77defcc5a6e4.png)

Problem is, `bike` itself - neat as it is - breaks a handful of our current guidelines. 😅
I think the reality is, the body parts of these ISO DOT style pictograms should probably be exempt from the 2px spacing rule, since they look much better with about a 1-1.5px spacing even on low dpi.

Here's what I think would be a harmonization of different transport icons (blue: existing and should be kept, red: existing and should be fixed, green: new icons):
![image](https://user-images.githubusercontent.com/17746067/166937141-80221e81-6843-4f83-8175-786c2c1a29d4.png)
![transport](https://user-images.githubusercontent.com/17746067/166935021-eeb2806d-4793-41d5-bd83-f66898fa8f16.png)

[...]

Issue #112, opened 2020-10-26
Write conventions for naming icons


locness3:

ericfennis:
Good one.

We now use more of a mix. I think I like the idea of naming the icons what is is presenting, we already do for a lot of icons, instead of naming it to the function or purpose of the icon.

delnyn:
that means we'll have to rename a lot of icons (e.g. `settings`, `edit`)

ericfennis:
@delnyn sort of true. This solution can have a big impact

delnyn:
or we can just write a naming convention without changing the name of the existing icons

ericfennis:
Done in #1518.
See: https://lucide.dev/guide/design/icon-design-guide#naming-conventions

PR #639, opened 2022-04-29
Add train icon

delnyn:
as discussed in #330  
  
100% scale
image

### DIFF
--- a/icons/train.svg
+++ b/icons/train.svg
@@ -1,0 +1,15 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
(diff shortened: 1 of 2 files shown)


ericfennis (approved):

karsa-mistmere:
To keep the naming of icons consistent, I think it would be a good idea to call this `train-2`/`transport-train`/`train-transport` and maybe also add a train icon matching the existing transport icons (I think the third steam train from #330 looks pretty good), since the addition of further transport icons similar to this one (e.g. car, bus, see #609) seems inevitable, and it would be great if `train-2` didn't look like `car` and `train` like `car-2`. :)

delnyn:
oh well  
yes  
what should I name it exactly

karsa-mistmere:
Not quite sure. We still don't have solid naming conventions: #112 
What do you think, @ericfennis?

ericfennis:
Hmm I think `train` is fine. I understand some people see the locomotive train version as a more obvious choice for this icon.    But I think we should call that one `locomotive`

PR #642, opened 2022-05-01
Add `waterfall` icon

lscheibel:
I needed this one for a soundboard.

image

### DIFF
--- a/icons/waterfall.svg
+++ b/icons/waterfall.svg
@@ -1,0 +1,15 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M2 6h4m16 0h-4" />
+  <path d="M6 18V7a2 2 0 0 1 2-2h8a2 2 0 0 1 2 2v11" />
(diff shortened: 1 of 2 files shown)


karsa-mistmere:
Not sure about this one. What would be the use case?

Also, I'm not convinced this is easily recognisable as a waterfall, how about an alternate design or adding a few lines of movement?
image

lscheibel:
Tbh these don't have the "feather" feeling for me. Taking a look at the original designs (https://feathericons.com/) most of them try to use as much whitespace as possible and I personally like the "light" ones best (like the clipboard).

As for the use-case: It is very niche, I needed it for a soundboard but I would totally understand it if we decided to skip this one!

karsa-mistmere:
Yes, as minimalist and light as possible is definitely preferable, but I think the icons still need convey reasonably well what they mean to represent and unfortunately, I don't believe the original proposal succeeds in this regard. 😕

Still not quite sure about the use case. On the soundboard, what does the waterfall icon stand for?

lscheibel:
It was the sound of water splashing 😄 Let's hear what others think, but I'd be fine with closing this one.

karsa-mistmere:
(Oh, _that_ kind of soundboard 😅)

karsa-mistmere:
@lscheibel: I think for this particular use case, [`waves`](https://lucide.dev/icon/waves) or [`droplets`](https://lucide.dev/icon/droplets) could work as well?

karsa-mistmere:
@lscheibel: I've been revisiting this topic time and again in the last two weeks and made a few alternate drafts and ideas (fountain?).
image

ericfennis:
This icon design needs more prototyping and research. For more discussion about this icon design, you can continue in this thread. But I close this for now because it is inactive.

PR #643, opened 2022-05-01
Add `undo` and `redo` alternative

lscheibel:
A client asked for the more traditional variants of the undo icon. I thought I'd submit it as a PR but feel free to just close it if you feel that the current one is enough.

image

image


karsa-mistmere:
Nice one! I like these much better than the current ones. Maybe the lower part could be just a bit shorter though?
undoredo
image

ericfennis:
@lscheibel Love the alternatives!

lscheibel:
@karsa-mistmere Yes agreed, I updated the PR

ericfennis (approved):
Awesome work!

Issue #599, opened 2022-04-22
[Icon Requests from Feather] `forest`


karsa-mistmere:
As per https://github.com/lucide-icons/lucide/issues/119 and https://github.com/feathericons/feather/issues/56:

![image](https://user-images.githubusercontent.com/17746067/164629355-a1f0709f-4f7e-46b9-b29b-3314c14675d8.png)
![forest](https://user-images.githubusercontent.com/17746067/164629422-1c940178-f1f0-4b21-931c-c1f6bd668825.png)

lscheibel:
What do you think of something with less details?

<img width="599" alt="image" src="https://user-images.githubusercontent.com/44374653/166160117-56c2a3f6-160b-41dd-9df2-314a43bf57df.png">

karsa-mistmere:
@lscheibel Nice one, can you provide a 100% view for comparison?

lscheibel:
Sure, here you go:
![image](https://user-images.githubusercontent.com/44374653/166304793-7cfc7f5e-6348-4d8c-88a0-9f3a6e5fc30b.png)

karsa-mistmere:
Looks good to me. I don't have a strong preference, although a matching single tree would be nice to have as an alternate choice.
What do you think, @ericfennis ?

ericfennis:
Hmm I don't have a strong preference either. There is not really a "iconic" forest icon. 
Maybe we can make some more icons?

PR #644, opened 2022-05-01
Add `tree` icon

lscheibel:
I needed this one for a soundboard.

image

### DIFF
--- a/icons/tree.svg
+++ b/icons/tree.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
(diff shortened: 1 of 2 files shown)


karsa-mistmere:
All in all nice, but I think the trunk needs to be taller, as deciduous trees' trunks usually are, otherwise it looks a little more like a shrub than it does a tree:
image
image

lscheibel:
I like the last one, I thought it looked a bit like cotton candy but I think in context it works. I updated the PR

ericfennis:
@lscheibel @karsa-mistmere Nice, I've one suggestion I'm curious what you think of it.
Maybe rename this to `tree-broadleaf` and create add another icon for `tree-pine`. I can image sometimes the pine tree is for some people more obvious choice.

karsa-mistmere:
@ericfennis Yeah, totally agree, there should be different kinds of trees. There's actually already an issue in progress: #599
I would maybe name this `tree-deciduous` (but it definitely needs it at least as a tag).

Here's a mashup overview of current design proposals by @lscheibel and myself.
image
tree

ericfennis:
@karsa-mistmere I like the top row. But with the current `tree-deciduous` icon.

karsa-mistmere:
@ericfennis: I guess that can work, but I actually quite grew to like the top row version, it's quite unmistakable and I think @lscheibel is right that the second row version does in fact look a bit like cotton candy.

PR #645, opened 2022-05-01
Add `save-as` icon and `save` alternative

lscheibel:
I had these a bit more modern looking alternatives for save icons lying around for a while now.

This would also address https://github.com/feathericons/feather/issues/609

image

### DIFF
--- a/icons/save-2.svg
+++ b/icons/save-2.svg
@@ -1,0 +1,15 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M6 17v1a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V6a2 2 0 0 0-2-2H8a2 2 0 0 0-2 2v1" />
+  <path d="m10 15 3-3-3-3" />
+  <path d="M2 12h11" />
+</svg>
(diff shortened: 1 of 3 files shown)


karsa-mistmere:
Not bad (certainly solves the save as fiasco), but maybe the arrow could be a bit more pronounced (this would also make the icon more visually centered):
image
save

karsa-mistmere:
Sidenote: if we add this alternate save icon, I think a matching load icon should be trivial. :)

ericfennis:
@lscheibel Nice love the alternative ones.

lscheibel:
I agree, that the arrow is better readable when increased in size, however just increasing it's size makes the icon feel packed/dense. Instead we could also increase the bounds size like so:

image

image

This goes against the 1px margin guideline, however I like this much better than truncating the arrows tail.

karsa-mistmere:
Yeah, that particular guideline is pretty crucial, so that 3px strokes do not overflow or get clipped.

I've been doodling around, here are a few alternate versions:
image
save icons

I'm also been wondering whether it is preferable to keep the arrows consistent between the two icons (top row) or keep the centre of gravity about the same (bottom row):
image

What do you think, @ericfennis?

ericfennis:
Sorry for late reply, was on vacation.
Hmm I'm not sure yet of the whole concept of the icon. I'm not sure if this icon is recognisable enough as a "save-as" icon.
Do we have some examples of icons have this as a save-as icon?

karsa-mistmere:
Yeah, I also have my doubts. From what I could find this design originates from a submission to a competition to redesign the save as icon:
- https://dribbble.com/shots/1747652-the-new-save-icon?utm_content=buffer71f67&utm_medium=social&utm_source=twitter.com&utm_campaign=buffer
- https://web.archive.org/web/20150601021752/http://speichern-unter.net/beitrag.html?project=1244

Not sure if anyone ever actually used it, at least I couldn't find any icon sets actually containing an icon like this for either save or save as.

ericfennis:
This one needs more exploration, to something recognisable. I close this one for now.

PR #647, opened 2022-05-02
Added `map-pin-off`, `navigation-off` & `navigation-off-2`

karsa-mistmere:
As per #119 and https://github.com/feathericons/feather/issues/841
image


ericfennis:
nice icons!
I think the name of `navigation-off-2` should be changed to`navigation-2-off`

ericfennis (approved):

Issue #650, opened 2022-05-03
Guidelines vs Rules


lscheibel:
This is a problem that has been here since the beginning and after seeing more icons being "fixed" (and doing so myself) I feel the need to get this out of my head. Please keep in mind, that this is only my opinion and I understand that not everybody will feel the same but maybe this can at least spark some discussion:

In my opinion, our guidelines should not be declared as rules.

I feel the guidelines have been interpreted too much as rules, instead of guiding principles. I understand that trying to define the "feather" look in text is difficult, but this is even more reason to not put too much weight on our design guide. This is also reflected in [Coles answer when asked to declare guidelines](https://github.com/feathericons/feather/issues/171) which used the words constraints and guidelines and put a clear distinction between them.

Let's take for example the `shopping-cart` and the `shopping-bag` icons:

![image](https://user-images.githubusercontent.com/44374653/166443234-854e47ef-40ca-43ba-8217-ae801d9d231f.png)

To me one icon looks smaller than the other. Now let's take a look at the feather icons:

![image](https://user-images.githubusercontent.com/44374653/166443517-b9a5ec39-423d-42b1-a943-5e4e05d5409a.png)

To me, these feel much more balanced, even though the sides of the cart poke out of the "safe-zone". In the world of typography this is better known as [overshoot](https://learn.scannerlicker.net/2014/09/03/the-art-of-eyeballing-part-3-overshooting/) and over at feathericons.com similar decisions have been made along a lot of icons in order to keep a consistent *visual* size between icons. Now we can't be as exact with it as type designers, after all we are constraint to a 24x24 (mostly) pixel-perfect grid, however I believe that this was the main motivation for what Cole Bemis called "similar optical volumes". In fact, this principle hasn't even made it into our rules and I would understand why: It is practically impossible to measure. It cannot be a rule because it is inherently just a blurry guideline. But as undefined it may be, I believe it is the main driver to keep a consistent set of icons.

Please do not take this as personal criticism of anybody. I myself have "verschlimmbessert" some icons in a very similar way but I thought I could at least write down my thoughts on this.

karsa-mistmere:
I think you're definitely onto something, my two pennies on the matter:

I think Feather and Lucide's guidelines are equally outdated. I completely agree that some guidelines should be marked as requirements/constraints and some should be marked as recommendations, and we do definitely need to add some extra guidelines for sake of consistency.

---

As for our current guidelines, I think we can all agree that the following are trivial and must be defined as requirements:
[...]

PR #653, opened 2022-05-04
Fixes `shopping-cart`

karsa-mistmere:
This PR adds slightly more visual weight and a somewhat better centre of gravity to the `shopping-cart` icon, as per #650.
Also attached are a few alternate styles up for discussion:

image
shopping cart


ericfennis:
I think this update is fine. The alternatives are cool, but I think that should be another discussion for a second `shopping-cart` icon.

ericfennis (approved):

PR #662, opened 2022-05-10
Added `star-off`

karsa-mistmere:
## Icon Request

* Icon name: `star-off`
* Use case: removing starred status or rating, unlike/dislike
image
image


ericfennis (approved):
Nice!

PR #666, opened 2022-05-10
Adds `fingerprint`

karsa-mistmere:
## Icon Request

* Icon name: `fingerprint`
* Use case: biometric or 2 factor authentication, data fingerprinting
image

image
image


ericfennis:
hmm I'm not sure about this one. Some lines look a bit wiggly.
What do think of using the [`lucide`](https://github.com/lucide-icons/lucide/blob/master/site/public/logo.svg) logo and adjust it a bit to a fingerprint (not sure if this is possible).

karsa-mistmere:
Yeah, you're totally right, and you're not gonna believe this, but I actually already have a draft based on the Lucide logo. 😂
(I guess GMTA?)

I've also made a cleaner and a more geometric version of the original:
image

I'm now definitely more partial to the 2nd and 3rd, they fit much better with Lucide.

ericfennis:
@karsa-mistmere haha nice!
But looking at it, I think it won't work.
I think the two first are the best. I'm not sure which one. 
What do you think?

karsa-mistmere:
@ericfennis Good question, the first looks more organic and hence a bit more natural and legible, whereas the second fits much better with the very geometric style of the rest of our icons and is pixel perfect.

I also made a 100% view, but forgot to attach it, so here it is, hopefully this is also helpful:
fingerprint

ericfennis:
Ok let's go for the first one!

karsa-mistmere:
Thanks to both of you for the feedback, I've updated the PR.

ericfennis (approved):
Awesome!

PR #667, opened 2022-05-10
Fixes `pin` and adds `pin-off`

karsa-mistmere:
I think the current `pin` icon has several issues, the style is inconsistent with the rest of the icons, it has very little visual weight, not to mention the rendering issue.
Here is my proposal to fix these issues, and maybe also add a `pin-off` icon while we're at it (usually when pinning stuff, unpinning/removing the pin also comes up).
Also included are 45deg versions up for discussion (for maybe including as alternate styles?).

image
pin


lscheibel:
I think I like the top version the best 👍

ericfennis:
I agree with @lscheibel .

karsa-mistmere:
@ericfennis: perfect, that's what I included in the PR. 😊

ericfennis (approved):

PR #668, opened 2022-05-10
Improve the appearance of `thumbs-up` & `thumbs-down`

karsa-mistmere:
I think these icons could to with a bit of a makeover to improve their legibility, what do you think of these changes?
image
thumbs


ericfennis:
I like it, these looks less chonky.

PR #676, opened 2022-05-17
Fixes combined path rendering issues

karsa-mistmere:
This PR fixes a bunch of icons that render incorrectly because of unclosed combined paths that should have been broken up into their parts.
image


ericfennis:
nice work!

ericfennis (approved):

PR #678, opened 2022-05-21
feat: add `ice-cream` icon

wojtekmaj:


karsa-mistmere:
Nice icon, although I think three scoops might make it a bit too dense.
I've actually just designed these a few days ago:
0A63AB67-28EF-4585-B354-D39D51F0BDBF

wojtekmaj:
That's cool @karsa-mistmere! Think you could get that raised as a PR anytime soon?

PR #691, opened 2022-05-30
Fixes `box`, `package-*`, adds `package-open`

karsa-mistmere:
This PR unifies and fixes the inner lines of box & package icons:
image

I also adds an open variant:
image
image


ericfennis:
should it not be `package-open`? Because it has the lid or am I wrong?

karsa-mistmere:
I guess it could be, yes. :)
I've updated the PR accordingly.

ericfennis (approved):

Issue #683, opened 2022-05-24
Hotel related icons


pcout:
Can you please add hotel related icons ?
Thanks.

karsa-mistmere:
@pcout: Can you please provide a list and use cases?

pcout:
Well, I was hoping all the related features an hotel has.
I've attached some examples found in a quick google search.
Thanks.

![image](https://user-images.githubusercontent.com/687515/170289729-4b9f79ac-caca-4da4-9afc-5b7f917242df.png)
![image](https://user-images.githubusercontent.com/687515/170289877-995e1b3e-5084-495b-b6c9-4d27ffc6deb8.png)
![image](https://user-images.githubusercontent.com/687515/170289952-50197a26-8d38-485a-85d5-cb340e1eb18e.png)

ericfennis:
@pcout Could be nice to have list to work with. Also to check which icons are relevant for the icon set. We can't just add them all. Some are too specific.

pcout:
Here a list. We can think of more.
-------------------------------------------------------------------------
Hotel
Guest (man)
Guest (woman)
2 guests
2 ou more guests
Baby
Child
Reception bell
Receptionist
Clock
Mini fridge
Bathtube
Bath cabin
Lamp(s)
Room key
Key card
Single bed (front view)
Single bed (side view)
King size bed (front view)
King size bed (side view)
Door sign
AVAC
Air conditioner
Suitcase
Luggage
Calendar (wih booking)
Calendar (wihout booking)
Cocktail
Swimming pool
Some sports (horse riding, golf, tennis, swimming etc)
-------------------------------------------------------------------------

karsa-mistmere:
These ones we already have, I believe:
- [x] Hotel: not sure if adding a separate hotel icon is feasible, probably just use [`building`](https://lucide.dev/icon/building) or [`building-2`](https://lucide.dev/icon/building)
- [x] Guest (man/woman): just use [`user`](https://lucide.dev/icon/user), there's an [ongoing discussion](https://github.com/feathericons/feather/pull/46) about adding gender icons but it's a complex issue.
- [x] 2 guests, 2 or more guests [`users`](https://lucide.dev/icon/users) or [`user-plus`](https://lucide.dev/icon/user-plus)
- [x] Clock: [`clock`](https://lucide.dev/icon/clock) or [`alarm-clock`](https://lucide.dev/icon/alarm-clock)
[...]

PR #693, opened 2022-05-30
Adds `air-conditioner` icon

karsa-mistmere:
## Icon Request

* Icon name: air conditioner
* Use case: for smart home applications, using on hotel/booking pages/apps
* Originates from: #683
image
image

image
image


ericfennis:
I'm not sure about this icon, but well designed.
I think they mostly use a [`snowflake`](https://lucide.dev/icon/snowflake) for air conditioning.

Checked on the internet here are some examples.
Airbnb:
image
Booking.com: 
image

karsa-mistmere:
@ericfennis Yeah, I kind of agree, I even suggested `snowflake` in #683. :)
Not sure if this has any other use cases, maybe it makes sense to have a unique icon in smart home apps, where snowflake could also stand for other appliances e.g. fridge, freezer etc.
(Then again, `thermometer-snowflake` could still work there I think).

ericfennis:
@karsa-mistmere Tnx for the explaination. You have a point. 
Maybe we should rename this to: `air-vent-grille` or something. Because can also be used for other things with air. For example air-ventilation.

ericfennis (approved):

PR #696, opened 2022-06-01
Adds `list-start`, `list-end`, `list-music` & `list-video`

karsa-mistmere:
## Icon Request

* Icon name: playlist
* Use case:
  * used for audio or video playback / playlists / queues
  * `list-start` and `list-end` also useable when ordering lists (for the actions move to top/start & move to bottom/end)
image

image
image


ericfennis (approved):
very clean!

PR #698, opened 2022-06-01
Fixes `archive` & adds `archive-restore`

karsa-mistmere:
## Icon Request

* Icon name: unarchive
* Use case: reversing the action of archiving an item.
image image
* Also breaks up archive into its separate parts to ensure smooth rendering across platforms and software.
* Originally from: https://github.com/feathericons/feather/issues/1155

image
image


ericfennis:
nice!

ericfennis:
@karsa-mistmere conflicts in the tags file

ericfennis (approved):

PR #702, opened 2022-06-02
Adds `scroll`

karsa-mistmere:
## Icon Request

* Icon name: scroll / log
* Use case: used for logs, log entries
image


ericfennis (approved):
Really nice!

PR #711, opened 2022-06-10
Add alt table icon

abejenaru:
Sync table icon from upstream Feather project


karsa-mistmere:
It't nice to merge new stuff from Feather, but I think we should also keep the Lucide version as a separate style.

ericfennis:
I agree with @karsa-mistmere .

Issue #649, opened 2022-05-02
New lucide.dev homepage design request


ericfennis:
We need a new homepage design. And I'm curious how others think about this.

## The problem
The current one is fine but performance is getting worse. 
The lighthouse insight is ranking us lower because the current homepage DOM size is to big. 
<img width="1183" alt="image" src="https://user-images.githubusercontent.com/11825403/166317396-1ac22801-bc7d-4bbd-bc28-e07d9b3629ec.png">
<img width="527" alt="image" src="https://user-images.githubusercontent.com/11825403/166317715-bf2564d5-7cc7-414c-874a-ae6d4a5f2ad3.png">
That's because we render all the icons on the homepage. So my Idea is to move the icons page to his own page, keeping the homepage fast and we can also reorder the homepages more focusing on more info e.g. add some more info about Lucide and the community.

## New Icons page
I started with the icons page already, also added categories so you can also view all the icons by category. We can also expand this with more features.

<img width="1679" alt="image" src="https://user-images.githubusercontent.com/11825403/166316961-020bf2c6-6c86-404f-be36-364645f5976f.png">

# What we need, a new design!

A new homepage design, designed in Figma or other design program, would be nice to view it online with public link.

The design should contain:
- link to icons page
- Github link
- download icons button?
- all the platforms we support, (we currently have all those icons above the buttons).
- maybe comparison things

It's a homepage so it can something cool and special, so be creative. Would be nice if it matches the current style and component library we use (ChakraUI).

Ideas are welcome!, Please leave comments below so we can discuss them!

karsa-mistmere:
Yeah, I think load time is definitely an issue, and in the long run we're probably going to have to let go of the whole concept of having all icons visible on a single page.

As for categories: I really like the general idea, but I'm not quite sure if grouping by categories is the way to go (since, as I've already said in #94, categories have significant overlaps), I'd probably opt for a faceted search.

Here's a pretty rough first draft of what I imagine the front page and the icons page could look like:
https://www.figma.com/proto/MAGha7ERy1Suu2scKvRw1Q/lucide%2Fhomepage?page-id=0%3A1&node-id=2%3A196&viewport=242%2C48%2C0.58&scaling=min-zoom

ericfennis:
@karsa-mistmere Woww that's loooks really lit :fire:, nice work!!
Really like customization panel, shows you what you can do with it.

[...]

PR #715, opened 2022-06-13
Add Discernible Name to Inaccessible Links

twanmulder:
Issue:
Fixes an accessibility issue where the anchor tags in the `Header` component, that link to the documentation of all the different frameworks, don't have a discernible name that accessibility tools can recognise.

Screenshot 2022-06-13 at 16 36 59


ericfennis (approved):
@twanmulder Welcome to the Lucide project! And thanks for this fix! 

I checked it before, we have some more issues to fix on the lucide website. 
One of the other things is the performance, I also mentioned in #649.

PR #717, opened 2022-06-13
Adds icons for various design tools

karsa-mistmere:
## Icon Request

* Icon name: design tools
* Use case: on UI design interfaces

image
design
(The complete set designed is shown on the 100% view, I've chosen not to include all of them, since some are bit too random).

There's a naming conflict with the current `scale` icon, not sure how this should be resolved, that one probably should have been called `scales`.


karsa-mistmere:
> Woow these icons look really cool 🚀!
@ericfennis Thanks! What do you think about the naming conflict?

PR #724, opened 2022-06-14
Adds `square-oblique` (aka "diamond")

karsa-mistmere:
## Icon Request

* Icon name: oblique square aka "diamond"
* Use case: a very basic regular polygon shape that can be used as a building block of future icons.
image

image
image

One thing I'm not quite sure about: maybe this should be called `diamond` instead for future ease of naming (e.g. `warning-diamond`).


ericfennis:
good idea!

ericfennis:
oh nevermind the it's called gem instead of diamond. Yeah let's name this one diamond

ericfennis (approved):

PR #731, opened 2022-06-16
Add Magento 2 Hyva package to docs

GrimLink:
I love what you all doing with this Icon pack, I made a Magento 2 module for Hyva Themes that uses Lucide -> https://github.com/Siteation/magento2-hyva-icons-lucide

This PR will add the icon to homepage header, and adds a page to to the docs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "generate:supersprite": "node ./scripts/generateSuperSVG.mjs",
     "optimize": "node ./scripts/optimizeSvgs.mjs",
     "addtags": "node ./scripts/addMissingKeysToTags.mjs",
-    "generate:changelog": "node ./scripts/generateChangelog.mjs"
+    "generate:changelog": "node ./scripts/generateChangelog.mjs",
+    "postinstall": "husky install"
   },
   "devDependencies": {
     "@atomico/rollup-plugin-sizes": "^1.1.4",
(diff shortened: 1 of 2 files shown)


GrimLink:
I am not sure how to add a section to the package page, not clear how this is handled,
I suspect this done trough each package.json in the packages folder.

But is there an easy way for 3de part modules to be added to this list?

ericfennis:
Sorry for the late reply. I appreciate the work, nice package.
I'm not sure if we need to place "not official lucide" packages on the homepage to avoid confusion among users. I now laravel there as wel but I think we should move that one as wel.
I think it is better if we make a "third party packages" section on the packages page.

GrimLink:
@ericfennis is there anything I can help with?

ericfennis:
@GrimLink Sorry for the late reply and lack of help with this implementation. It's a bit on a low prio right now. 

If you like to add it you can, we want to create a third-party section on the packages page on the site. We can create a json file for it to specify the package name and etc. We already have an json file for the first party packages, so we can base it on that one.

ericfennis:
@GrimLink I've added `https://github.com/lucide-icons/lucide/blob/main/site/src/data/packageData.thirdParty.json`, a data list for third-party packages. It will show up on https://lucide.dev/packages.

GrimLink:
@ericfennis super!
I have not got a chance to to take a look to improve the 3de party implementation.
So this great.

I will create a new PR with with only the change to the `packageData.thirdParty.json` 😉

PR #747, opened 2022-07-04
fix: fix lucide-react-native failing to publish

wojtekmaj:
I believe that should fix the issues we're dealing with.

https://github.com/lucide-icons/lucide/runs/7116958776?check_suite_focus=true

Note: I was not able to get Yarn classic installed on my machine, so the changes related to lockfile are not included here. Sorry about that!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -496,6 +496,7 @@ jobs:
         pre-build,
         lucide,
         lucide-react,
+        lucide-react-native,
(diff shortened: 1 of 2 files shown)


ericfennis (approved):

wojtekmaj:
@ericfennis What about yarn.lock update?

ericfennis:
@wojtekmaj Will check it now!

PR #748, opened 2022-07-04
Migrate to Yarn 3

wojtekmaj:


ericfennis:
I'm not sure about the choice to move to Yarn 3. Not about Yarn 3 but I'm curious if maybe an other Package Manager can benefit our work more. I hear a lot about [PNPM](https://pnpm.io/) the last days, because of the high performance, maybe this is something for Lucide as well.

PR #746, opened 2022-07-03
Add more music icons and another mic icon

it-is-not:
music and mic icon

PR #751, opened 2022-07-06
Add more music icons and another mic icon

it-is-not:
This one will merge into `main` (original #746)


karsa-mistmere:
The changes to tags.json are missing.

PR #754, opened 2022-07-18
add "aspect-ratio-16x9" or "rectangle" icon

Reund0:
Hi there.
I was creating buttons for my image editor and needed a rectangle to show 16:9 cropping option.
But when I searched the Lucide icons, there wasn't any rectangle or aspect-ratio icon.
So I open a PR for this icon. 

PS: aspect ratio of the rectangle icon is 16:9.


icons/aspect-ratio-16x9.svg
```diff
@@ -0,0 +1,13 @@
...
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect x="2" y="6.44" width="20" height="11.25" rx="2"/>
```
karsa-mistmere, line 12: This looks really ugly on low dpi, while it's not a true 16:9 aspect ratio, I suggest to go with the one below, and just call it `rectangle-horizontal`.
```suggestion
  <rect x="2" y="6" width="20" height="12" rx="2"/>
```

image

Reund0: Yes, you're right. I change the pixel sizes and also add a vertical version of the icon but it seems there are problems.

tags.json
```diff
@@ -2950,6 +2950,18 @@
     "music",
     "connect"
   ],
+  "rectangle-horizontal": [
+    "rectangle",
+    "aspect ratio",
+    "16:9",
+    "horizontal",
```
ericfennis, line 2957:
```suggestion
    "horizontal"
```

ericfennis (changes requested):
Remove trailing commas in JSON file

icons/aspect-ratio-16x9.svg
karsa-mistmere: Upon further consideration, while I think such rectangle icons would be useful, especially as building blocks of future icons, I'd prefer to keep them free of any explicit aspect ratio, such as: 
image

Reund0:
Oh sorry i forgot. It's done.

icons/rectangle-vertical
```diff
@@ -0,0 +1,13 @@
+<svg
```
ericfennis, line 1: This file is missing the `.svg` extension

ericfennis (changes requested):

ericfennis (approved):

PR #783, opened 2022-08-06
Add color prop to "Additional Props" example

### LINKED ISSUE #782: Missing prop in Lucide Svelte additional props example
When looking at the documentation for Lucide Svelte, I noticed that the "additional props" example does not have any additional props.
image

j-cleme:
The additional props example previously had no props at all, this adds the color prop to solve that problem.

Resolves #782


ericfennis (approved):
Thanks for the update of the docs!

PR #785, opened 2022-08-17
Add `arrow-up-down` icon

it-is-not:
image

There is already an `arrow-left-right` icon but not an `arrow-up-down` icon (this icon has just been turned around). You could use it as a reorder icon or as a mobile data icon.

### DIFF
--- a/icons/arrow-up-down.svg
+++ b/icons/arrow-up-down.svg
@@ -1,0 +1,16 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <polyline points="11 17 7 21 3 17" />
+  <line x1="7" y1="21" x2="7" y2="9" />
+  <polyline points="21 7 17 3 13 7" />
+  <line x1="17" y1="15" x2="17" y2="3" />
+</svg>
--- a/icons/arrow-up-down.svg
+++ b/icons/arrow-up-down.svg
@@ -9,8 +9,8 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <polyline points="11 17 7 21 3 17" />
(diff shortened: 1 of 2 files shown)


tags.json
```diff
@@ -318,6 +318,17 @@
...
+  "arrow-up-down": [
+    "bidirectional",
+    "direction",
+    "swap",
+    "switch",
+    "network",
+	"mobile data",
+	"internet",
+	"reorder",
+	"move"
```
karsa-mistmere, line 330:
```suggestion
    "mobile data",
    "internet",
    "reorder",
    "move"
```
I think `reorder` & `move` could also be added to `arrow-left-right`.

icons/arrow-up-down.svg
```diff
@@ -0,0 +1,16 @@
...
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <polyline points="11 17 7 21 3 17" />
+  <line x1="7" y1="21" x2="7" y2="9" />
+  <polyline points="21 7 17 3 13 7" />
+  <line x1="17" y1="15" x2="17" y2="3" />
```
karsa-mistmere, line 15:
```suggestion
  <polyline points="11 17 7 21 3 17"/>
  <line x1="7" y1="21" x2="7" y2="9"/>
  <polyline points="21 7 17 3 13 7"/>
  <line x1="17" y1="15" x2="17" y2="3"/>
```

karsa-mistmere:
Nice icon, I wasn't quite sure about the name; to me this should be `arrow-down-up`, but then again, I also think the horizontal version should be `arrow-right-left`, so I guess yours should be fine.

I've made a few suggestions to the formatting in line with JSON and SVG code styles.

it-is-not:
Thank you @karsa-mistmere.

I think we could change the name later on if necessary, or we could add the other name as a tag.

karsa-mistmere:
Or maybe as an alias once we have support for those. :)

Can you please also add `reorder` and `move` as tags to `arrow-left-right`?

karsa-mistmere (approved):

ericfennis (approved):

PR #809, opened 2022-10-16
Add icon `check-check`

ericfennis:
Add double check icon.
image

100%
check

Contributes to #119

### DIFF
--- a/icons/check-check.svg
+++ b/icons/check-check.svg
@@ -9,6 +9,6 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M18 7 7 18l-5-5" />
-  <path d="m22 11-7.5 7.5L13 17" />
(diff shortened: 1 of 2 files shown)


icons/check-check.svg
```diff
@@ -0,0 +1,14 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M18 7 7 18l-5-5" />
+  <path d="m22 11-7.5 7.5L13 17" />
```
karsa-mistmere, line 13: These are 1px off to the bottom compared to `check`, I guess that probably wasn't intentional.
```suggestion
  <path d="M18 6 7 17l-5-5" />
  <path d="m22 10-7.5 7.5L13 16" />
```

PR #812, opened 2022-10-16
Fix `signal-zero`

azdle:
Currently the icons `signal-zero` and `signal-low` are exactly the same.

signal-low signal-zero

Previously called out here https://github.com/lucide-icons/lucide/pull/428#issuecomment-1099289022 and confirmed by @ericfennis, the author of the original PR.

I've just manually edited `signal-zero` to remove the second bar.


ericfennis (approved):
nice catch!

PR #827, opened 2022-10-19
Better tags 1

ANTARCTUS:
I upgrade the tags list with adding the tag "maths" and "shape". And with upgrading "social", "code" and "logo".

### DIFF
--- a/tags.json
+++ b/tags.json
@@ -3958,7 +3958,9 @@
     "park",
     "nature"
   ],
-  "trello": [],
+  "trello": [
+    "logo",
+  ],
   "trending-down": [
     "statistics"
   ],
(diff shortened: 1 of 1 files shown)


tags.json
```diff
@@ -3958,15 +3982,18 @@
     "park",
     "nature"
   ],
-  "trello": [],
+  "trello": [
+    "logo",
```
ericfennis, line 3986: Remove trailing coma, json don't like that
```suggestion
    "logo"
```

ericfennis (commented):
Thanks for adding these tags, one small change

ericfennis (approved):

PR #832, opened 2022-10-21
Docs: fix and extend the generic vue3 icon component example

friedow:
The generic icon component example for vue3 was missing an import for `computed` which has become an optional import with the composition api in vue3. I've fixed that and also refactored both examples to utilize the recently introduced `<script setup>` which increases the readability of components while reducing their lines of code.

The minimal usage example now looks like this:
``` html
<template>
  <Camera
    color="red"
    :size="32"
  />
</template>

<script setup>
import { Camera } from 'lucide-vue-next';
</script>
```

Furthermore I've extended the example to proxy all properties (`size`, `color`, `strokeWidth` and `defaultClass`) to the specific icon components. Things like this are now possible with the example:
``` html
<template>
  <div id="app">
    <Icon name="Airplay" size="20" stroke-width="1"  />
  </div>
</template>
```

### DIFF
--- a/docs/packages/lucide-vue-next.md
+++ b/docs/packages/lucide-vue-next.md
@@ -35,15 +35,8 @@ You can pass additional props to adjust the icon.
   />
 </template>
 
-<script>
-// Returns Vue component
+<script setup>
 import { Camera } from 'lucide-vue-next';
-
-export default {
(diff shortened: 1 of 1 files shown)


ericfennis (approved):
Tnx this is much better!

PR #835, opened 2022-10-24
Make lucid-angular compatible with angular 14

ekowbaah:


ericfennis (commented):
@ekowbaah Nice work, are you sure this is still working for older versions of angular as well?

ekowbaah:
> @ekowbaah Nice work, are you sure this is still working for older versions of angular as well?

@ericfennis I will do some manual testing with previous versions and let you know.

ekowbaah:
@ericfennis tested against previous versions >= angular11 and works fine

ericfennis:
Awesome!

ekowbaah:
Hi @ericfennis, realized the lucid angular build was failing due to the tests not running. Are you working on that? Seems its due to an outdated chromium version

ericfennis:
Yeah noticed it. Will check

ekowbaah:
Hello @ericfennis , I was wondering if you got the chance to check why the angular build was failing.

PR #838, opened 2022-10-25
Add packageManager to package.json to allow usage with Corepack

wojtekmaj:
For devs with Node + Corepack, but no pnpm, contribution will be seamless, as pnpm will be installed transparently in the background while running the first command.

https://pnpm.io/installation#using-corepack

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,5 +62,6 @@
   },
   "lint-staged": {
     "icons/*.svg": "node ./scripts/optimizeStagedSvgs.mjs"
-  }
+  },
+  "packageManager": "pnpm@7.14.0"
 }


package.json
```diff
@@ -62,5 +62,6 @@
   },
   "lint-staged": {
     "icons/*.svg": "node ./scripts/optimizeStagedSvgs.mjs"
-  }
+  },
+  "packageManager": "pnpm@7.14.0"
```
karsa-mistmere, line 66: Nice one, thanks.

karsa-mistmere (approved):

ericfennis (approved):

PR #846, opened 2022-10-25
Reduce __NEXT_DATA__ size by removing content and author.commit properties from IconEntity

wojtekmaj:
Reduces `__NEXT_DATA__` size on main page by 202694 bytes (~31%) by removing `content` property (it's now calculated based on `src` property) and removing `author.commit` data altogether.

### DIFF
--- a/site/src/components/IconWrapper.tsx
+++ b/site/src/components/IconWrapper.tsx
@@ -1,7 +1,7 @@
 import { forwardRef, SVGProps } from 'react';
 
 interface IconWrapperProps extends SVGProps<SVGSVGElement> {
-  content: string;
+  src: string;
 }
 
 export const IconWrapper = forwardRef<SVGSVGElement, IconWrapperProps>((props, ref) => {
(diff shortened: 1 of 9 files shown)


site/src/components/IconWrapper.tsx
```diff
@@ -17,11 +17,12 @@ export const IconWrapper = forwardRef<SVGSVGElement, IconWrapperProps>((props, r
     strokeLinejoin: 'round',
   };
 
-  const { content, ...rest } = props;
+  const { src, ...rest } = props;
   const attrs = {
     ...defaultAttrs,
     ...rest,
   };
+  const content = src.replace(/<svg[^>]*>|<\/svg>/g, '');
```
wojtekmaj, line 25: Core change is here. We no longer need to pass `content`, which is essentially the same thing as `src` minus `<svg>` wrap.

wojtekmaj:
I'll think how we could improve this further. :)

site/src/components/IconWrapper.tsx
ericfennis: Nice findings!

PR #851, opened 2022-10-25
Update GitHub Actions actions/checkout to v3

wojtekmaj:
Fixed dependency warning:

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout
```

### DIFF
--- a/.github/workflows/lucide-angular.yml
+++ b/.github/workflows/lucide-angular.yml
@@ -12,8 +12,8 @@ jobs:
   lucide-angular:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v3.4.1
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
           node-version: 16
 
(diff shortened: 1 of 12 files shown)


.github/workflows/lucide-angular.yml
```diff
@@ -45,4 +45,3 @@ jobs:
 
       - name: Test
         run: pnpm --filter lucide-angular test
-
```
karsa-mistmere, line 48: Again, great job. Only thing I'm not sure about is removing the new lines at the end of files, if anything I would've added a new line to release.yml. @ericfennis? 🤔

wojtekmaj: I have a setting on to trim unnecessary whitespace. Can roll that back manually if you want!

ericfennis: I think it's fine, I checked we have still a EOF.

ericfennis (approved):

PR #853, opened 2022-10-25
Split commands in documentation to make it possible to use Copy to clipboard icon

wojtekmaj:
GitHub adds Copy to clipboard button to all code snippets. By splitting npm and yarn specific snippets into two separate snippets we make it possible to copy the command dev is interested in in one click instead of manually selecting it and copying with Ctrl+C.

### DIFF
--- a/docs/packages/lucide.md
+++ b/docs/packages/lucide.md
@@ -6,15 +6,19 @@ Implementation of the lucide icon library for web applications.
 
 ### Package Managers
 
-``` bash
+```sh
 npm install lucide
-#or
+```
+
+or
+
+```sh
 yarn add lucide
 ```
 
 ### CDN
 
-``` html
+```html
 <!-- Development version -->
 <script src="https://unpkg.com/lucide@latest/dist/umd/lucide.js"></script>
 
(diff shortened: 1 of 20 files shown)


docs/packages/lucide.md
```diff
@@ -98,8 +102,8 @@ createIcons({
   icons: {
     Menu,
     ArrowRight,
-    Globe,
-  },
+    Globe
```
karsa-mistmere, line 105: Wow, great work, that's a _huge_ list of improvements and I think most of them are definitely no-brainers e.g. splitting alternate scripts is definitely the way to go.

The only thing I'm not quite sure about is — you guessed it — the lack of trailing commas, especially in light of Lucide scripts themselves using a code style that _does_ use trailing commas.
I'm personally persuasible, although I perhaps do lean more towards having trailing commas if only because of cleaner diffs. What do you think, @ericfennis?

ericfennis: I think this is for a code example in the documentation. This should be code that is easy to read, "copy and pasteble" and with code styling that is closest to default javascript. I think trailing comas is not standard used in vanilla javascript. For the rest of the codebase, yes we should use trailing comas.

ericfennis (approved):

docs/packages/lucide.md
wojtekmaj: There changes were done automatically by Prettier on save. Can roll them back if you want me to!

PR #855, opened 2022-10-25
Lucide for compose

eltonkola:
I created a for the other day, and made a compose version of the library. Under the hood it is using a Kotlin script with some dependencies to transform each svg icon, into a ImageVector.

Take a look at it, and let me know if this is something you may want merged into the main project 

I tested it by using jitpack.io, and works on android and compose desktop (jvm)


ericfennis:
Hi @eltonkola,

Thanks for creating this software. And sorry for the late reply.
We discussed this if we want this in our repository. but we decided to not merge this in the Lucide repository because we think it is too much to maintain.
But don't hesitate to start a repo for this, we can link your repo in the README and on https://lucide.dev/packages page like we do with other implementations of lucide.

PR #856, opened 2022-10-26
Add missing key to map loops for site

GrimLink:
Currently when running the site it will throw warnings that the `map()` needs a key.

This pr will fix this by adding the missing keys to the loop.


ericfennis (approved):
Thanks!

PR #858, opened 2022-10-26
Add show more link for packages on Homepage

GrimLink:
This makes it more clear there are more options than just the ones shown on the homepage.

Also removes the Laravel icon from the homepage,
since it is already part of the 3de party section.


ericfennis:
Nice addition!

ericfennis (approved):

PR #860, opened 2022-10-26
Alchemy. Symbol of Gold.

Evgene-Kopylov:
Symbol of Gold.


karsa-mistmere:
I'm sorry, but no use case was provided and this would be a duplicate of https://lucide.dev/icon/circle-dot

PR #867, opened 2022-11-02
downgrade puppeteer to make angular build work

ekowbaah:
Downgraded the puppeteer version so the angular build can succeed


ericfennis:
tnx for the fix @ekowbaah !

PR #868, opened 2022-11-02
Improve "Why choose Lucide over Feather" section in README.md

locness3:
- Update number of added icons (congratulations to everyone btw, this is impressive)
- Mention official packages
- Expand on "Active community"

Sorry I forgot to fill the commit description, can you manage to do it for the merge commit?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,10 @@ Lucide is trying to expand the icon set as much as possible while staying faithf
 
 ### Why choose Lucide over Feather Icons
 
-- Lucide already expanded the icon set by 130+ in less than a year, so more icons to work with.
+- More icons to work with : Lucide has already expanded Feather by over 550 icons.
+- Official librairies and integrations with popular frameworks and design tools.
 - Well maintained code base.
-- Active community.
+- Active community, regularly growing and improving the set.
 
 ## Table of Contents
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,9 +17,10 @@ Lucide is trying to expand the icon set as much as possible while staying faithf
 
 ### Why choose Lucide over Feather Icons
 
-- Lucide already expanded the icon set by 130+ in less than a year, so more icons to work with.
+- More icons to work with : Lucide has already expanded Feather by over 550 icons.
```
karsa-mistmere, line 20: Nice catch, although I'd rather go with something more vague, so that we don't have to update this over and over and over again. :)
```suggestion
- More icons to work with: Lucide already has many times the icons Feather does.
```

locness3: maybe change "the icons" to "the amount of icons"

locness3: Or maybe, simpler : Lucide already has hundreds of icons more than Feather.

karsa-mistmere: > Or maybe, simpler : Lucide already has hundreds of icons more than Feather does.

Nice one, I like that, updated the suggestion accordingly.

locness3: Comitted.

karsa-mistmere (approved):

PR #876, opened 2022-11-09
Add site categories

ericfennis:


karsa-mistmere:
I've finally managed to look at this, and I'm not quite sure about this approach.

It looks clean, but:
1) doesn't quite work with paging which should be an absolute must for v1.0
2) will inevitably lead to duplicates in the icon list which should be undesirable
3) the list of categories is pretty hard to overview (which isn't a must, but should maybe still be something to strive for)

It looks pretty neat as a showcase, but maybe it's not very functional from a UX point of view, I think I'd rather stick with either a sidebar of the categories such as on https://fontawesome.com/search or a dropdown filter:
image

ericfennis:
@karsa-mistmere Sorry for the late reply.
I agree with you. I think we can make it a bit simpeler.
I think the font awesome approach is better. Also CSS.gg is also doing a great job. See their categories https://css.gg/app

PR #878, opened 2022-11-09
make `lucide-react` use `lucide` package instead of pre-generated files

Bricklou:
This PR change the way react manage icons, now it uses icon data from the `lucide` package instead of pre-generated ones
Component are still generated.


ericfennis:
Some of this we already discussed a bit in discord.

@karsa-mistmere and I had a chat about this. But we think this change is not worth it.

The overall idea is not bad, actually from an architectural perspective this is maybe better. For every build for each package we read out the icons directory and create icons nodes for each package this seems redundant. Why not centralize this, and import this in each package? (What this PR is about)

I will try to give some explanations why this is not worth it.

### 1. Will not really simplify the packages

When building each package we currently read the icons directory and inject the IconNode for each package. Instead, we will then import it from lucide. But to create all the exports for each icon in the main exports file we still need to read out the icons directory and create those exports. So this won’t really simplify the way the packages we release. Unless we change the API drastically as you mention with this:

```jsx
import { LucideIcon } from 'lucide-react'
import { Home } from 'lucide'

const MyComponent = () => {
	// This will not work with tree-shaking. Or a complex build-tool package is needed.
	return <LucideIcon name="home">
	// Or
	return <LucideIcon icon={Home}>
}
```

But that will cost DX and a Major API change. All the apps using the lucide will need to migrate. This is a no-go, for the success of this project.

### 2. This will add a lot of complexity to the build process. 

All our 10 packages will get a strict dependency on the “lucide” package. This means we can’t release the packages until the main lucide package is released. What if there is a bug in the lucide package? All the packages need to be updated. A lot of things we need to take into account when releasing. What we currently don’t have.

### 3. Problems with package managers.
    
This maybe is totally not an issue, but want to mention it as well. Currently, when someone is using ‘lucide-react’ it only installs one package. But with this change, they will need two. This probably will go totally fine. But what if there is a weird bug in imports or a version mismatch? Bugs and issues we would like to avoid.
    
### 4. Lucide package contains code that is not supported by all environments.

Lucide package was initially created as a library for browser clients to render icons. “document.createElement” is for example used. So some environments will not support this. For example “lucide-react-native” is a NodeJS environment, that will not be compatible with 'document'. (I can imagine we also create a separate package to maybe fix this, but still.)

### 5. Maybe a lot more problems we couldn’t think of will appear.

## To sum up

I think the way everything is set up now is simple and straightforward. Adding this will add unnecessary complexity to a simple icon library. We want to keep it as maintainable and simple as possible to work with for maintainers and contributors.

But as I said earlier the Idea is not bad! So @Bricklou I really appreciate the time and effort you put into it to explore this idea.

Issue #888, opened 2022-11-18
Unable to use with Nuxt components


vmdumitrache:
## Prerequisites

* Version: 0.101.0
* Are you running from source/main: Yes
* Are you using a released build: Yes
* Operating system: Ubuntu 22.04
* Bits: 64

## Step to reproduce

Follow installation instructions for `lucide-vue` for Nuxt setup by adding `'lucide-vue/nuxt'` to `buildModules`. In other words:
- `npm install lucide-vue`
- modify `nuxt.config.js` to include `'lucide-vue/nuxt'` as a `buildModule`
```js
export default {
  buildModules: ['lucide-vue/nuxt']
};
```

### Actual behavior
`npm run dev` fails.

## Any message or error

Cannot find module 'lucide-vue/nuxt' when running `npm run dev`.

PR #889, opened 2022-11-18
add nuxt.js to `package.json` files

vmdumitrache:
I've opened this PR in relation to this [issue](https://github.com/lucide-icons/lucide/issues/888).

I've noticed that the Nuxt components integration works as expected until 0.84.0 (inclusive). It seems that the only thing that changed since then was the addition of the `files` property in `/packages/lucide-vue/package.json`, which only contained `dist`.
I've updated it to also contian `nuxt.js`, assuming this is what prevents `pnpm build` from removing it.

I would like to apologise if my approach is incorrect - my experience with front-end ecosystems is very limited, but I'm more than happy to learn.

### DIFF
--- a/packages/lucide-vue/package.json
+++ b/packages/lucide-vue/package.json
@@ -19,7 +19,8 @@
   "unpkg": "dist/umd/lucide-vue.min.js",
   "sideEffects": false,
   "files": [
-    "dist"
+    "dist",
+    "nuxt.js"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


ericfennis (approved):
yess this is correct.
We forgot to add it when we introduced the files property in package.json.
Thanks for fixing this!

vmdumitrache:
An absolute pleasure and thank you for the awesome icons!

PR #892, opened 2022-11-24
Make solidjs package compatible with SSR and SolidStart

ottomated:
The current `lucide-solid` package uses `solid-js/h`, which [has several caveats](https://github.com/solidjs/solid/tree/main/packages/solid/h), including incompatibility with SolidStart.

This PR rewrites the solid package to offload the compilation to the end user, which is Ryan's recommendation:
image

This is tested and working in an example SolidStart app. However, it might break existing dependencies. Further testing is required.


ericfennis:
@ottomated Nice work will check it out!

ericfennis:
@ottomated I tested this with typescript but unfortunately it was still not working nicely.
I switched to using the [`rollup-preset-solid`](https://github.com/solidjs-community/rollup-preset-solid) from the solid community, so we can still use our rollup plugins, I implemented this in #899 with aliases support.

Thanks for inspiration to fix this for solid start. I'm closing this one.

PR #913, opened 2023-01-13
Svelte: Allow strings for `size` and `strokeWidth`

luca1197:
**Issue:**

When using the Svelte implementation, the `size` and `strokeWidth` are required to be numbers. Because of this, the values for these props need to be passed as a Javascript / TS expression, since values of props are strings by default. See these examples:

```svelte
<script lang="ts">
	import { X } from 'lucide-svelte'
</script>

<X size="3" strokeWidth="3"/>
<!-- Error also occurs when using <X size=3 strokeWidth=3/> -->
```

This example causes a Typescript error:
image

Using a Javascript expression fixes this issue, but is very inconvenient and feels unnecessary / unnatural:
```svelte
<script lang="ts">
	import { X } from 'lucide-svelte'
</script>

<X size={3} strokeWidth={3}/>
```

**Solution:**

Update the types of the `size` and `strokeWidth` props to also allow strings. This works fine in my testing, using my first example with the string values also works in in the end, but results in Typescript type errors. This change should not be breaking in any way, since it just allows one more type to be used.


ericfennis (changes requested):
I'm sorry for not getting back to you sooner.
We weren't sure if this was a welcome change. 
Looking at other Icon libraries they also support strings as values. So let's implement this. We currently still generating the distribution types file with a script. 
It's located under `packages/lucide-svelte/scripts/buildTypes.mjs`
There need these types to be added as well.

luca1197:
Thank you for the feedback, I added the types in that file as well. 👍

ericfennis (approved):
Awesome work!

PR #927, opened 2023-01-28
add battery-warning icon

Kaladii:
Can be useful when there's something wrong with the battery
image
image
battery-warning


ericfennis (approved):
@Kaladii Nice work!

PR #930, opened 2023-01-28
add picture-in-picture icon

cd16b:
- **Name of the icon:** picture-in-picture icon
- **What is the purpose of this icon?:** switch to picture-in-picture mode
- **Icon**: picture-in-picture3 (1)

Immagine 2023-01-28 202843


ericfennis:
@cd16b Nice work!
I think would be nice to have also the variant without the arrow.
This would be useful for example to indicate if picture-in-picture mode is active.

cd16b:
Added, let me know if the padding is ok.
- "icon:" picture-in-picture-active

Immagine 2023-01-29 121350

cd16b:
- **other design:**
Immagine 2023-01-29 131630

ericfennis:
Nice I like the last one. Can you try to increase the bigger rectangle with 1px at the bottom, then it will match with the `monitor` icon.

cd16b:
- **icon:** picture-in-picture-active
Immagine 2023-01-29 215302

ericfennis:
Nice very clean! Like this one

karsa-mistmere:
picture-in-picture
Hm... not totally sure, it's ever so slightly off center both to the left and to the top, but this is once again one of those edge cases where pixel perfection and aesthetics collide, and it can't be corrected for without damaging either the pixel perfectness or the visual weight of the icon. The bottom right icon shows a version where both pixel perfection and centre of gravity are so to speak optimal, but the icon itself is a bit meh. 😅

tags.json
```diff
@@ -3257,6 +3257,17 @@
   "phone-outgoing": [
     "call"
   ],
+  "picture-in-picture-active": [
```
karsa-mistmere, line 3260: I think these should be just two alternatives not "states", so no `picture-in-picture-active`, just `picture-in-picture` and  `picture-in-picture-2`.

ericfennis (approved):
nice work LGTM!

karsa-mistmere:
@ericfennis: I think we could do with some more tags, maybe:
`pop out`, `always on top`, `window`, `inset` and `multitask` (I think the first two for sure).

icons/picture-in-picture-2.svg
```diff
@@ -0,0 +1,13 @@
...
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="m14 13h6c1.108 0 2 0.892 2 2v3c0 1.108-0.892 2-2 2h-6c-1.108 0-2-0.892-2-2v-3c0-1.108 0.892-2 2-2zm7-4v-3c0-1.16-0.84-2-2-2h-15c-1.05 0-2 0.95-2 2v10c0 1.05 0.95 2 2 2h4" />
```
karsa-mistmere, line 12:
```suggestion
  <path d="M21 9V6a2 2 0 0 0-2-2H4a2 2 0 0 0-2 2v10c0 1.1.9 2 2 2h4" />
  <rect x="12" y="13" width="10" height="7" rx="2" />
```

PR #932, opened 2023-01-31
Add more tags to icons

wojtekmaj:
Adds more tags to icons:

* "exclamation mark" tag added to all icons with exclamation mark on them
* "accumulator", "charge" tags added to battery related icons
* "on", "off", "switch" tags added to power-off icon
* "switch" tag added to power icon


ericfennis (approved):
Thanks, nice work!

PR #933, opened 2023-02-01
Fixes `globe-2` icon

karsa-mistmere:
The current icon has a lot of misaligned paths (circled in blue) as well as a weird sharp corner (red):
image


ericfennis (approved):
Perfection 👌

PR #942, opened 2023-02-11
add title attr for "Available for" icons on home

alitnk:
Should make it easier to distinguish react and react-native - shows a tiny hint when you hover over it.


ericfennis (approved):
Good one!

PR #896, opened 2022-12-01
Deprecate tags.json

karsa-mistmere:
This PR will deprecate the use of tags.json

PR #946, opened 2023-02-13
Added github comment icon workflow

jguddas:
This automatically creates and updates a comment in PRs that shows a preview of all icons that have been changed or updated.

#### Performance
The action itself is quick due to not running any calculations or even starting up Node.
The API route is cashed and takes less than 200ms to return the rendered svg.

image

### Permission problems
This workflow only runs for PRs that are created from inside this repository.

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/


karsa-mistmere:
This looks pretty neat, although useful as it is, I think we'll still have to do some validations manually.
I think the white "midlines" will need to be a bit thinner.
What exactly do the colours represent?

jguddas:
The different colors are, so you can easily see when things are not 100% connected, like here in the cat ear for example.

image

ericfennis:
@jguddas Nice work, looks really cool!
I think this really helps the review process of icons.
Maybe we can add the `100%` scale variant of the icon somewhere as well.

I noticed that the builds are failing for site and the github action. Can you check what is going on?

karsa-mistmere:
@jguddas: do you think it could be possible to different shapes for points within a path and endpoints of paths?
I think that could be super useful for finding misaligned paths and checking if paths are incorrectly merged or broken up for no reason.

jguddas:
> @jguddas: do you think it could be possible to different shapes for points within a path and endpoints of paths?

Great idea.

There are 3 things:

1. Path elements
2. Path segments (the things split in `d` using `M`ove command)
3. Path commands.

I would like to highlight move commands by adding the big start/end circle.
What do you think?

I don't know how to best highlight the path elements.

- On the right side, the color is based on the path commands.
- On the left side, I changed it, so the colors are different for each Path element.

karsa-mistmere:
Nice, the one on the right seems much more legible to me apart from colours that are a bit too close in hue.
A few more questions:
* Why is the top control point from the head's circle missing? Is it a bug or is it on purpose?
* Is it maybe possible to differentiate closed and non-closed paths? (again, for the head's circle it is not important which point is the starting point, since it's closed)

jguddas:
@ericfennis I don't have access to the build, but I can tell you why the action is failing.

The action fails due to being run from a fork, I will see if I can find a quick fix.

jguddas:
> Why is the top control point from the head's circle missing? Is it a bug or is it on purpose?

It's due to being a circle element instead of a path, what do you think of only showing control points for path elements?

> Is it maybe possible to differentiate closed and non-closed paths?

Good idea, I think it would definitely be nice to only show the start/end node for open paths.

jguddas:
- I have removed the start and end marker for closed paths.
- I have removed the control points from non path elements (see head).
- I have changed it, so the control points are only shown for start and end of path segments.
- I have submitted a pr to svgo to optimize closed paths https://github.com/svg/svgo/pull/1751
- I have submitted an issue to svgo to optimize arcs (like here in the wheel chair wheel) https://github.com/svg/svgo/issues/1748

site/src/pages/api/gh-icon/[data].ts
```diff
@@ -0,0 +1,278 @@
...
+    }
+  }
+
+  res.setHeader('Content-Type', 'image/svg+xml');
+  res.status(200).end(
+    [
+      '<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">',
+
+      // grid
+      '<rect x=".1" y=".1" width="23.8" height="23.8" rx="1" fill="#fff" stroke="#E2E8F0" stroke-width=".1"/>',
```
Comment (karsa-mistmere) [inline site/src/pages/api/gh-icon/[data].ts]: I think it might be a good idea to extract all of these styling attributes to dedicated CSS classes.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,51 @@
...
+        id: generate-comment
+        run: |
+          delimiter="$(openssl rand -hex 8)"
+          echo "body<<$delimiter" >> $GITHUB_OUTPUT
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            cat "$file" | # get file content
+            tr '\n' ' ' | # remove line breaks
+            sed -e 's/<svg[^>]*>/<svg>/g' | # remove attributes from svg element
+            base64 -w 0 | # encode svg
+            sed "s|.*|<img width=\"50%\" alt=\"$file\" src=\"https://lucide.dev/api/gh-icon/&.svg\"/>|"
```
karsa-mistmere, line 31: Not totally sure about creating an API endpoint for this purpose, couldn't we just as easily delegate the generation to a script and embed the base64 encoded SVG files into the `src` attribute?

```html
<img src="data:image/svg+xml;base64,${encodedSvgContent}"/>
```

That way we also would only have to add the dependencies as dev deps.

jguddas: Sadly, GitHub does not support `data`-URLs.

karsa-mistmere: Oh, damn, it does indeed not. :(
And evidently Github's API [doesn't allow uploading images either](https://github.com/peter-evans/create-or-update-comment/issues/68). :(

site/src/pages/api/gh-icon/[data].ts
Comment (jguddas) [reply site/src/pages/api/gh-icon/[data].ts]: I do think we should extract it, not sure how well CSS inside SVG is supported.
```ts
const styles = { gridStrokeWidth: 0.1 }
svgson.stringify({ name: 'rect', attributes: { x: styles.gridStrokeWidth, … } })
```

.github/workflows/pull-request.yml
jguddas: Yes and it's a bit of a security issue if we would run this inside public forks.
Maybe in the future we can have a create/upload icon button on the website that shows you a preview and possible errors.

site/src/pages/api/gh-icon/[data].ts
Comment (jguddas) [reply site/src/pages/api/gh-icon/[data].ts]: This feels like a good solution, what do you think?
```tsx
import { renderToString } from "react-dom/server";

const Grid = ({ fill, stroke, strokeWidth, radius }: { strokeWidth: number, stroke: string, fill?: string, radius?: number }) => (
  <>
    <rect
      width={24 - strokeWidth / 2}
      height={24 - strokeWidth / 2}
      x={strokeWidth / 2}
      y={strokeWidth / 2}
      fill={fill}
      stroke={stroke}
      strokeWidth={strokeWidth}
      rx={radius}
    />
    <path
      stroke={stroke}
      strokeWidth={strokeWidth}
      d={new Array(Math.floor(24 - 1))
        .fill(null)
        .flatMap((_, i) => [
          `M${strokeWidth / 2} ${i + 1}h${24 - strokeWidth}`,
          `M${i + 1} ${strokeWidth / 2}v${24 - strokeWidth}`,
        ])
        .join("")}
    />
  </>
);

console.log(
  renderToString(
    <svg
      width={24}
      height={24}
      viewBox="0 0 24 24"
      fill="none"
      stroke="currentColor"
      strokeWidth={2}
      strokeLinecap="round"
      strokeLinejoin="round"
    >
      <Grid strokeWidth={.1} stroke="#e2e8f0" fill="#fff" radius={1} />
    </Grid>
  )
);
```

.github/workflows/pull-request.yml
karsa-mistmere: A possible alternative approach I found through some Googling: we can add a dedicated repository branch and upload the image files there via the [repository contents API](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#create-or-update-file-contents).

My greatest issue with this approach is that we can't really cache dynamic content on the site, as it became evident in #896.

@ericfennis: what do you think?

site/src/pages/api/gh-icon/[data].ts
Comment (karsa-mistmere) [reply site/src/pages/api/gh-icon/[data].ts]: `<style>` elements within SVG files are [fully supported](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/style).
Not sure about the code example above, could we not simply use lines instead of rectangles?

Comment (jguddas) [reply site/src/pages/api/gh-icon/[data].ts]: You are right, styles are supported :+1:

.github/workflows/pull-request.yml
jguddas: There shouldn't be any caching problems since this doesn't use the file system in any way, right?

karsa-mistmere: Well, the issue is exactly that it _should_ be cached, so that the site doesn't have to generate it over and over again.

Adding an appropriate cache control header preferably with an obscenely large `max-age` should in theory work: https://vercel.com/guides/how-to-configure-the-cache-control-response-header-in-vercel-projects#using-edge-functions

jguddas:
I moved the whole thing into a react component :tada: 
image

.github/workflows/pull-request.yml
ericfennis: Cache control will work if the content for each request changes, it will bypass the settled cache. So I think it's ok to use our API for it.

jguddas:
That should be everything.

jguddas:
@ericfennis can I get your okay?

ericfennis:
@jguddas Sorry for the late reply, was a bit busy.
I think this is still a great tool for reviewing icons.
I really want to test it before we merge this, maybe we can add a dummy icon in this PR to see it? 
I notice that the Vercel build is failing, you need to update the pnpm lockfile. Run `pnpm install`.
Build error:
image

Also the `site/api/gh-icon.ts` file is missing in the PR, is it removed?

jguddas:
Well I tried to test it with the preview deployment and it doesn't work.

https://lucide-btt1rfnxj-lucide.vercel.app/api/gh-icon/PHN2Zz4gICA8cGF0aCBkPSJNMiAxN2E1IDUgMCAwIDAgMTAgMGMwLTIuNzYtMi41LTUtNS0zLTIuNS0yLTUgLjI0LTUgM1oiIC8+ICAgPHBhdGggZD0iTTEyIDE3YTUgNSAwIDAgMCAxMCAwYzAtMi43Ni0yLjUtNS01LTMtMi41LTItNSAuMjQtNSAzWiIgLz4gICA8cGF0aCBkPSJNNyAxNGMzLjIyLTIuOTEgNC4yOS04Ljc1IDUtMTIgMS42NiAyLjM4IDQuOTQgOSA1IDEyIiAvPiAgIDxwYXRoIGQ9Ik0yMiA5Yy00LjI5IDAtNy4xNC0yLjMzLTEwLTcgNS43MSAwIDEwIDQuNjcgMTAgN1oiIC8+IDwvc3ZnPiA=.svg

But replacing `https://lucide-btt1rfnxj-lucide.vercel.app` with `localhost:3000` does work :thinking:.

jguddas:
Now it works.

https://github.com/jguddas/github-icon-worflow-test/pull/3

site/src/components/SvgPreview/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+        xmlns="http://www.w3.org/2000/svg"
+        width={24}
+        height={24}
+        viewBox="0 0 24 24"
+        fill="none"
+        stroke="currentColor"
+        strokeWidth={2}
+        strokeLinecap="round"
+        strokeLinejoin="round"
+      >
```
ericfennis, line 142:
```suggestion
const SvgPreview = React.forwardRef<SVGSVGElement, { src: string; showGrid?: boolean, style?: React.CSSProperties }>(
  ({ src, showGrid = false, style }, ref) => {
    const paths = getPaths(src);
    return (
      <svg
        ref={ref}
        xmlns="http://www.w3.org/2000/svg"
        width={24}
        height={24}
        viewBox="0 0 24 24"
        fill="none"
        stroke="currentColor"
        strokeWidth={2}
        strokeLinecap="round"
        strokeLinejoin="round"
        style={style}
      >
```

site/src/pages/api/gh-icon/[data].tsx
```diff
@@ -0,0 +1,14 @@
+import SvgPreview from '../../../components/SvgPreview';
+
+export default async function handler(req, res) {
+  // ReactDOMServer needs to be imported dynamically
+  // https://github.com/vercel/next.js/issues/43810
+  const ReactDOMServer = (await import('react-dom/server')).default;
+
+  const src = Buffer.from(req.url.split('/').at(-1).slice(0, -4), 'base64').toString('utf8');
+
+  const svg = Buffer.from(ReactDOMServer.renderToString(<SvgPreview src={src} showGrid />));
```
Comment (ericfennis) [inline site/src/pages/api/gh-icon/[data].tsx:10]:
```suggestion
  const svg = Buffer.from(ReactDOMServer.renderToString(
    <SvgPreview 
      src={src}
      showGrid
      style={{
        background: '#fff'
      }}
    />
  ));
```

ericfennis (changes requested):
Nice work @jguddas !

Looking at you PR, In notices that the icon is hard to see in darkmode github.
image

I've added some suggestions to add a background.
Also we should set the cache headers, so vercel will cache our requests.

ericfennis:
Also how do we deal with a PR with multiple icons? 🤔

jguddas:
How it also works in dark mode.

https://github.com/jguddas/github-icon-worflow-test/pull/10

ericfennis (approved):
@jguddas Nice this looks event better!
Nice work. Good to go for me

karsa-mistmere:
@jguddas: Really nice, I love the latest changes (dark mode support + extra 1px outline to check for gaps). :)

karsa-mistmere:
@jguddas: I think this is almost good to go, but I still strongly believe CSS classes should be used to style the paths, since there are a looooot of repeating attribute values, and dark mode support should be handled not by using a `<picture>` tag and generating separate preview files, but via CSS `prefers-color-scheme` media queries embedded within the preview files.
(It could also be a good idea to insert HTML comments before each section, stating the purpose of that group of paths (grid, centerlines, outlines, points, arcs etc.).

jguddas:
It would be nice to not have to use the picture tag, but I'm not sure how I would go about cleanly integrating style tags.

What about getting rid of the duplications by using groups like this?

```diff
- <p d="…" strokeWidth={5} />
- <p d="…" strokeWidth={5} />
+ <g strokeWidth={5} >
+   <p d="…" />
+   <p d="…" />
+ </g>
```

karsa-mistmere:
Yeah, grouping is also a good idea, although I'd probably go for BEM style CSS classes instead of using the `id` attribute, should the SVGs ever need to be embedded in an HTML page (some debug page on the app review site maybe?) where the global IDs would no longer be unique.

ericfennis (approved):

karsa-mistmere:
I've posted a PR for this branch that solves dark mode support via CSS only, eliminating the need for building separate icons for light and dark modes.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,55 @@
...
+        id: generate-comment
+        run: |
+          delimiter="$(openssl rand -hex 8)"
+          echo "body<<$delimiter" >> $GITHUB_OUTPUT
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            cat "$file" | # get file content
+            tr '\n' ' ' | # remove line breaks
+            sed -e 's/<svg[^>]*>/<svg>/g' | # remove attributes from svg element
+            base64 -w 0 | # encode svg
+            sed "s|.*|<picture><source media=\"(prefers-color-scheme: dark)\" srcset=\"https://lucide.dev/api/gh-icon/&-dark.svg\"/><source media=\"(prefers-color-scheme: light)\" srcset=\"https://lucide.dev/api/gh-icon/&.svg\"/><img width=\"48%\" title=\"$file\" alt=\"$file\" src=\"https://lucide.dev/api/gh-icon/&.svg\"/></picture> |"
```
karsa-mistmere, line 35: Why does this have to be 48%?

jguddas: Both work, but the bottom one looks better with the spacing between the images.
```html
<img w="50%"/><img w=50%/>
<img w="48%"/> <img w=48%/>
```

karsa-mistmere: I see. I wonder if some degree of responsivity could be achieved, but can't really see a very obvious solution. 🤔

karsa-mistmere: This is kind of hacky, but `width="400"` currently works kind of OK (on desktop it's two icons in a row, on mobile it's 1 icon at full width).

karsa-mistmere:
@jguddas: any chance of also adding a curvature preview in v2.0? 😅🤪

image
as in https://www.figma.com/community/plugin/932729267119671976/Curvature

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,55 @@
...
+        id: generate-comment
+        run: |
+          delimiter="$(openssl rand -hex 8)"
+          echo "body<<$delimiter" >> $GITHUB_OUTPUT
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            cat "$file" | # get file content
+            tr '\n' ' ' | # remove line breaks
+            sed -e 's/<svg[^>]*>/<svg>/g' | # remove attributes from svg element
+            base64 -w 0 | # encode svg
+            sed "s|.*|<img width=\"48%\" title=\"$file\" alt=\"$file\" src=\"https://lucide.dev/api/gh-icon/&.svg\"/> |"
```
jguddas, line 35:
```suggestion
            sed "s|.*|<img width=\"400\" title=\"$file\" alt=\"$file\" src=\"https://lucide.dev/api/gh-icon/&.svg\"/> |"
```

jguddas:
An edit page that makes it easy for people to contribute by showing them errors and a preview of their icon would be super nice IMO.

karsa-mistmere (approved):

jguddas:
Okay @karsa-mistmere, @ericfennis lets merge this!

jguddas:
Okay @karsa-mistmere, @ericfennis lets merge this!

PR #952, opened 2023-02-16
Add split-square icon

Patchethium:
Add the `split block` icon in VSCode's `Jupyter Notebook` mode.

This icon is used in splitting current text cell into two cells at the cursor position.

Screenshot 2023-02-17 00 24 21

Screenshot 2023-02-17 00 25 10

Hope I got the tags right!


ericfennis:
@Patchethium Nice icon!
I can imagine a horizontal version is useful as well.
So instead of `split-square` we get `split-square-horizontal` and split-square-vertical`.

Patchethium:
Should be okay now.

ericfennis (changes requested):
Nice work!
One final thing in the SVG code.

Patchethium:
Should be okay now.

ericfennis (approved):
Awesome, good to go!

PR #956, opened 2023-02-24
Add top-hat icon

ArrowAced:
I made a basic top hat icon. Hopefully it's okay. What do you think?
top-hat
top-hat 2


ArrowAced:
Some possible adjustments, don't know which one I like the best
image

ericfennis:
Hi @ArrowAced,

Can you describe a good use case for this icon?

ericfennis:
Close this due to inactivity. I will reopen this one if a good use case is given

PR #980, opened 2023-03-13
add `expand` and `collapse` icons

greg-hoarau:
---
name: New icon
about: Add a new icon to the library
labels: "🎨 <icon"
---

- **Name of the icon** : expand-[right | left | up | down] and collapse[right | left | up | down]
- **Tags (alternative names for this icon)** (add them in as a separate json file using the same icon name) :
- **What is the purpose of this icon?** : Shows that one can click it to expand or collapse an area (example: sidenav)
- **100% scale preview** : 
COLLAPSE_DOWN
EXPAND_DOWN
COLLAPSE_UP
EXPAND_UP
COLLAPSE_RIGHT
EXPAND_RIGHT
COLLAPSE_LEFT
EXPAND_LEFT

- **Have you considered alternative possibilities** for its naming or design? :

### DIFF
--- a/icons/collapse-down.json
+++ b/icons/collapse-down.json
@@ -1,0 +1,12 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "arrow",
+    "collapse",
+    "fold",
+    "vertical"
+  ],
+  "categories": [
+    "arrows"
+  ]
+}
--- a/icons/collapse-down.svg
+++ b/icons/collapse-down.svg
@@ -1,0 +1,13 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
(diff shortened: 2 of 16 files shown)


jguddas:
Not sure, feels a lot like got to last/first, I would assume a collapse being just a simple chevron.

ericfennis (commented):
Like the icons! Clean and simple design.
I'm not sure about the name, maybe we can think of a more generic name for this set of icons. Because some of them have multiple meanings/usecases. (That's good!)

For example, the `collapse-down` icons could also be used for "downloads"
And `collapse-right` for the keyboard `tab` key.

Maybe we can go for `arrow-right-line-left` or something.
We can add "aliases" and "tags" for all the icons for all the usecases.

karsa-mistmere:
Closing this, as this has been superseded by #1158 which has more generic icon names.

PR #231, opened 2021-02-23
add settings-2 icon

mittalyashu:
| Preview |
| --- |
| CleanShot 2021-02-23 at 19 44 20@2x |

PR #987, opened 2023-03-28
Fix `settings-2`, improve metadata, and add more `settings-*` alternates

danielbayley:
Fixes to #231, plus metadata and alternates…

settings-2-vertical

settings-3

settings-3

### DIFF
--- a/icons/diamond.svg
+++ b/icons/diamond.svg
@@ -9,12 +9,5 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect
-    width="15.56"
-    height="15.56"
-    x="12"
-    y="1"
-    rx="2.41"
-    transform="rotate(45 12 1)"
-  />
+  <path d="M2.7 10.3a2.41 2.41 0 0 0 0 3.41l7.59 7.59a2.41 2.41 0 0 0 3.41 0l7.59-7.59a2.41 2.41 0 0 0 0-3.41l-7.59-7.59a2.41 2.41 0 0 0-3.41 0Z" />
 </svg>


karsa-mistmere:
We maintainers have talked and feel like adding this variant as a standalone icon does not seem justified, as they are virtually identical.

However, I feel like none of these should have actually been named "settings" in the first place. 🙄 
`settings` should have been `cog`
`settings-2` should have been `sliders-2` or  `sliders-{something}` at the very least.

@ericfennis: can we rename these and add aliases?
I'm not sure which to keep out of `settings-2` and `settings-3`, I actually prefer this one, but that's pretty subjective. :)

ericfennis:
@karsa-mistmere Yeah we could do that, but add aliased for `settings-{number}` is a bit tricky. Because what will happen if we add another version of settings icon. Then we have a conflict with this alias. but I agree this icon should be named sliders. 

Looking at all the other settings, sliders, cogs icons. It's a bit of a mess. This should be nice to clean this up. For example to maybe rename all the "settings" icons because the name is not very generic, but that will be a breaking change. Maybe we could do this for lucide 1.0.

For this PR:
I agree with @karsa-mistmere, maybe it is better to replace the `settings-2` icon. This one looks better I think.

danielbayley:
> This one looks better I think.

I actually really like the existing one 🙈 😅

Are we dismissing this version having both `vertical` and `horizontal` variants?

I agree it’s a mess between `settings`, `sliders` and `cog`… Do we agree that:

* The existing `sliders` should be `sliders-vertical`
* This (`settings-2`) should be `sliders-2` (or `sliders-`[`horizontal`/`vertical`]`-2`)
* `settings` should be `cog`
* `cog` should be `cog-2`

?

karsa-mistmere:
My two cents:

- `settings` should be `gear`
- `sliders` should be `sliders-vertical`
- `settings-2` should be something like `sliders-large-vertical` (but I still prefer your design, it makes it more obvious that they're sliders, not circles on sticks
- `cog` should either remain `cog` or if we want to be semantically correct, it could become `cogwheel`

danielbayley:
> My two cents:
> 
> * `settings` should be `gear`
> * `sliders` should be `sliders-vertical`
> * `settings-2` should be something like `sliders-large-vertical`
>   (but I still prefer your design, it makes it more obvious that they're sliders, not circles on sticks)
> * `cog` should either remain `cog` or if we want to be semantically correct, it could become `cogwheel`

@karsa-mistmere See #1300 …

PR #988, opened 2023-03-28
Add alternative `bug` icons

danielbayley:
Some variations…

bugs


karsa-mistmere:
Nice designs, but as with #987, adding both is simply not justified, they're just far too similar.

I do emphasise, since I myself don't really like the current bug icon and pretty much always use this alternate design as a manual entry instead. :P 
bug-2

danielbayley:
> adding both is simply not justified, they're just far too similar.

@karsa-mistmere Yeah, I think we should settle on just one version, so maybe replace the existing bug icon? Just a single commit now…

icons/debug.svg
```diff
@@ -0,0 +1,21 @@
...
+>
+  <path d="m10 3 1 2" />
+  <path d="m14 3-1 2" />
+  <path d="M20 5c0 1.3-.6 2.4-1.6 3.2" />
+  <path d="M8 9C5.8 9 4 7.2 4 5" />
+  <line x1="4" x2="8" y1="12" y2="12" />
+  <path d="M18.4 15.8c1 .7 1.6 1.9 1.6 3.2" />
+  <path d="M4 19c0-2.2 1.8-4 4-4" />
+  <polygon points="12,15 17,12 12,9" />
+  <path d="M14.1 18.4c-.6.4-1.3.6-2.1.6h0c-2.2 0-4-1.8-4-4V9c0-2.2 1.8-4 4-4h0c.7 0 1.4.2 2 .6" />
```
karsa-mistmere, line 20: Hah, that's actually pretty neat how the play icon fits exactly within. :D

It's 1px higher up than the regular `bug` icon though:
```suggestion
  <path d="m10 4 1 2" />
  <path d="m14 4-1 2" />
  <path d="M20 6c0 1.3-.6 2.4-1.6 3.2" />
  <path d="M8 10c-2.2 0-4-1.8-4-4" />
  <path d="M4 13h4" />
  <path d="M18.4 16.8c1 .7 1.6 1.9 1.6 3.2" />
  <path d="M4 20c0-2.2 1.8-4 4-4" />
  <path d="m12 16 5-3-5-3v6Z" />
  <path d="M14.1 19.4A4.01 4.01 0 0 1 8 16v-6c0-2.2 1.8-4 4-4 .7 0 1.4.2 2 .6" />
```

ericfennis:
Hmm I think not agree with replacing the existing bug icon. They are similar if you look them in scaled version but on 100% they look different. So maybe reconsider this?

danielbayley:
> Hmm I think not agree with replacing the existing bug icon. They are similar if you look them in scaled version but on 100% they look different. So maybe reconsider this?

I mean the existing bug icon is pretty weak compared to any of [these](https://github.com/lucide-icons/lucide/pull/988#issue-1644236855)… The `debug` idea should be assessed separately…

karsa-mistmere:
While I do personally like the more organic look, there's still the issue of the legs being too close to each other resulting in too high of a density near the body, which is especially present at higher stroke widths & smaller sizes:
image

danielbayley:
> While I do personally like the more organic look, there's still the issue of the legs being too close to each other resulting in too high of a density near the body, which is especially present at higher stroke widths & smaller sizes: image

Addressed in the latest commit…

karsa-mistmere (approved):

ericfennis:
Can we add this one as `bug-2`?

danielbayley:
> Can we add this one as `bug-2`?

Do you mean `debug`, or having `bug`, `bug-2` and `debug`?

danielbayley:
> Can we add this one as bug-2?
> @danielbayley yes and maybe should have`debug-2` as well

Done and done.

ericfennis (commented):
@danielbayley I think, `bug` and `bug-2` should use the same body.
image

Maybe the `debug` icons are a bit too dense, with the play on the body.
I created an alternative design. I'm curious about what you think.
image

danielbayley:
> `bug` and `bug-2` should use the same body.

Fixed.

> Maybe the `debug` icons are a bit too dense, with the play on the body.

Hmm, they do maintain the 2px spacing between elements…

danielbayley:
I personally prefer the version with the play icon in the middle, but see what you guys think… @karsa-mistmere @ericfennis @jguddas Also added a `bug-off` variation.

jguddas:
+1 for having the play triangle in the middle.

IMO neither the straight nor the curved arms feel super awesome.

danielbayley:
>  neither the straight nor the curved arms feel super awesome.

They can’t be neither straight nor curved 😅

I much prefer the curved top & bottom legs version, personally…

> some antennae?

Already has those… Do you mean make them longer?

danielbayley:
Just completely remodelled these. See what you guys think…?

ericfennis:
Nice you tried to experiment more. But I'm not in a favor of replacing the current bug icon. Because I think it's minimalistic and simple what the lucide style is. I'm open to new variants. I think I liked the version before these ones.

I think the `bug-off` is a nice icon, I'm not sure about the use cases for `x` and `chevron` icons. So let's keep them out.

danielbayley:
I'm not sure about the use cases for `x` and `chevron` icons.

`bug-x` is `debug`, and `bug-chevron` is `debug-console` (could also be `debug-alt`) in [codicons](https://code.visualstudio.com/api/references/icons-in-labels#icon-listing)…

danielbayley:
> I think I liked the version before these ones.

@ericfennis Reverted the last commit. Just need to agree on which debug/play variation… @karsa-mistmere @jguddas

karsa-mistmere:
@danielbayley: The play shape violates the guidelines and I'm not a fan of the chevron variant.

danielbayley:
> @danielbayley: The play shape violates the guidelines and I'm not a fan of the chevron variant.

That set was reverted…

karsa-mistmere:
> That set was reverted…

Oh, my bad, I only checked out the updated first post.

---

I still think this is a case of "TMI", i.e. "too many [damn] icons": the two bug shapes are just way too similar to justify having both, but we seem to be at an impasse; I do believe an update of sorts is justified (not necessarily these updates, but some adjustment of size and centre of gravity) while @ericfennis seems to be against it. 😄 

Let's keep this open for now and see if we can gather some more input from the community? 🤷‍♂️

danielbayley:
> we seem to be at an impasse

Yeah, it’s annoying because this is one of the most prominent icons, permanently in the VS Code sidebar…

Screenshot 2023-04-25 at 5 34 15 pm

> the two bug shapes are just way too similar to justify having both

Yeah, ideally any variations would be substantially different enough to warrant adding a `-2`…
Consensus is hard sometimes 🙈

I do like some of my (substantially different) [alternatives above](https://github.com/lucide-icons/lucide/pull/988#issue-1644236855)… :)

danielbayley:
> 
> 
> I like those a lot more

What if we keep the existing `bug` icon, adding the `debug` and `bug-off` versions of it, and add some of these as the `bug-*-2` variation… Would that be an acceptable compromise? @ericfennis @karsa-mistmere @jguddas

jguddas:
I think we should get rid of the old bug and use the new one for everything.

danielbayley:
> I think we should get rid of the old bug and use the new one for everything.

Agreed. @jguddas Something very strange going on with the comment bot… https://github.com/lucide-icons/lucide/pull/988#issuecomment-1498280167

jguddas:
> > I think we should get rid of the old bug and use the new one for everything.
> 
> Agreed. @jguddas Something very strange going on with the comment bot… [#988 (comment)](https://github.com/lucide-icons/lucide/pull/988#issuecomment-1498280167)

Can you resolve the conflicts and see if that helps @danielbayley?

danielbayley:
> https://github.com/lucide-icons/lucide/pull/988#issuecomment-1498280167
> Can you resolve the conflicts and see if that helps @danielbayley?

@jguddas That seems to have partially fixed it, but still something strange going on with `bug-play`…

ericfennis:
I want to vote on this one. Everyone can vote with a reaction emoji.

image

ericfennis:
Ok, current design it is!

ericfennis:
I've one final request for the `bug-x` icon. I think the X in the center of the bug is not really communicating the "remove" on this one. I think we should place the X on the same location as the `bug-play` icon.

danielbayley:
> I want to vote on this one. Everyone can vote with a reaction emoji.

Nice! We need this feature in comment bot form though… :)

> I've one final request for the bug-x icon. I think the X in the center of the bug is not really communicating the "remove" on this one. To me it looks a bit like it's part of the bug. I think we should place the X on the same location as the bug-play icon.

@ericfennis I _much_ prefer the x in the centre… But if you insist I have added the same variant pair as [`file-x`](https://lucide.dev/icons/file-x)[`-2`](https://lucide.dev/icons/file-x-2) etc.

jguddas:
`x` feels to small, can we split the `bug-x` of into a separate PR and get the rest merged?

ericfennis:
@danielbayley Hmm yeah but to me, the X not communicating the right thing here.
And I agree with @jguddas X is also too small.

jguddas:
Would be great if you could create a PR for `bug-x` than we can see the low dpi preview for the x.

PR #991, opened 2023-03-28
Add `atom`ic icons

danielbayley:
atom


karsa-mistmere:
Oh, cool, I wanted to add an atom icon a while back, but my design was always too dense because of the 6-fold rotational symmetry, and since I didn't think it could work (i.e. still be recognisable as an atom) with 4-fold symmetry, I didn't even try. What a bummer, it very clearly does work. 😅 

One thing though, I think we should go for a perfectly symmetric version (this also solves the issue of the <2px gap between the atom core and the electron cloud):
atom

danielbayley:
> I think we should go for a perfectly symmetric version

@karsa-mistmere Do you mean make the electrons orbit at a 90° angle to each other? (they’re currently ~45°…)

karsa-mistmere:
> @karsa-mistmere Do you mean make the electrons orbit at a 90° angle to each other? (they’re currently ~45°…)

I mean like this (having 4-fold rotational symmetry instead of just 2-fold):
atom

icons/atom.svg
```diff
@@ -0,0 +1,15 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="12" cy="12" r="1" />
+  <ellipse transform="matrix(0.7071 -0.7071 0.7071 0.7071 -4.9706 12)" cx="12" cy="12" rx="5.4" ry="12" />
+  <ellipse transform="matrix(0.7071 -0.7071 0.7071 0.7071 -4.9706 12)" cx="12" cy="12" rx="12" ry="5.4" />
```
karsa-mistmere, line 14: While pretty neat, I'm afraid we don't support transforms, so they'll have to be removed. :(
```suggestion
  <path d="M20.2 20.2c2.04-2.03.02-7.36-4.5-11.9-4.54-4.52-9.87-6.54-11.9-4.5-2.04 2.03-.02 7.36 4.5 11.9 4.54 4.52 9.87 6.54 11.9 4.5Z" />
  <path d="M15.7 15.7c4.52-4.54 6.54-9.87 4.5-11.9-2.03-2.04-7.36-.02-11.9 4.5-4.52 4.54-6.54 9.87-4.5 11.9 2.03 2.04 7.36.02 11.9-4.5Z" />
```

icons/atom.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "atomic",
+    "nuclear",
+    "physics"
```
karsa-mistmere, line 6:
```suggestion
    "physics",
    "particle",
    "element",
    "molecule"
```

karsa-mistmere (changes requested):

danielbayley:
@karsa-mistmere Also, a couple of alternates…

Artboard 1@30x

karsa-mistmere:
We don't support fills, so there's that. 😅 

Not sure about the alternate design you pushed, I'd more likely think of that as orbiting planets not as an atom.

danielbayley:
> We don't support fills

They aren’t fills 😉 (those are just 2 strokes…)

karsa-mistmere:
> They aren’t fills 😉 (those are just 2 strokes…)

That's kind of even worse, since we do want to keep support for customizable stroke width. :)

danielbayley:
> > They aren’t fills 😉 (those are just 2 strokes…)
> 
> That's kind of even worse, since we do want to keep support for customizable stroke width. :)

I see, makes sense… Just the top row is relevant then… Could be renamed as `orbit` and moved to a different PR…?

ericfennis (approved):

PR #994, opened 2023-03-28
Add `slash`/`dot-square` icons and `tags` for `git diff`

danielbayley:
dot-square

slash-square


karsa-mistmere:
Arrows are pretty generic, so I think adding those are a no brainer, but can you provide use cases for the first two?

Another issue: please make sure that all squares are represents by `<rect>` tags.

danielbayley:
> can you provide use cases for the first two?

Sure, `dot-square` is [diff-modified](https://primer.style/design/foundations/icons/diff-modified-16) and `slash-square` is [diff-ignored](https://primer.style/design/foundations/icons/diff-ignored-16).

@karsa-mistmere As an aside, I think the naming convention should be changed to `square-*`/`circle-*` etc…

icons/square-equal.svg
```diff
@@ -0,0 +1,15 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect width="18" height="18" x="3" y="3" rx="2" ry="2" />
+  <line x1="8" x2="16" y1="9" y2="9" />
+  <line x1="8" x2="16" y1="15" y2="15" />
```
karsa-mistmere, line 14: Let's use the same equals sign as in https://github.com/lucide-icons/lucide/pull/1032
```suggestion
  <path d="M7 10h10" />
  <path d="M7 14h10" />
```

karsa-mistmere (changes requested):
If we add git as a tag to some of these icons, they should probably also appear in the `development` category.

karsa-mistmere:
@danielbayley: Do you have `pnpm` installed and running?
I'm pretty sure these should have been modified by the `lint-staged` pre commit hook. 🤔 
(polylines to paths)

danielbayley:
> Do you have `pnpm` installed and running?

I do now (and am in fact contributing to that project)… I probably just hadn’t set it up before some of these earlier PRs; I’ll run over them manually…

icons/arrow-right-square.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "git",
+    "diff",
+    "renamed",
```
karsa-mistmere, line 6:
```suggestion
```

Not sure about these tags, they only apply to a very specific use case, let's keep this generic.

icons/plus-square.json
```diff
@@ -1,6 +1,9 @@
 {
   "$schema": "../icon.schema.json",
   "tags": [
+    "git",
+    "diff",
+    "plus",
```
karsa-mistmere, line 6:
```suggestion
```

As above (and plus is part of the name).

icons/minus-square.json
```diff
@@ -1,6 +1,9 @@
 {
   "$schema": "../icon.schema.json",
   "tags": [
+    "git",
+    "diff",
+    "minus",
```
karsa-mistmere, line 6:
```suggestion
```

As above (and minus is part of the name).

icons/square-dot.svg
karsa-mistmere, line 1: This is almost the same as `dice-1`, so not totally sure.

icons/square-equal.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "calculate",
+    "maths"
```
karsa-mistmere, line 5:
```suggestion
    "calculate"
```

Just the category will suffice in the future I think.

icons/square-slash.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "git",
+    "diff",
+    "ignored"
```
karsa-mistmere, line 6: While this is a nice icon to have, I'm not sure how universal these use cases are. 🤔

karsa-mistmere (changes requested):

icons/square-equal.json
danielbayley: So does the search function treat `categories` and `tags` (and `aliases`, for that matter) the same then?

icons/square-slash.json
danielbayley: What do you mean exactly? Well it is `diff-ignored` in both octicons and codicons… Are you proposing to remove these `tags`? Shouldn’t we be aiming to make search as useful as possible?

icons/square-dot.svg
danielbayley: Well this was taken from https://lucide.dev/icons?search=circle-dot, which I figured was the direct equivalent, but I see what you mean… Although that dot is _very_ small 😖 (I understand it needed to be to match the other `dice`)

icons/square-equal.json
karsa-mistmere: Not sure about right now, but it certainly should and it will in the future (as well as display both in the list).

icons/square-slash.json
karsa-mistmere: Search should be useful, I'm just not entirely sure if this shape truly has  this universal meaning.

It's especially funny for GitHub to call this "diff ignored", when in actuality they're using it for slash commands. 😄 

D659C218-EC4F-4C8D-A19D-3B249BC92A68

danielbayley: Yeah I noticed that before. It’s used for both (like a good, useful icon should be)… pretty sure Iv’e seen it in GitHub Desktop and Atom also.

> not entirely sure if this shape truly has this universal meaning.

Isn’t that the whole point of `tags` though—to include other possible (relevant) meanings? By which logic:
Obviously `slash` is part of the file name already, but maybe we should also add `commands` as a tag then…?

PR #997, opened 2023-03-29
Add `iteration` icons

danielbayley:
Like https://primer.style/design/foundations/icons/iterations-16

### DIFF
--- a/icons/iteration-ccw.json
+++ b/icons/iteration-ccw.json
@@ -1,0 +1,11 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "arrow",
+    "right"
+  ],
+  "categories": [
+    "arrows",
+    "design"
+  ]
+}
--- a/icons/iteration-ccw.svg
+++ b/icons/iteration-ccw.svg
@@ -9,6 +9,6 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M20 11c0-4.4-3.6-8-8-8s-8 3.6-8 8 3.6 8 8 8h8" />
-  <polyline points="17 16 20 19 17 22" />
+  <path d="M20 10c0-4.4-3.6-8-8-8s-8 3.6-8 8 3.6 8 8 8h8" />
+  <polyline points="16 14 20 18 16 22" />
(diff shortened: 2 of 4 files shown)


ericfennis (commented):
@danielbayley Nice icon!

I think we should rename this to `iteration-ccw` and create another one with the opposite direction. Like the `refresh` icons.

karsa-mistmere:
Really nice icons, but I'm a bit on edge about their visual weight and centre of gravity.

Maybe they could be either moved 1px up or increased by 1 unit in radius, e.g.:
image

In Figma: https://www.figma.com/file/dO1uOdCGn0AHhY0RJyKHc8/lucide%2Fiterations?node-id=0%3A1&t=weYM3Odsf546Dc43-1

Not totally sure, lmkwyt. 🤔

danielbayley:
> Really nice icons, but I'm a bit on edge about their visual weight and centre of gravity.
> 
> Maybe they could be either moved 1px up or increased by 1 unit in radius, e.g. (icons 4 to 6 for comparison): image
> 
> In Figma: https://www.figma.com/file/dO1uOdCGn0AHhY0RJyKHc8/lucide%2Fiterations?node-id=0%3A1&t=weYM3Odsf546Dc43-1
> 
> Not totally sure, lmkwyt. 🤔

Well the options are actually more limited than it probably seems, due to the arrow and the 1px padding around the edge. I think the best solution is just to push it up by 1px, which I committed…

ericfennis:
@danielbayley I think the arrows are now to small with your last commit. If you compare it with other arrow icons from lucide on 100% you can see that the arrow of this icons are not well matching with the other designs. I think it's better to increase the whole icon a bit. I think the second icon from @karsa-mistmere looks more like it.

danielbayley:
@ericfennis @karsa-mistmere Pushed…

ericfennis:
@danielbayley The arrow looks still to small compared to the other arrow icons.

danielbayley:
> @danielbayley The arrow looks still to small compared to the other arrow icons.

@ericfennis `push`ed…

ericfennis (approved):
Nice work, this looks better!

PR #998, opened 2023-03-29
Improve `inbox` and add `inbox-*` icons (doubling as `git`[`pop`]`stash` icons)

danielbayley:


danielbayley:
Potentially more…

inboxes

karsa-mistmere:
Not quite sure about the flat design. Our current `inbox` icon surely is an odd one out (we don't normally use 3D-ish perspectives), but if we go for a rectangular design, it must not by any larger than `square` is, and I think a more traditional notch is probably more recognisable:
image

This, however, does not solve the issues above, I'm afraid. :)
Sidenote-ish: what exactly would be the use cases for these extra icons?

danielbayley:
> what exactly would be the use cases for these extra icons?

Mostly for `git stash` and `git pop stash`, like for example in _[GitKraken](https://gitkraken.com)_ (below), but I thought they should probably be a set, related to todos/mail/notifications/whatever the use case is for the existing `inbox` icon…

Screenshot 2023-03-30 at 5 54 44 pm


> I think a more traditional notch is probably more recognisable

I though more rounded fits better with the rest of the set being very round in general…

inbox-stash-3d

inbox-stash-tile

karsa-mistmere:
It's mostly the size that's the issue: square icons should not be larger than `square` itself to keep things looking consistent.

I'd be fine with the circular notch, it's just that it doesn't fit that well with the down pointing arrow. :)

danielbayley:
> It's mostly the size that's the issue: square icons should not be larger than `square` itself to keep things looking consistent.

Artboard 1@10x

karsa-mistmere:
Not bad, although I'm still not absolutely sold about the flat inbox design, as said, it's maybe a better fit for Lucide in particular, but much less obvious than the very standard inbox shape we currently have.

How about a very slight redesign of the current inbox icon and using the same arrow shape as in `file/folder-in` and `file/folder-out`?
image

danielbayley:
> Added or changed icons
> 
>    

Cool! Super useful bot. Nice work @jguddas!

ericfennis (dismissed):
LGTM!

icons/inbox-full.svg
```diff
@@ -0,0 +1,16 @@
...
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <line x1="8" y1="13" x2="3" y2="13" />
+  <line x1="16" y1="13" x2="21" y2="13" />
+  <path d="M16,13c0,2.2-1.8,4-4,4s-4-1.8-4-4" />
+  <path d="M7.2,4C6.5,4,5.8,4.4,5.4,5.1L3,13v6c0,1.1,0.9,2,2,2h14c1.1,0,2-0.9,2-2v-6l-2.5-7.9c-0.3-0.7-1-1.1-1.8-1.1 H7.2z" />
```
karsa-mistmere, line 15: Three of these paths could be merged (both here and in every other icon):
```suggestion
  <path d="M3 13h5c0 2.2 1.8 4 4 4s4-1.8 4-4h5" />
  <path d="M7.2 4c-.7 0-1.4.4-1.8 1.1L3 13v6c0 1.1.9 2 2 2h14a2 2 0 0 0 2-2v-6l-2.5-7.9c-.3-.7-1-1.1-1.8-1.1H7.2Z" />
```

icons/inbox-in.svg
karsa-mistmere, line 1: Still not really a fan of how the arrow's head and the corners of the notch meet, how about maybe introducing a tiny bit of rounding to the corners to somewhat mitigate that sharpness?

image

(√2/2 px rounding in this instance)

karsa-mistmere (changes requested):

icons/inbox-empty.svg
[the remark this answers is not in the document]
danielbayley: Ah yeah, good catch. They’re simply the wrong way around. Too many tabs open… 🙈

karsa-mistmere (changes requested):
I feel bad, for not approving this already, but there are still a few issues. 🙈 
1) There's a sharp corner in the top left corner that I believe was unintentional.
2) All of them are slightly off center to the bottom, I'd prefer if they were perfectly centered like our current inbox icon is.

danielbayley:
> I feel bad, for not approving this already, but there are still a few issues. 🙈
> 
> 1. There's a sharp corner in the top left corner that I believe was unintentional.
> 2. All of them are slightly off center to the bottom, I'd prefer if they were perfectly centered like our current inbox icon is.

Not sure what happened… Illustrator seems to introduce problems at times.

karsa-mistmere (changes requested):
Hm... these appear the same to me, maybe I'm miscommunicating.

What I mean is that this new inbox is not vertically centered:
image

The sharp corner is also still there:
image

danielbayley:
> The sharp corner is also still there:

Fixed. Not sure what happened there…

> this new inbox is not vertically centered:

Well it’s 18x17px (to fit nicely around the inside elements), so there’s a decision to make… Unless you’re suggesting to move it up 0.5px? (surely not)

1. Make it 18x18px by extending the top area by 1px
2. Make it 18x18px by extending the bottom area by 1px
3. Leave it 18x17px and either leave it as is, or move the whole thing up by 1px… still not vertically centered

Here’s what those look like:

Untitled

I think:

1. Messes up the inside elements, as they don’t fit as nicely
2. Makes the base look too chunky

What do you think?

ericfennis:
@danielbayley Let's go for a squared icon. so 18x18 (top or bottom-right). Half pixels is not an option that will look blurry on low-resolution screens.

karsa-mistmere:
@ericfennis: We could also maybe go for one that's 20x18?

image
inbox

ericfennis:
@karsa-mistmere oh yeah that's also possible!

danielbayley:
20x18 version pushed… @karsa-mistmere @ericfennis See what you think?

jguddas:
You can use a single arc with a `4px` radius instead of the two curves `M15.7 13.5a4 4 0 0 1-7.4 0`

danielbayley:
> You can use a single arc with a `4px` radius instead of the two curves `M15.7 13.5a4 4 0 0 1-7.4 0`

These links are broken… Can you post the full `path`?

karsa-mistmere:
@jguddas: I'm afraid that path you posted is not quite right, it makes the path not pixel perfect anymore:
image

danielbayley:
> > @jguddas: I'm afraid that path you posted is not quite right, it makes the path not pixel perfect anymore:
> Oh no, you are right 😞

Can that not be done then? Leaving this one ready to go…?

karsa-mistmere (changes requested):
`inbox-in` and `inbox-out` have different arrow sizes, let's make them the same size (either 6x6 or 6x8 both).

danielbayley:
> `inbox-in` and `inbox-out` have different arrow sizes, let's make them the same size (either 6x6 or 6x8 both).

So here are possible options, as far as I can tell… Unfortunately, they all have trade-offs: 😞

Untitled

Untitled

Row:
1. (existing): Arrows fit nice and snug, but are _slightly_ different lengths
2. (6x8): Arrows equal length, but spoils `inbox-out` tray base shape
3. (6x7): Compromise, but `inbox-in` doesn’t line up nicely
4. (6x6): Arrows equal length, but `inbox-in` doesn’t line up nicely at the top, plus `inbox-out` arrow doesn’t fit nicely at the bottom now either
5. (6x8): Arrows equal length, but I think a slight violation with `inbox-out`

@karsa-mistmere @ericfennis @jguddas What do we think?

Personally I went for option (row) 1…

ericfennis:
I agree with @karsa-mistmere. I think it's more important the the arrows are on smart place. And equal then other arrow icons.

danielbayley:
Hmm, ok well if we are going with 6x6px arrows, that actually opens up a bunch of other possibilities… 🙈

Like matching the `square` icon base shape better…

Untitled

Untitled

Untitled

Untitled

@karsa-mistmere @ericfennis Thoughts on these?

ericfennis:
@danielbayley Nice designs, I really like the first set, just simple and nice to look at.

karsa-mistmere:
@danielbayley: Hard to compare these without 100% views. Can you provide one with one row for each version of icons?

danielbayley:
> @danielbayley: Hard to compare these without 100% views. Can you provide one with one row for each version of icons?

Sure:

Untitled

Rows 1 & 2 are 18x16px, slightly different variations. Row 3 is 20x16px, row 4 is 20x18px, and row 5 is 18x18px…

Untitled

karsa-mistmere:
Thanks, but this isn't at 100%. 😅
By a 100% view we mean one where 1px is literally 1px, so icons are rendered at 24by24px.

danielbayley:
> By a 100% view we mean one where 1px is literally 1px, so icons are rendered at 24by24px.

@karsa-mistmere I was just taking screenshots at 100% (‘actual size’) in Illustrator, but it’s actually double size… Anyway, see updated https://github.com/lucide-icons/lucide/pull/998#issuecomment-1534652733 …

karsa-mistmere:
> > By a 100% view we mean one where 1px is literally 1px, so icons are rendered at 24by24px.
> 
> @karsa-mistmere I was just taking screenshots at 100% (‘actual size’) in Illustrator for convenience, but it’s actually double size in practice… Anyway, see updated [#998 (comment)](https://github.com/lucide-icons/lucide/pull/998#issuecomment-1534652733)

That's very definitely still not at 100%, I'm afraid. 😕 
Instead of making a screenshot of your Illustrator page, you're going to have to export it (that will also get rid of artboard outlines).

> Hopefully #1180 being `merge`d will reduce the need for screenshots anyway…

Well, in general surely, in this scenario, not really, since you normally wouldn't have 4 or 5 versions for each icon in a PR.

jguddas:
Before the refine it looked perfect, now it feels too small IMO.

danielbayley:
> That's very definitely still not at 100%, I'm afraid.

It definitely is showing as 24x24px for me (@2x, so 48x48, as I’m on a ‘retina’ device… It renders for me the same as the [preview](https://github.com/lucide-icons/lucide/pull/998#issuecomment-1494857232) bot now)… But here is the direct export anyway…

Artboard 1

danielbayley:
> Before the refine it looked perfect, now it feels too small IMO.

Think iv’e laid out all the possible options here https://github.com/lucide-icons/lucide/pull/998#issuecomment-1534652733 and here https://github.com/lucide-icons/lucide/pull/998#issuecomment-1537216126 … Which row do you think? @jguddas

jguddas:
> Which row do you think?

3

karsa-mistmere:
Thanks, this works. :)

Looking at the 100% (or 1x) versions, I still feel like the arrows are a bit weak and also that maybe the `inbox-in`/`inbox-out` use cases aren't as apparent as they should be, since the arrows don't actually point from and to the outside of the inbox.

How about something like this?

image
Artboard 1
(other iterations for comparison)

danielbayley:
> I still feel like the arrows are a bit weak
> the arrows don't actually point from and to the outside of the inbox.

Probably because they were shrunk to 6x6px… :)

> How about something like this?

Not sure about the version without any sides… it looks too much like a phone (`phone-incoming`/`outgoing`…) and also feels out of place with the others.

I think putting aside the arbitrary restriction of the 6x6px arrow frees up some potentially better options…

Artboard 1

Artboard 1@2x

Artboard 1@4x

danielbayley:
> This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 5 days.

😢

jguddas:
> > This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 5 days.
> 
> cry

@karsa-mistmere deserves the break. @ericfennis what about working on the [bus-factor](https://en.wikipedia.org/wiki/Bus_factor) for the reviews?

danielbayley:
> @karsa-mistmere deserves the break. @ericfennis …

No doubt! Maybe the `stale` bot is a bit too enthusiastic…?

ericfennis:
@danielbayley @jguddas Sorry guys. I got so much work lately. Full-time job and lot things to do beside that. I hope next week I will get some more time to complete some PRs.

So I'm not hit by a bus 😅.

danielbayley:
> Sorry guys…

@ericfennis No worries! Tell me about it… Part of my motivation to get all these drafts over the line is for other projects… Trying to focus more on one thing at a time, rather than always being spread too thin.

> So I'm not hit by a bus 😅.

Ha remember to look both ways! 🚌👀

karsa-mistmere:
To be honest, I wanted to give this one some rest, as I was and still am less than enthusiastic about these changes, and I still feel like something like this would just work better for `in` and `out`:

inboxes

(Note that these are the same arrows as in `file-in` and `folder-in`)

karsa-mistmere:
We should also take into consideration that `inbox` currently uses the exact same shape as `hard-drive`.

danielbayley:
> We should also take into consideration that `inbox` currently uses the exact same shape as `hard-drive`.

Agreed the outer shape should match `hard-drive`. But also for consistency, the bevel should match [`wallet-cards`](https://lucide.dev/icons/wallet-cards)—as was the original intent… The more jagged edges don’t really fit the curvature of all the other icons either.

ericfennis:
The icons don't look very consistent with each other.

I think we could improve this by using the same place as the dot for all the `plus` and `min` icons. And for the `inbox-in` and `inbox-out` use the same arrow as the `phone-forwarded`.

image

Issue #670, opened 2022-05-15
We're not accepting new Brand icons


mittalyashu:
Thank you for considering to add a new brand/company icon to Lucide. We're no longer accepting new brand icons.

**To know more:** https://github.com/feathericons/feather/issues/763
**We recommend using:** https://github.com/simple-icons/simple-icons

mittalyashu:
@ericfennis Can you please pin this issue?

locness3:
Why is it unpinned

danielbayley:
@mittalyashu @ericfennis @karsa-mistmere @jguddas Would adding a Markdown icon violate this? (even if an abstraction)

ericfennis:
@danielbayley markdown icon we should accept I think. Commonly used to indicate markdown fields for example.

locness3:
still a brand

mittalyashu:
Kind of similar, if we start accepting library and framework logo's (React, Vue.js)

danielbayley:
> React for one is objectively too detailed for our guidelines.

@mittalyashu I added https://lucide.dev/icons/atom, which you could use as [P](https://preactjs.com)/React…

HabeebOpe:
I don't know your reason for not adding brand icons, but icon platforms should have brands cause it's been use by a lot of people everyday

vexkiddy:
looks like this is a sensitive topic, but platform icons are used all over the place for log in buttons. eg you have github, which i could use for a log in with github button, but no apple or google. would it be possible to just add platform icons only ? so just windows, google, apple ?

karsa-mistmere:
@vexkiddy: All of these are protected by copyright, so no, unfortunately we cannot officially add them, since they would need to be altered to match our guidelines.

busybox11:
The GitHub and GitLab icons are marked as deprecated, and will eventually be removed for legal reasons IIRC. Nothing we can do here.

khanakia:
How to use simple icons or convert them to lucide icon ?

karsa-mistmere:
> How to use simple icons

Download SVG, use as SVG. Or something similar. Depends on framework really.

> or convert them to lucide icon ?

You don't. They have roughly the same dimension, but it's a completely different icon set.

VIKTORVAV99:
How about creating a separate package with just brand icons. They don't have to follow the same design guidelines as all the other Lucide icons but share the same API and have the same tree shaking abilities.

[...]

PR #1001, opened 2023-03-30
Add `spiral` and `lucide` icon

danielbayley:
spiral

lucide

### DIFF
--- a/icons/lucide.json
+++ b/icons/lucide.json
@@ -1,0 +1,13 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "swirl",
+    "twirl",
+    "spin",
+    "pattern",
+    "lucide"
+  ],
+  "categories": [
+    "shapes"
+  ]
+}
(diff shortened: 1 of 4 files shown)


karsa-mistmere:
Brand icons, even our own 🙂, are deprecated: #670 
Any use cases for the spiral?

danielbayley:
> Any use cases for the spiral?

Well it started as an alternate for the existing `tornado`…

image

But then, I thought it looked better as `spiral`, then naturally also made it into the _Lucide_ icon…

> Brand icons, even our own 🙂, are deprecated: #670

…which I didn’t realise would even fall under #670.

karsa-mistmere:
I think from a weather report use case people rarely expect to look at a tornado from above. 😄

danielbayley:
> I think from a weather report use case people rarely expect to look at a tornado from above. 😄

At that point, it would be too late for the report 😅

I have a demo weather app as a `create-react-app` base template, and was thinking of doing that particular icon this way…

karsa-mistmere (changes requested):
@danielbayley: The `lucide` icon has to go and the spiral will need some proposed use cases before this can move forward.

PR #1002, opened 2023-03-30
Improve metadata for keys, and modify `arrow-big-*` icons to include `shift` key

danielbayley:
<kbd>⇧</kbd>

See also #1013.

### DIFF
--- a/icons/arrow-big-down-dash.json
+++ b/icons/arrow-big-down-dash.json
@@ -1,0 +1,11 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "key",
+    "download"
+  ],
+  "categories": [
+    "arrows",
+    "files"
+  ]
+}
--- a/icons/arrow-big-down-dash.svg
+++ b/icons/arrow-big-down-dash.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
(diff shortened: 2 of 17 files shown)


karsa-mistmere:
Too similar to: https://lucide.dev/icon/arrow-big-up?search=arrow-big-up

I'd rather just add the `shift` tag to this existing icon.

ericfennis:
@danielbayley we could try that. Looking at your design of the "big arrow" it seems a bit more balanced than the current big arrows. Maybe we could try a side-by-side comparison and decide?

ericfennis (approved):
Awesome work @danielbayley !

ericfennis (changes requested):
@danielbayley Sorry one last thing, I think the name for `arrow-big-up-2` is not right because it is not another variant of it. Rename it to: `arrow-big-up-dash` or something, and why don't we add this one also for other directions (if we do, it is maybe better to create a separate PR if we do this)?

danielbayley:
>  and why don't we add this one also for other directions (if we do, it is maybe better to create a separate PR

We could, but it’s just meant to be <kbd>caps lock</kbd>… Not sure what use case for the others would be?

ericfennis:
@danielbayley Not sure as well but best to keep the generic, so adding them is possible

ericfennis (approved):

karsa-mistmere:
Hm... I could've sworn that's what I'd originally seen in this PR, but what do you think of something more like this for `arrow-big-up-dash`?

image

danielbayley:
> what do you think of something more like this

@karsa-mistmere Implemented…

PR #1004, opened 2023-03-31
Add `sort-alphabetic-`[`asc`/`desc`]ending icon

danielbayley:
Artboard 1@30x

### DIFF
--- a/icons/sort-alphabetic-asc.json
+++ b/icons/sort-alphabetic-asc.json
@@ -1,0 +1,10 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "filter"
+  ],
+  "categories": [
+    "text",
+    "layout"
+  ]
+}
--- a/icons/sort-alphabetic-asc.svg
+++ b/icons/sort-alphabetic-asc.svg
@@ -1,0 +1,17 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <line x1="19" y1="8" x2="13" y2="8" />
+  <path d="M13,10V7c0-1.7,1.3-3,3-3h0c1.7,0,3,1.3,3,3v3" />
+  <polyline points="13,14 19,14 13,20 19,20" />
+  <path d="M9,7L6,4L3,7" />
+  <path d="M6,4v16" />
+</svg>
--- a/icons/sort-alphabetic-desc.json
+++ b/icons/sort-alphabetic-desc.json
@@ -1,0 +1,10 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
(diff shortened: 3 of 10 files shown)


karsa-mistmere:
The letter `A` does not have sufficient inner spacing, I'm afraid. :(

I actually have these in my icon drafts, but since I was unhappy with the shape, I never submitted it:
image

karsa-mistmere:
Not fantastic, but this spacing is maybe very-very barely sufficient. :/ 
image

danielbayley:
Shouldn’t the arrow be on the left, same as https://lucide.dev/?search=sort?

danielbayley:
Shouldn’t the arrow be on the left, same as https://lucide.dev/?search=sort?

@karsa-mistmere See above…

karsa-mistmere:
These are pretty old designs, made for Feather before Lucide even came to existence, but I don't think that's really the point. :)

What's noteworthy though is that the arrows are of a different size than in `sort-asc` and `sort-desc`. They could be swapped around but the smaller arrow makes them not very well balanced. :/ 

image

ericfennis:
I agree with @ karsa-mistmere, the use case for lowercase is not very solid. Most of the time you sort by alphabet or numbers.

danielbayley:
> I agree with @ karsa-mistmere, the use case for lowercase is not very solid. Most of the time you sort by alphabet or numbers.

I think there is a slight misunderstanding here. I wasn’t proposing lowercase sort as a separate use case to say, uppercase sort (although they can technically be slightly different in terms of functionality…) I was simply playing with lowercase a-z as a way to visually line things up nicely as a set of `sort-*` icons, a-z meant to represent only alphabetic sort _in general_, if that makes sense…?

karsa-mistmere:
I think we got you, we're just saying that using `a..z` in this use case is (maybe too) unusual (for it to actually work) — I haven't been able to find a single sort icon that uses lowercase letters.

danielbayley:
> I think we got you

Ok cool, not sure what difference it really makes (aside from I suppose _possibly_ implying some kind of separate, case-sensitive sort) but see `push`ed an alternative approach anyway…

karsa-mistmere:
Hey @danielbayley,

I've just opened a new PR (#1028), since I have a somewhat unrelated bone to pick with these icons and didn't want to make you do what I believe is the necessary work to sort things out, since it's a _lot_ of refactoring. 😅 

I hope I didn't step on your toes, it was all done in good faith. 🙈

danielbayley:
> Hey @danielbayley,
> 
> I've just opened a new PR (#1028), since I have a somewhat unrelated bone to pick with these icons and didn't want to make you do what I believe is the necessary work to sort things out, since it's a _lot_ of refactoring. 😅
> 
> I hope I didn't step on your toes, it was all done in good faith. 🙈

Fine with me… I have enough to do! 😅 I see you stole some of these… So I should go ahead and just close this PR?

karsa-mistmere:
Nah, not yet, I still want to see what @ericfennis thinks of that other one.
And obviously, should that one get merged, you'll be duly credited. :)

ericfennis:
Closed by #1028

PR #1011, opened 2023-04-03
Add `git` `repository` icons

danielbayley:


karsa-mistmere:
The bookmark still violates the guidelines I'm afraid, maybe something like this could work:
```svg
  <path d="M16 16v6l-2-1.5-2 1.5v-6" />
```
image


Not sure about the names in general and the need for the bookmark though, I'd prefer keeping `book` as it is, adding the `repository` tag, removing the bookmarks and naming everything along the lines of `book-plus`, `book-minus`, `book-key`, `book-arrow-down` etc.

icons/book-marked.svg
```diff
@@ -0,0 +1,15 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M4,19V5c0-1.7,1.3-3,3-3h13v14" />
+  <path d="M8,22H7c-1.7,0-3-1.3-3-3l0,0c0-1.7,1.3-3,3-3h13v6" />
+  <polyline points="16,16 16,22 14,20 12,22 12,16" />
```
karsa-mistmere, line 14: I think this tiny bookmark is just far too dense at the bottom and the alteration of the current `book` icon seems more like a necessary evil to accommodate for this rather than a definite improvement. 😐

I still propose not to include the bookmark on the other icons at all, as for this icon to maybe include a bookmark more similar in shape to our existing [`bookmark`](https://lucide.dev/icon/bookmark?search=bookmark) on the book itself, such as this:
```suggestion
  <path d="M4 19.5A2.5 2.5 0 0 1 6.5 17H20" />
  <path d="M6.5 2H20v20H6.5A2.5 2.5 0 0 1 4 19.5v-15A2.5 2.5 0 0 1 6.5 2Z" />
  <path d="M11 2v7l2.5-1.56L16 9V2" />
```

karsa-mistmere (changes requested):

karsa-mistmere:
`book-copy` doesn't have the necessary 1px padding, other than that LGTM.

danielbayley:
> `book-copy` doesn't have the necessary 1px padding, other than that LGTM.

It was misaligned. Fixed now…

karsa-mistmere (approved):
Great work, @danielbayley, I think this is good to go if @ericfennis also approves. :)

PR #1017, opened 2023-04-05
Refine `heart-*` icons

danielbayley:
Related to #1018.

PR #1018, opened 2023-04-05
Add missing playing cards suit icons

### LINKED ISSUE #144: playing card suit symbols
I put together a solitaire game for android. First I tried using the unicode symbols, but was surprised to learn that they commonly receive a fixed coloring when rendered as text (i.e. spades and clubs are black and diamonds and hearts rendered red, despite explicit application text styles). This was true in my case, due to whatever rendering engine treats Flutter `Text` elements.

So I looked for some SVG's for it, and found some that worked well enough.  I find after playing the game a bit though, at the scale of my phone's display, the spades icon (♠) and clubs icon (♣) are not differentiated well enough, causing some eye strain.

I did not see any great alternatives in other icon sets. I can try to submit a PR if it's a welcome addition here.

## Icon Request

* Icon name: playing card suits (hearts (filled), spades, clubs, diamonds)
* Use case: implementing or creating an icon for a card game

in the app, without enough differentiability:

these clubs are visually different from the spades, the lobes (?) of the clubs are nearly completed circles:
220px-4coloraces1

standard cards showing similar style, good difference between clubs and spades
bicycle_cards

this is an example of not-so-great difference spades/clubs, unfortunately prevalent among free icon svgs

here's a vintage style clubs which seems to completely differentiate from spades, but it's less iconic, my preference would be similar to the standard cards above
vintage-clubs

danielbayley:
Closes #144. Related to #1017.


karsa-mistmere:
Any use cases for the suites themselves and more importantly the cards?

danielbayley:
> Any use cases for the suites themselves and more importantly the cards?

#144

karsa-mistmere:
@danielbayley: Yes, this explains the suits themselves, but not the cards depicting the suits as I was very specifically asking.

(I'm not even entirely sure we need the suits, but since they're Unicode, I'm OK with adding those.)

danielbayley:
I see, well it seems wrong to already have `heart` and `diamond` without the other 2. But sure, I can remove the cards…

karsa-mistmere:
Again, I'm fine with potentially adding the cards but for every new icon request and submission we need use cases.

As for the new suits, I'm a bit on edge, on one hand the stems seem a tad too short, but I don't really like my "fix" either:
image

danielbayley:
> on one hand the stems seem a tad too short, but I don't really like my "fix" either

Yeah I tried that too… Doesn’t really work unless we’re ok with the `heart`s being different sizes…

ericfennis (approved):
Hmm, I tried some things as well but doesn't make it better.
So I think this design is fine, it will work.

Issue #902, opened 2022-12-19
Justification for using path vs line/rectangle elements in SVGs


aschaap:
Is there a reason a lot of icons have rectangles and lines? Is SVGO's conversion to paths not trusted, or are semantics important?

ericfennis:
@aschaap Yes there are some reasons but also improvements can be made.
In some cases, rounded rectangles have less code than paths.
### Example
```svg
<svg
  xmlns="http://www.w3.org/2000/svg"
  width="24"
  height="24"
  viewBox="0 0 24 24"
  fill="none"
  stroke="currentColor"
  stroke-width="2"
  stroke-linecap="round"
  stroke-linejoin="round"
>
  <rect x="3" y="3" width="18" height="18" rx="2" ry="2" />
</svg>
```
vs
```svg
<svg
  xmlns="http://www.w3.org/2000/svg"
  width="24"
  height="24"
  viewBox="0 0 24 24"
  fill="none"
  stroke="currentColor"
  stroke-width="2"
  stroke-linecap="round"
  stroke-linejoin="round"
>
  <path d="M19 3H5a2 2 0 0 0-2 2v14c0 1.1.9 2 2 2h14a2 2 0 0 0 2-2V5a2 2 0 0 0-2-2Z" />
</svg>
```

Beside that when lucide forked from feathericons we kept a lot of SVG code the same.
I think we should convert some icons to paths, because it is less code.
For example menu icon:

```svg

```

```svg
<svg
  xmlns="http://www.w3.org/2000/svg"
  width="24"
  height="24"
  viewBox="0 0 24 24"
  fill="none"
  stroke="currentColor"
  stroke-width="2"
  stroke-linecap="round"
  stroke-linejoin="round"
>
  <line x1="4" y1="12" x2="20" y2="12" />
  <line x1="4" y1="6" x2="20" y2="6" />
  <line x1="4" y1="18" x2="20" y2="18" />
</svg>

```

aschaap:
Awesome, good to know that smaller file size is the goal.

PR #936, opened 2023-02-04
add rat icon

henri42:
Here is a little rat icon.
I hope it respects the guidelines correctly !

rat

PR #1020, opened 2023-04-06
Hand-optimizations for size

aschaap:
Some optimization types are separated into their own commit, in case these are found to be objectionable.

No new icons, just size reductions where the final result is effectively or exactly identical.

Relevant issue: #902

### DIFF
--- a/docs/icon-design-guide.md
+++ b/docs/icon-design-guide.md
@@ -55,6 +55,8 @@ Here are rules that should be followed to keep quality and consistency when maki
 
 Before an icon is added to the library, we like to have readable and optimized svg code.
 
+Never use [`<use>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use). While it may sometimes seem like a good way to optimize file size, there's no way to ensure that the referenced element IDs will be unique once the SVGs are embedded in HTML documents.
+
 ### Global Attributes
 
 For each icon these attributes are applied, corresponding to the above rules.


aschaap:
I see <use /> is not rendered properly by the workflow.

karsa-mistmere:
@aschaap: I've found multiple instances where separate paths have been merged into a single one, this is something we've deliberately chosen not to do as such compound paths sometimes render incorrectly in Figma amongst others.

(E.g. the two separate lines of arrows have been merged).

karsa-mistmere:
> I see <use /> is not rendered properly by the GitHub Action.

Not sure we even have support for `<use>` in general, it might also break font outlining, we'll have to check.

aschaap:
> @aschaap: I've found multiple instances where separate paths have been merged into a single one, this is something we've deliberately chosen not to do as such compound paths sometimes render incorrectly in Figma amongst others.
> 
> (E.g. the two separate lines of arrows have been merged).

Interesting, I found you in #936 mentioning Figma also does not render zero-length paths with caps as dots.

Is this decision to avoid compound paths (and any others like it) documented anywhere beyond the Readme/website? I.e., are there other Figma limitations to be aware of, and are there other similarly limited use cases to support?

karsa-mistmere:
> Is this decision to avoid compound paths (and any others like it) documented anywhere beyond the Readme/website?

It isn't, but I guess we should probably add this (and some others) to the [icon design guide](https://lucide.dev/docs/icon-design-guide), thanks for bringing this to my attention. In fact the guidelines are in dire need of some rewrite anyway: #650

aschaap:
Thank you for the quick response. I found a generic 'improve SVG support' https://forum.figma.com/t/improve-svg-support/2505 dating back to April 2021. Do you know whether there is an issue for the Figma SVG path problems? 

If Figma's SVG support is an outlier, would it make sense to adapt the package exporter to separate the combined paths instead?

jguddas:
I'm working on automating some of those optimizations.
One question, which elements are preferred over `path`?

My opinion:

- `rect`, `circle`, `ellipse` should be used instead of `path`.
- `path` should be used instead of `line`, `polygon`, `polyline`.

You can already do things like decompounding and converting `line`, `polygon`, `polyline` to `path` using https://lucide-site-git-feat-added-edit-page-jguddas.vercel.app/edit.

But I haven't built the `path` to `rect`/`circle`/`ellipse` feature, and I wanted to ask your opinion first.

aschaap:
@jguddas : I agree with your opinion, but would add nuance to the preference of `rect` over `path`: `rect` with `rx` is shorter than the equivalent `path` with arcs, but without `rx` a `path` is often shorter.

As for `circle`, I'm not sure one can create a truly equivalent `path`; converting `path' to `circle' would require some loss of precision/approximate matching.

ericfennis:
@aschaap Can you rebase, your PR. I opmized the rest of the icons, so this PR should be a bit cleaner now.

aschaap:
@ericfennis Your optimization PR #1025 seems to only rearrange attributes? If you used a tool for this, would it make sense to solve all the merge conflicts it creates? It would seem easier to reapply them after this PR, unless I'm missing something.

ericfennis:
@aschaap I thought you were optimizing the icons by accident. So I thought to do it I a separate PR we could see you changes better. Are your changes now gone after rebasing?

aschaap:
I optimized icons with the intention of doing so (along with learning more of the SVG Path spec). Rebasing does not eliminate my changes, and is taking time due to having to merge it with rearranged attributes.

PR #1023, opened 2023-04-07
Add `find-replace` icons

danielbayley:
As mentioned in another PR, I’m putting together a VS Code extension for Lucide product and file icons…

This PR is meant to cover the find/replace panel:

Screenshot 2023-04-07 at 9 51 38 pm

See also: _[codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html)_

### DIFF
--- a/icons/case-lower.json
+++ b/icons/case-lower.json
@@ -1,0 +1,13 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "text",
+    "letters",
+    "characters",
+    "font",
+    "typography"
+  ],
+  "categories": [
+    "text"
+  ]
+}
--- a/icons/case-lower.svg
+++ b/icons/case-lower.svg
@@ -1,0 +1,16 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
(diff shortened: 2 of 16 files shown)


karsa-mistmere:
Hm... I guess something like `a=>b` would be more customary, but there's probably no real way we're going to make that work in Lucide. 😄
This is my best attempt so far and it leaves a lot to be desired:
image

A few issues/ideas though:

1. shouldn't these be simply `replace` and `replace-all`?
   (In practice the "find" part will have been over and done with).

3. They're 1px off center vertically, I think adjustment like this could work:
image
(1px extra vertical gap with arrows being 1px longer).

4. Overall, I like the abstraction, but I think the second arrow might be misleading, we're replacing one thing for another, not switching them around, so maybe something like these would better represent replace and replace all?
image

danielbayley:
1. Good point. I was just going (at this initial draft stage) by what names are on tooltips in VS Code (which I think match _codicons_), but I’m sure some of these can be improved on, or named more appropriately for Lucide…

2. Yeah that works

3. I think I have it:

Untitled

ericfennis (approved):

karsa-mistmere:
Nice ones, LGTM!

PR #1027, opened 2023-04-08
Add `copy-*` icons

danielbayley:
or `close-all`…


icons/x-square-all.svg
```diff
@@ -0,0 +1,16 @@
...
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect width="13" height="13" x="9" y="9" rx="2" ry="2" />
+  <path d="M5 15H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9a2 2 0 0 1 2 2v1" />
+  <line x1="18" x2="13" y1="13" y2="18" />
+  <line x1="13" x2="18" y1="13" y2="18" />
```
jguddas, line 15:
```suggestion
  <rect x="9" y="9" width="13" height="13" rx="2" ry="2"/>
  <path d="m13 13 5 5"/>
  <path d="m18 13-5 5"/>
  <path d="M5 15H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9a2 2 0 0 1 2 2v1"/>
```

ericfennis (approved):

karsa-mistmere:
Hm... since this is exactly the same icon as `copy`, just with an added `x`, maybe it should be named `copy-x` instead?

And while we're there, maybe `copy-plus`, `copy-minus` and `copy-check` could also be useful? (not totally sure, but maybe for some clipboard magic or diff actions).
(for the latter, the base icon would have to be modified to be of an even numbered size see below)
image

icons/copy.svg
```diff
@@ -9,6 +9,6 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect x="9" y="9" width="13" height="13" rx="2" ry="2" />
-  <path d="M5 15H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9a2 2 0 0 1 2 2v1" />
+  <path d="M4 16c-1.1 0-2-.9-2-2V4c0-1.1.9-2 2-2h10c1.1 0 2 .9 2 2" />
```
ericfennis, line 12: I'm not sure with this change. there is more code added.
I think we can better use the old rect and path.

icons/copy-equal.svg
```diff
@@ -0,0 +1,16 @@
+<svg
```
ericfennis, line 1: What is the use case exactly of this icon?

icons/copy.svg
danielbayley: Yeah, Illustrator seems reluctant to output `rect`s. I’ll just code these manually instead…

icons/copy-equal.svg
danielbayley: I was just aiming for all `circle-*`, `square-*` and `copy-*` (essentially just multiple squares) icons to be more of a set with the same set of elements…

ericfennis: Ah check, I get your thinking.
But I'm not sure about the use case. Let's keep this one out if we couldn't come up with a use case.

karsa-mistmere:
@danielbayley There's a conflict from concurrent optimization of icon files, can you resolve it, then I think this is also good to go, great work. :)

ericfennis (approved):
Awesome work!

PR #1030, opened 2023-04-10
Add `messages-square` icon

danielbayley:


karsa-mistmere:
Nice!

I actually have a few drafts for this which are pretty similar (heck, the smaller speech bubble is literally identical), but I actually prefer yours, as it's more consistent with the existing `message-square` icon:
image

karsa-mistmere (approved):

icons/messages-square.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "comment",
+    "chat",
+    "conversation"
```
karsa-mistmere, line 6:
```suggestion
    "conversation",
    "copy",
    "multiple"
```

danielbayley:
> Nice!
> 
> I actually have a few drafts for this which are pretty similar (heck, the smaller speech bubble is almost identical), but I actually prefer yours, as it's more consistent with the existing `message-square` icon: image

I do like the last 2 though… Maybe add one as an alternate?

ericfennis:
I prefer the last one. The best use case for this one is I think a "conversation" which is more two message balloons on each site talking to each other. The current proposal is I think a bit off

icons/messages-square.svg
```diff
@@ -0,0 +1,14 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M14 9c0 1.1-.9 2-2 2H6l-4 4V5c0-1.1.9-2 2-2h8c1.1 0 2 .9 2 2v4z" />
+  <path d="M18 9h2c1.1 0 2 .9 2 2v10l-4-4h-6c-1.1 0-2-.9-2-2" />
```
ericfennis, line 13: I think we could use a bit more of the canvas, the icons will look better on 100% scale. The message balloon looks a bit squished now.

Suggestion improvement (right): 
image


```suggestion
  <path d="M14 9a2 2 0 0 1-2 2H6l-4 4V4c0-1.1.9-2 2-2h8a2 2 0 0 1 2 2v5Z" />
  <path d="M18 9h2a2 2 0 0 1 2 2v11l-4-4h-6a2 2 0 0 1-2-2v-1" />
```

ericfennis (changes requested):

ericfennis (approved):
Nice work!! LGTM

PR #1031, opened 2023-04-10
Add `filter-full` icon

danielbayley:
Implements `filter-filled` from [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html)…

### DIFF
--- a/icons/filter-full.json
+++ b/icons/filter-full.json
@@ -1,0 +1,10 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "funnel",
+    "hopper"
+  ],
+  "categories": [
+    "layout"
+  ]
+}
--- a/icons/filter-full.svg
+++ b/icons/filter-full.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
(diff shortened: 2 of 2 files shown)


karsa-mistmere:
Hm... any use cases for this one? I've consulted the tags but couldn't find anything.

danielbayley:
> Hm... any use cases for this one? I've consulted the tags but couldn't find anything.

For example in VS Code (I think with `git` lens extension)—but I’ve seen in other places—the filter icon is either filled or outlined depending on whether active filters are applied…

Screenshot 2023-04-11 at 8 40 36 am

Since Lucide doesn’t do fill, there is no other way to represent a difference.

karsa-mistmere:
Well, there's also that indicator dot. 😛

Dunno, just because we don't support filled icons out of the box, it doesn't mean people can't force fill them if they want to, anyhow, I'll let @ericfennis decide, personally, I'm not really a fan of this one. 😅

ericfennis:
I agree with @karsa-mistmere. Looking at other icon libraries nobody has an icon like this one. Probably nobody will use it.
I close this one, If other people think differently about this, they can react to this PR. Maybe we will reconsider and reopen it.

PR #1032, opened 2023-04-10
Complete `circle-*` set

danielbayley:
`circle-equal` can double as `debug-breakpoint-*` from [codicons](https://code.visualstudio.com/api/references/icons-in-labels#icon-listing)


icons/circle-equal.svg
```diff
@@ -0,0 +1,15 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <line x1="8" x2="16" y1="9" y2="9" />
+  <line x1="8" x2="16" y1="15" y2="15" />
```
karsa-mistmere, line 13: I think this has to be about the same ratio as our regular `equals` icon, so maybe something more like this?
```suggestion
<path d="M7 10H17" />
<path d="M7 14H17" />
```

icons/circle-slash.svg
karsa-mistmere, line 1: A bit unsure about the nomenclature of this one, we already have a `circle-slashed` and a `slash` (that's actually also a `circle-slash`). Maybe `circle-slash-inside`?

karsa-mistmere: Naaah, `circle-off` would be totally different, see below. :D
image

`slash` for sure is a misnomer, but [I don't like `no-entry` either](https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcRfL8Md7FrU1s8IzEqtPGPH5M2LaJnTF44NDg&usqp=CAU).
Maybe `ban`?

danielbayley: > `circle-off` would be totally different

Yeah I also meant it’s probably styled wrong too… or do we need both?

> Maybe `ban`?

Works for me…

karsa-mistmere (approved):
I just brought up `circle-off` as an example, but I guess it doesn't hurt, I'm OK with merging this as is as long as @ericfennis is.

ericfennis (approved):

PR #1033, opened 2023-04-10
Add metadata for `chevron-right-square` icon

danielbayley:


karsa-mistmere:
Haha, yes, we totally forgot to add these, this is definitely good to go.

PR #1035, opened 2023-04-10
Add/refine [`un`]`plug` icons

danielbayley:


karsa-mistmere:
Maybe we could also update plug-zap to use the same shape as `plug` and use that same shape here as well?

An asymmetric shape would perhaps better communicate that the male and female connectors are just parts of the whole.

icons/unplug.svg
karsa-mistmere, line 1: Oh, much much better, I think! :)

icons/plug-zap-2.svg
karsa-mistmere, line 1: @ericfennis: what do you think of dropping our current one in favour of this one? 🤔

This one sticks to the guidelines much better I think.

icons/plug-zap-2.svg
```diff
@@ -0,0 +1,17 @@
...
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M10 20c-1.1 1.1-3 1-4.1-.1l-1.8-1.8C3 17 2.9 15.1 4 14l2-2 6 6-2 2z" />
```
karsa-mistmere, line 12: Maybe just something that I think is a pretty neat trick: if you make diagonal corners have a rounding of 2.41px, the corner will be absolutely pixel perfect:
```suggestion
  <path d="m3.7 17.7 2.6 2.6a2.4 2.4 0 0 0 3.4 0L12 18l-6-6-2.3 2.3a2.41 2.41 0 0 0 0 3.4Z" />
```
image

icons/plug-zap-2.svg
danielbayley: I like both, but I’m easy…

ericfennis: I think it's better to keep both. But not sure

danielbayley: > I think it's better to keep both.

ericfennis (approved):

icons/plug.svg
```diff
@@ -10,7 +10,7 @@
   stroke-linejoin="round"
 >
   <path d="M12 22v-5" />
-  <path d="M9 7V2" />
-  <path d="M15 7V2" />
+  <path d="M9 8V2" />
+  <path d="M15 8V2" />
   <path d="M6 13V8h12v5a4 4 0 0 1-4 4h-4a4 4 0 0 1-4-4Z" />
```
karsa-mistmere, line 15: Saving the world one byte at a time... 😆 
```suggestion
  <path d="M18 8v5a4 4 0 0 1-4 4h-4a4 4 0 0 1-4-4V8Z" />
```

danielbayley: > Saving the world one byte at a time...

If only that was all it took…

Issue #94, opened 2020-10-13
Remove brand icons?


locness3:
Since we don't add any brand icons, I think we should remove the current ones. What do you think?

delnyn:
I agree

ericfennis:
I agree, maybe we can move the current brand icons to a separate repo. I can image people want to use it

ericfennis:
@locness3 @delnyn What about the idea, to create a separate package in packges folder called: 'lucide-brands'. We can move all the brand icons to this separate package, if people like to have this type of brand icons they can use the package.
In the readme we can notice that the brand icons are deprecated and we don't update/maintain them.

delnyn:
Oh yes! Should I clone this one and remove the icons?

locness3:
I don't know, but I think it'd be best if people didn't use them.

mittalyashu:
How about create deprecated section for icons instead of completely removing them?

dgdavid:
Just for the record, this [is/was also a concern in the Feather project](https://github.com/feathericons/feather/issues/763), where is stated that

> [simple-icons/simple-icons](https://github.com/simple-icons/simple-icons) repository already covers everything, uses SVG and the same 24x24 `viewBox`, so it could be a great replacement for people that need brand icons.

mikeott:
Agreed. Brand icons have no business in icon sets.

delnyn:
soooo? we just delete them all?

ericfennis:
I'm thinking of defining the ultimate v1.0 for Lucide and I think this should be a part of it.

delnyn:
what do you mean by "defining the ultimate v1.0 for Lucide"?

SMAH1:
We are going with a lot of requests to add icons. I'm think split project. Part for medical, part for programing, part for brand , ... .
In this case, 'quality, version, update, ...' of each package are different.
In angular, we can one project for handle of use multi package icon library.

delnyn:
I don't understand lol

ericfennis:
> what do you mean by "defining the ultimate v1.0 for Lucide"?
@delnyn I will create a separate discussion for that later. I need write down some points for that.

ericfennis:
@SMAH1 Hmm I'm not fan of this idea. I think it wil add some unnecessary complexity. And can be really confusion as well.

karsa-mistmere:
[...]

PR #1037, opened 2023-04-11
Fix YouTube icon appearance with fill color

wojtekmaj:
When fill color was applied to YouTube icon, it appears as a rectangle with rounded corners, nothing more:

Zrzut ekranu 2023-04-11 o 09 39 58

This PR fixes that by replacing polygon with second path added with `fill-rule="evenodd"`.

Zrzut ekranu 2023-04-11 o 09 43 29

The default appaerance remains unchanged. Before:

Zrzut ekranu 2023-04-11 o 09 44 41

After (I swear it's a different screenshot 😆):

Zrzut ekranu 2023-04-11 o 09 44 18


karsa-mistmere:
Hmm... I'm not sure with #94 being in progress and all, at this point why not just use Simpleicons instead?

https://simpleicons.org/?q=youtube

wojtekmaj:
I'd rather not have multiple icon libraries unless I'm forced to, and it happens that I don't need more brand icons than Lucide offers at a time. I respect that no new icons are added, but I also fail to see the reason not to fix the old ones, no? :)

wojtekmaj:
Can we get this merged, pretty please? 🥺

ericfennis:
Nope sorry we don't accept compound paths. These will nog work well in design programs.

PR #1039, opened 2023-04-11
Add `tag-arrow-*` and `delete-forward` icons

danielbayley:
These cover the `debug-stackframe*` _[codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html)_, and also reperesent debugger in Firefox:

Untitled
<br/>
<br/>

Also related to #1042, and #1319.


karsa-mistmere (approved):

danielbayley:
Note that these had to be renamed, as they were conflicting with `arrow-*-square` icons, since #994 was merged…

ericfennis:
I'm not sure about these names as well. What are other libraries naming these icons?

danielbayley:
> I'm not sure about these names as well. What are other libraries naming these icons?

Well other libraries have different naming schemes… whereas this one tries to be more literal (which I think is better). As mentioned above, these are called `debug-stackframe*` in _codicons_, so no help there…

@ericfennis What about `arrow-pantagon-*`?

ericfennis:
Or maybe `tag-arrow-top`,`tag-arrow-right`, `tag-arrow-bottom, `tag-arrow-left` etc ?

jguddas:
Do we want to generally have the same icon in different orientations?

danielbayley:
> Do we want to generally have the same icon in different orientations?

I’m only interested in the ones pointing right, personally. I just figured they should probably be a set…

But no problem with dropping the others…? @ericfennis @jguddas @karsa-mistmere

ericfennis:
If we drop the others maybe we can think of better name.

Maybe just. `tag-2`?
What is the use case with the dot?

danielbayley:
> Maybe just. `tag-2`?

@ericfennis What about just `tag-right`[`-dot`]?

> What is the use case with the dot?

As [mentioned](https://github.com/lucide-icons/lucide/pull/1039#issue-1662091107) these cover `debug-stackframe` [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html)… So the one with the dot is `debug-stackframe-active`.

ericfennis:
@danielbayley Is it not more convenient to fill the icon for the active state?

danielbayley:
> @danielbayley Is it not more convenient to fill the icon for the active state?

Definitely not more convenient. Not sure even possible, with how VS Code icon themes work…

PR #1040, opened 2023-04-11
Make repository installable on Apple Silicon

wojtekmaj:
* Overrides sharp version used by `svg-outline-stroke` with a newer one to make it installable on Apple Silicon
* Bumps Puppeteer version to a newer one that ships with Chromium that has arm64 builds


pnpm-lock.yaml
```diff
@@ -222,7 +225,7 @@ importers:
       prop-types: 15.8.1
       react: 18.0.0
       react-dom: 18.0.0_react@18.0.0
-      react-native: 0.69.3_isc7crj7cqzlyuhxlfh2hqq5ui
+      react-native: 0.69.3_react@18.0.0
```
wojtekmaj, line 228: Gosh PNPM is a mess. I have no clue all of this stuff is changing.

ericfennis:
@wojtekmaj hmm, looks like the angular test are not working anymore now

karsa-mistmere:
Yeah, I think I've actually fixed that in #939

wojtekmaj:
Do you think we could maybe cherry pick the fix?

karsa-mistmere:
Fixing it was a mess, so we probably can't cherry pick it, but I guess we could make the same changes.
Or maybe just merge #939 then rebase this and see if the issue is solved. 😄

ericfennis:
@wojtekmaj With the current lockfile everything works fine on my Apple Silicon. Let me know if you still have problems.

PR #1041, opened 2023-04-11
Realign baseline/add additional `case-*` icons

danielbayley:
* camelCase
* snake_case
* kebab-case


karsa-mistmere:
I think it would be better for consistency to have kebab and snake case to use the same letter shapes, maybe something like this (3px radius for both instead of 4px for one and 2px for the other):
image

Not sure about the use cases for these though. Where does VSCode use icons like this?
(I'm guessing that's where these originate from :D )

icons/case-snake.svg
```diff
@@ -0,0 +1,17 @@
...
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <line x1="7" x2="7" y1="10" y2="15" />
+  <circle cx="4.5" cy="12.5" r="2.5" />
+  <line x1="11" x2="13" y1="15" y2="15" />
+  <circle cx="19.5" cy="12.5" r="2.5" />
+  <line x1="17" x2="17" y1="8" y2="15" />
```
karsa-mistmere, line 16:
```suggestion
  <path d="M7 10v5" />
  <circle cx="4.5" cy="12.5" r="2.5" />
  <path d="M11 15h2" />
  <circle cx="19.5" cy="12.5" r="2.5" />
  <path d="M17 8v7" />
```

icons/case-kebab.svg
```diff
@@ -0,0 +1,17 @@
...
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="4.5" cy="12.5" r="2.5" />
+  <line x1="7" x2="7" y1="10" y2="15" />
+  <line x1="11" x2="13" y1="12" y2="12" />
+  <circle cx="19.5" cy="12.5" r="2.5" />
+  <line x1="17" x2="17" y1="8" y2="15" />
```
karsa-mistmere, line 16:
```suggestion
  <circle cx="4.5" cy="12.5" r="2.5" />
  <path d="M7 10v5" />
  <path d="M11 12h2" />
  <circle cx="19.5" cy="12.5" r="2.5" />
  <path d="M17 8v7" />
```

icons/case-camel.svg
```diff
@@ -0,0 +1,16 @@
...
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="7" cy="12" r="3" />
+  <line x1="10" x2="10" y1="9" y2="15" />
+  <path d="M14 7h4c1.1 0 2 .9 2 2s-.9 2-2 2h-4V7z" />
+  <path d="M14 11h4.5c1.1 0 2 .9 2 2s-.9 2-2 2H14v-4z" />
```
karsa-mistmere, line 15:
```suggestion
  <circle cx="7" cy="12" r="3" />
  <path d="M10 9v6" />
  <path d="M14 11h4.5a2 2 0 0 1 0 4H14V7h4a2 2 0 0 1 0 4" />
```

karsa-mistmere (changes requested):

karsa-mistmere (approved):

jguddas:
What about PASCAL_CASE?

karsa-mistmere:
Do you mean PascalCase? 🤔

danielbayley:
> What about *PascalCase?

@jguddas Would/could that be covered by https://lucide.dev/icons?search=case-upper? Difficult to express with only 2 letters…

karsa-mistmere:
> > What about *SCREAMING_SNAKE_CASE?
> 
> @jguddas Would/could that be covered by https://lucide.dev/icons?search=case-upper? Difficult to express with only 2 letters…

I think @jguddas meant an icon depicting `A_B` (or some other letters), which is `UPPER_SNAKE_CASE`, also known as `SCREAMING_SNAKE_CASE`

danielbayley:
> I think @jguddas meant an icon depicting … `UPPER_SNAKE_CASE`

I see. Well that one is a bit tricky because the upper case B can’t be made any smaller… But see what you think of my latest `push`ed changes… @jguddas @karsa-mistmere

ericfennis (approved):

jguddas:
How about making them all align to the same baseline rather than just entering them vertically?

danielbayley:
> How about making them all align to the same baseline rather than just entering them vertically?

@jguddas @karsa-mistmere Fixed.

jguddas (approved):

karsa-mistmere (commented):
`case-snake-upper` sticks out like a sore thumb because of the different letter `A` shape, but I'm not sure how it could be fixed. :/

jguddas:
> `case-snake-upper` sticks out like a sore thumb because of the different letter `A` shape, but I'm not sure how it could be fixed. :/

We could change all the other A's to match :man_shrugging:

danielbayley:
> `case-snake-upper` sticks out like a sore thumb because of the different letter `A` shape, but I'm not sure how it could be fixed. :/

@karsa-mistmere @jguddas What about something like these…?

preview

jguddas:
> > `case-snake-upper` sticks out like a sore thumb because of the different letter `A` shape, but I'm not sure how it could be fixed. :/
> 
> Karsa Jakob Guddas What about something like these…?
> 
> preview

Looks terrible IMO.

jguddas:
camel and lower case are also feel inconsistent.

PR #1042, opened 2023-04-11
Add `*-dot` icons

danielbayley:
These correspond to `debug-step-*` _[codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html)_

### DIFF
--- a/icons/dot.json
+++ b/icons/dot.json
@@ -1,0 +1,20 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "interpunct",
+    "interpoint",
+    "middot",
+    "step",
+    "punctuation",
+    "characters",
+    "font",
+    "typography",
+    "type",
+    "text",
+    "center"
+  ],
+  "categories": [
+    "shapes",
+    "text"
+  ]
+}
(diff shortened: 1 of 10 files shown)


icons/arrow-down-dot.json
karsa-mistmere, line 1: Feels a bit incomplete and the names will clash should we decide to add other variations.
I'd suggest to go for a naming scheme like `arrow-up-from-dot` vs. `arrow-down-to-dot`.

karsa-mistmere (changes requested):

icons/arrow-down-dot.json
danielbayley: > Feels a bit incomplete

What’s missing?

karsa-mistmere: What I meant by incomplete is that I could easily see how someone out there at some point might also want any other variation of `arrow-(up|down)-(from|to)-dot`, so we should make sure the current names won't clash with adding these later on.

icons/dot.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "step",
```
karsa-mistmere, line 4:
```suggestion
    "interpunct",
    "interpoint",
    "middot",
    "step",
```

ericfennis (approved):

karsa-mistmere (approved):

PR #1043, opened 2023-04-11
Add/refine `play`/etc… icons

danielbayley:


karsa-mistmere:
Yeah, I'm afraid this just won't do the job. 😐 

I do realize that `stop` is missing from our playback icons (in fact, I have a bunch of playback icon drafts because of this), but - and this is a big but - I'm afraid `square` is not a good fit at all, since it has a completely different style and size than our other playback icons.
image

I can see two possible options:

1) we can add a dedicated `stop` icon that matches our playback icons:

image

2) we can update our current playback icons to match the style of `square`, with the latter standing in as `stop`:

image

@ericfennis, what do you think?

karsa-mistmere:
> I keep opening new cans of worms 😅

Yeah, haha, I've already found this can of worms, and was like
"Oh. Noooope, I'll deal with this later, thank you very much indeed \*glues-back-lid-of-can\*".

ericfennis:
@danielbayley @karsa-mistmere hmm interesting discussion. I think we should get some more opinions on this one. Version 2 from Karsa looks a bit more Lucide-like to me. But maybe from a functionality perspective not always the best choice.

danielbayley:
What about the middle ground, say if we round the corners by 1px instead of 2…?

karsa-mistmere:
So that's possibly option no. 3.

3. Using `square` as is, but other playback icons now having 1px rounded corners:
image

danielbayley:
> So that's possibly option no. 3.
> 
> 3. Using `square` as is, but other playback icons now having 1px rounded corners:
>    image

Yeah I think this is the way to go…

danielbayley:
> 3. Using `square` as is, but other playback icons now having 1px rounded corners:

@karsa-mistmere @ericfennis Pushed this approach…

icons/pause.svg
```diff
@@ -9,6 +9,6 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect width="4" height="16" x="6" y="4" />
-  <rect width="4" height="16" x="14" y="4" />
+  <rect width="4" height="18" x="6" y="3" rx="1" />
+  <rect width="4" height="18" x="14" y="3" rx="1" />
```
ericfennis, line 13: Not sure about increasing the height of this one, can we create side by side comparison? To me the pause icon looks a bit less balanced on 100% scale.

karsa-mistmere: Yeah, I think this also needs an increase in width to match the increase in height (see my original images).

danielbayley: > can we create side by side comparison?

Untitled

Untitled

> I think this would also need an increase in width to match the increase in height

See first row…

karsa-mistmere: Yeah, I think out of the three, the first row matches `square` the best.

ericfennis: @danielbayley Thanks. I checked it with other player control icons. And the first one looks the best, I think.

danielbayley: @ericfennis @karsa-mistmere Yep, that version was `push`ed…

ericfennis:
I think this is good to go, @karsa-mistmere what do you think?

jguddas:
I like them 👍

karsa-mistmere:
> I think this is good to go, @karsa-mistmere what do you think?

Hmm... To be frank, I'm not entirely happy about some of these changes consistency wise, I'll make a side by side comparison as soon as I can get to it.

karsa-mistmere:
OK, so here goes.

This is our current set:
image

Overall pretty consistent, but:
* Square sticks out like a sore thumb
* Skip and steps have inconsistent line heights
* The play shape itself is not very consistent

---

This is the current PR state:
image

It's more Lucide-ish, but:
* Pause and stop are significantly larger than the others
  (`pause` was increased in size to match `square` standing in for stop, but the others were not)

---

# Possible alternatives: 

A) Keeping a consistent arrow shape, but this leads to pretty inconsistent sizes and heights:

image

---

B) Keeping a (reasonably) consistent size and height while matching `square`, but this leads to disproportionate arrow shapes (especially in case of ff and rw):

image

---

C) Adding an extra `stop` icon that allows for better consistency in shape and size _within_ the group, but this makes the group less consistent in size with the rest of the icons in the library:

image

ericfennis:
@karsa-mistmere Thanks for this sum-up. This is very useful. 
For me, I need some time to decide what the best is.

I'm curious about what others think!

karsa-mistmere:
> @karsa-mistmere Thanks for this sum-up. This is very useful. For me, I need some time to decide what the best is.
> 
> I'm curious about what others think!

Yeah, I'm afraid there is no obvious "best" solution here, all of them come with compromises. 😅

danielbayley:
> all of them come with compromises. 😅

So often the case… 🙈

> I'm curious about what others think!

I’m leaning toward option C, but could live with B…

ericfennis:
Ok after looking a second time, option C looks the best to me. Wat are the comprises and inconsistencies beside the stop button?

jguddas:
> this makes the group less consistent in size with the rest of the icons in the library

That sounds not so good.

danielbayley:
> That sounds not so good.

Always tradeoffs 😞 

@jguddas Which option do you vote?

PR #1048, opened 2023-04-12
Improve `apple` metadata

danielbayley:


wojtekmaj:
I would vote for :-1:. This is an apple, a fruit, not Apple logo.

PR #1050, opened 2023-04-12
Add note on use of `<use>`

wojtekmaj:


docs/icon-design-guide.md
```diff
@@ -55,6 +55,8 @@ Here are rules that should be followed to keep quality and consistency when maki
 
 Before an icon is added to the library, we like to have readable and optimized svg code.
 
+Don't use [`<use>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use). While it may seem like a good way to optimize file size sometimes,there's no way to ensure that the referenced element IDs are unique when SVGs are embedded in HTML documents.
```
karsa-mistmere, line 58: Great idea, thanks, I've rephrased some parts.

I think in the long run, we're going to have to explicitly state (and possibly even enforce) which attributes on which elements are allowed at all.

```suggestion
Never use [`<use>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use). While it may sometimes seem like a good way to optimize file size, there's no way to ensure that the referenced element IDs will be unique once the SVGs are embedded in HTML documents.
```

karsa-mistmere (approved):

PR #1052, opened 2023-04-12
Optimize icons using html parser, not babel parser; remove semicolon workaround

wojtekmaj:


karsa-mistmere:
I actually prefer our current syntax, having one path per line helps keep things simple. 🤔

wojtekmaj:
I think the path can be in one line, it does so because of max width. Bigger max width however causes the SVG tag to be one-lined too. This, however, could be prevented by a setting that ensures that there's only one attribute per line.

karsa-mistmere:
Closing this for now, html parsing could be nice, but it should not break the currently existing format.

PR #1054, opened 2023-04-12
Add `rotate`/`refresh` alternate icons

danielbayley:
Closely related to #1176.

### DIFF
--- a/icons/refresh-ccw-2.json
+++ b/icons/refresh-ccw-2.json
@@ -1,0 +1,16 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "arrows",
+    "rotate",
+    "reload",
+    "rerun",
+    "synchronise",
+    "synchronize",
+    "circular",
+    "cycle"
+  ],
+  "categories": [
+    "arrows"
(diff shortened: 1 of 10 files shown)


danielbayley:
Alternative `refresh-cw-off` approaches:

Untitled

ericfennis:
Hmm I'm not sure about this one. I don't see why it is necessary to add a second variant of the refresh icon, to create the refresh `-off` icon. 
I want to ask you design more from the perspective of icons we already have in the lucide library instead of exactly try to reproduce what Codicon has. 
So try to design variant with the existing `refresh` icon first. If that not works, give a brief explanation why it is not working and why we need another one.

danielbayley:
> try to design variant with the existing `refresh` icon first.

I did try that before this PR (see `refresh-cw-off` in #1176)… I don’t think it works quite as well. Plus some of these sync variations also look nice…

karsa-mistmere:
I agree with @ericfennis, I'm afraid. A whole second set of refresh icons seems excessive, and this new group of icons are a bit inconsistent both within the set and with existing icons:
image

PR #1055, opened 2023-04-12
Add `blocks` icons

danielbayley:
`blocks-apart` doubles as [VS Code] `extensions` from [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html).

Alts:
Artboard 1@2x

Untitled


jguddas:
What about adding a border radius?

Not 100% happy, what do you think about something like this? icon (9) square (3)

danielbayley:
> What about adding a border radius?

Sure. See pushed (and optimised) changes…

ericfennis:
Really nice icons @danielbayley 
I agree with @jguddas.
Thanks for making the changes @danielbayley, those look more Lucide-like now.

danielbayley:
> Should the blocks icon not have the same rect as all the other layout icons?

It could be, but then the blocks are different sizes within this pair of icons, since `blocks-apart` can’t be made any bigger…

ericfennis:
Hmm, this one is hard. On the one hand, we want this icon with both versions matching and on the other hand, we want consistency. 

I tried to give it a shot, I used the blocks from `layout-grid` and it kinda works. But the combined blocks variant is a lot smaller. Let me know what you think
image

danielbayley:
> Hmm, this one is hard. On the one hand, we want this icon with both versions matching and on the other hand, we want consistency.

Yeah, seems like it should be simple, but then along come the tradeoffs… Like with so many things! 🙉

> I tried to give it a shot, I used the blocks from `layout-grid` and it kinda works. But the combined blocks variant is a lot smaller. Let me know what you think …

The rounding of `blocks-grid` is different to `blocks`. But anyway I think having `blocks-grid` and `grid-2` seems a bit redundant. So I’m leaning towards just keeping `blocks` and `grid-2`…

@ericfennis What about renaming https://lucide.dev/icons/grid to `grid-3x3` and this one `grid-2x2`?

ericfennis (approved):

ericfennis:
@danielbayley Can you rebase this PR?

danielbayley:
> @danielbayley Can you rebase this PR?

@karsa-mistmere Ready…

icons/blocks.svg
```diff
@@ -0,0 +1,14 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect width="7" height="7" x="14" y="3" rx="1" />
+  <path d="M10 21V8c0-.6-.4-1-1-1H4a1 1 0 0 0-1 1v12c0 .5.5 1 1 1h12c.5 0 1-.5 1-1v-5a1 1 0 0 0-1-1H3" />
```
karsa-mistmere, line 13: The corners could use arcs instead of cubic curves:
```suggestion
  <path d="M10 21V8a1 1 0 0 0-1-1H4a1 1 0 0 0-1 1v12a1 1 0 0 0 1 1h12a1 1 0 0 0 1-1v-5a1 1 0 0 0-1-1H3" />
```

karsa-mistmere (changes requested):

karsa-mistmere (approved):

ericfennis (approved):

PR #1058, opened 2023-04-12
Add `file-code-2` icon

danielbayley:
Related to #1068.


icons/file-code-2.svg
```diff
@@ -0,0 +1,16 @@
...
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M14.5 2H6c-1.1 0-2 .9-2 2v16c0 1.1.9 2 2 2h12c1.1 0 2-.9 2-2V7.5L14.5 2z" />
```
ericfennis, line 12: This path should be the same as the other file icons.
```suggestion
 <path d="M14.5 2H6a2 2 0 0 0-2 2v16a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V7.5L14.5 2z" />
```

ericfennis (changes requested):
Nice icon, left some code suggestions.

karsa-mistmere:
Just an idea: maybe if we add this, let's rename the current `file-code` to `file-code-2` and add this one as `file-code`, so that the layouts are consistent (no suffix => contents inside, `-2` suffix => contents to the side).

danielbayley:
> Just an idea: maybe if we add this, let's rename the current `file-code` to `file-code-2` and add this one as `file-code`, so that the layouts are consistent (no suffix => contents inside, `-2` suffix => contents to the side).

Yeah I noticed this inconsistency also… Would this require adding an alias in this particular case?

karsa-mistmere:
No, we can't add an alias for a name that will already taken.

icons/file-code-2.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "script",
+    "document"
+  ],
+  "categories": [
+    "files",
+    "coding"
```
karsa-mistmere, line 9:
```suggestion
    "development"
```

icons/file-code.json
```diff
@@ -2,7 +2,12 @@
   "$schema": "../icon.schema.json",
   "tags": [
     "script",
-    "document"
+    "document",
+    "gist",
+    "html",
+    "xml",
+    "property list",
+    "plist"
```
karsa-mistmere, line 10: Why not add all these tags to the other version?

karsa-mistmere (changes requested):

PR #1045, opened 2023-04-11
Add generic `window` icons

danielbayley:
Related to #1003…

PR #1061, opened 2023-04-13
Add `search` icons

danielbayley:
Many search icons are the flip of the existing icon… But I think it depends on, for example, which side of a search bar the user wants to display the icon:

image

I think the user should have a choice in this situation.

Also, `search-slash` (clear search) is meant to cover `search-stop` from [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html)

### DIFF
--- a/icons/search-2.json
+++ b/icons/search-2.json
@@ -1,0 +1,13 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "find",
+    "magnifier",
+    "magnifying glass"
+  ],
+  "categories": [
+    "text",
+    "account",
+    "social"
+  ]
+}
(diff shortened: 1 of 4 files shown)


danielbayley:
> Not sure, why exactly do we need this, why won't the current `search` icon do the job?

Sorry, forgot to write the description…

karsa-mistmere:
Not convinced, I've seen search icons using either version, but I don't believe I've never seen an icon set that had both.

karsa-mistmere:
Oh, not sure about `search-slash` either, to me it kind of looks like a lolly. :D

ericfennis:
@danielbayley I appreciate the work. But maybe take a look at my comment in #1045. 
We should not exactly try to reproduce from Codicons, but design more from the icons we already have. The Lucide search icon is just fine, we don't need a mirrored variant.

PR #1069, opened 2023-04-14
Add `combine` icon

danielbayley:
`combine` from [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html).


icons/combine.svg
```diff
@@ -0,0 +1,19 @@
...
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect width="4" height="4" x="2" y="2" />
+  <rect width="4" height="4" x="2" y="10" />
+  <rect width="4" height="4" x="2" y="18" />
+  <line x1="10" x2="14" y1="12" y2="12" />
+  <polyline points="12,10 14,12 12,14" />
+  <rect width="4" height="8" x="18" y="8" />
+  <line x1="18" x2="22" y1="12" y2="12" />
```
jguddas, line 18:
```suggestion
  <circle cx="4" cy="12" r="2"/>
  <circle cx="4" cy="20" r="2"/>
  <circle cx="4" cy="4" r="2"/>
  <rect x="18" y="8" width="4" height="8" ry="2"/>
  <path d="M10 12h4"/>
  <path d="m12 10 2 2-2 2"/>
```

danielbayley: Or more rounded blocks… (see pushed change)

jguddas: Feels a bit cramped no matter the radius, don't know how to fix it but having a bigger more visible arrow is needed IMO.

danielbayley: I have an idea…

danielbayley: Either something like (top left), or maybe better, adapt https://lucide.dev/icons?search=replace to `combine`. What do you think?

Untitled

jguddas: I like it :+1:

karsa-mistmere: Yeah, same here, I think using the opposite of `replace-all` should work pretty well.

danielbayley: Now I’m thinking this should actually be `group` and then also have a counterpart `ungroup`…

karsa-mistmere: Not quite convinced about that, but there are some pretty good ideas for a group/ungroup icon pair in https://github.com/WordPress/gutenberg/issues/15602

danielbayley: > ideas for a group/ungroup icon pair

Maybe that should be left for a separate PR then…

danielbayley:
Should there be some kind of `uncombine` to accompany this one?

ericfennis:
I'm not sure about this one. What's the actual purpose and why is it so abstract?

danielbayley:
> What's the actual purpose and why is it so abstract?

@ericfennis See https://github.com/lucide-icons/lucide/pull/1069#discussion_r1167816330

> I think any use case for this would likely be covered by #1108

Well `combine` and `group` are different things, but sure if we can find a way to convey both…

ericfennis (approved):
@danielbayley thanks for explaining! Look good to me then!

PR #1075, opened 2023-04-15
Simplified `book`

jguddas:


karsa-mistmere:
Separate paths should not be merged.

PR #1088, opened 2023-04-15
Optimized `pizza`

jguddas:


icons/pizza.svg
```diff
@@ -9,9 +9,9 @@
...
-  <path d="M15 11h.01" />
-  <path d="M11 15h.01" />
-  <path d="M16 16h.01" />
-  <path d="m2 16 20 6-6-20c-3.36.9-6.42 2.67-8.88 5.12A19.876 19.876 0 0 0 2 16Z" />
-  <path d="M17 6c-6.29 1.47-9.43 5.13-11 11" />
+  <path d="M15 11h.01"/>
+  <path d="M11 15h.01"/>
+  <path d="M16 16h.01"/>
+  <path d="m2 16 20 6-6-20A20 20 0 0 0 2 16"/>
+  <path d="M17.22 6A15 15 0 0 0 6 17.2"/>
```
karsa-mistmere, line 16: The crust line is currently a bit too small, but with this change it's actually larger than it should be:
image

When fixing it, I realized that it's not even of the correct radius:
image

This should work OK:
```suggestion
  <path d="M15 11h.01" />
  <path d="M11 15h.01" />
  <path d="M16 16h.01" />
  <path d="m2 16 20 6-6-20A20 20 0 0 0 2 16" />
  <path d="M5.71 17.11a17.04 17.04 0 0 1 11.4-11.4" />
```

PR #1097, opened 2023-04-16
feat: added backdrop to gh-icon

jguddas:
Adds a backdrop, so you can see the difference that is introduced by the PR.

### DIFF
--- a/site/src/components/SvgPreview/index.tsx
+++ b/site/src/components/SvgPreview/index.tsx
@@ -173,53 +173,69 @@ const ControlPath = ({
   );
 };
 
-const SvgPreview = React.forwardRef<SVGSVGElement, { src: string; showGrid?: boolean }>(
-  ({ src, showGrid = false }, ref) => {
-    const paths = getPaths(src);
-    const darkModeCss = `@media screen and (prefers-color-scheme: dark) {
+const SvgPreview = React.forwardRef<
+  SVGSVGElement,
+  {
+    src: string | ReturnType<typeof getPaths>;
+    showGrid?: boolean;
+    backdrop?: string;
+  } & React.SVGProps<SVGSVGElement>
+>(({ src, backdrop, children, showGrid = false, ...props }, ref) => {
+  const paths = typeof src === 'string' ? getPaths(src) : src;
+
+  const darkModeCss = `@media screen and (prefers-color-scheme: dark) {
   .svg-preview-grid-group,
   .svg-preview-shadow-mask-group,
   .svg-preview-shadow-group {
     stroke: #fff;
   }
 }`;
-    return (
-      <svg
-        ref={ref}
-        xmlns="http://www.w3.org/2000/svg"
-        width={24}
-        height={24}
-        viewBox="0 0 24 24"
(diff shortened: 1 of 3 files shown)


karsa-mistmere (commented):
Wow, great work! Once again, this is going to be super useful with some PRs. :) 

I have only one qualm with this, and that's using red, since it clashes with the path colors. How about the suggestion below + rearranging the "backdrop" to be the topmost layer?

That way we'd have something like this:
image

Not a perfect solution either, as the shadows look pretty similar to the backdrop, alternatively we could use a palette of colours for the paths that's devoid of any reds.

site/src/components/SvgPreview/index.tsx
```diff
@@ -173,53 +173,69 @@ const ControlPath = ({
...
+>(({ src, backdrop, children, showGrid = false, ...props }, ref) => {
+  const paths = typeof src === 'string' ? getPaths(src) : src;
+
+  const darkModeCss = `@media screen and (prefers-color-scheme: dark) {
   .svg-preview-grid-group,
   .svg-preview-shadow-mask-group,
   .svg-preview-shadow-group {
     stroke: #fff;
   }
 }`;
```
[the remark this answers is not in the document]
jguddas: What about something like this?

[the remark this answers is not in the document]
karsa-mistmere: Neat, but maybe let's not leave any gap or minor losses of detail might stay hidden.
How about also adding a pattern?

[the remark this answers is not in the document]
jguddas: Perfect :+1:

site/src/pages/api/gh-icon/[...data].tsx
```diff
@@ -0,0 +1,100 @@
...
+          >
+            <rect
+              x="0"
+              y="0"
+              width="24"
+              height="24"
+              fill="url(#pattern)"
+              opacity={0.5}
+              stroke="none"
+            />
```
Comment (karsa-mistmere) [inline site/src/pages/api/gh-icon/[...data].tsx:70]: This is pretty edge case, but maybe let's add a second `rect` with no pattern and a lower opacity so that even changes like this will be noticeable:
image

Comment (jguddas) [reply site/src/pages/api/gh-icon/[...data].tsx:70]: I prefer the status quo (left).

Comment (karsa-mistmere) [reply site/src/pages/api/gh-icon/[...data].tsx:70]: Well, yes, out of these two, definitely. But my suggestion would actually be something more like this:
demo

Comment (jguddas) [reply site/src/pages/api/gh-icon/[...data].tsx:70]: Ahh, now I get it.

ericfennis (approved):
Nice work!

karsa-mistmere:
@jguddas: I think there might have been an issue with this PR, our nice preview icons are now gone. :( 
E.g. https://github.com/lucide-icons/lucide/pull/1116

jguddas:
> @jguddas: I think there might have been an issue with this PR, our nice preview icons are now gone. :( E.g. #1116

Oups, #1117 should fix it.

PR #1099, opened 2023-04-16
Optimized `headphones`

jguddas:


karsa-mistmere:
Great work, a nice single line from start to end as it should be. ^_^

PR #1100, opened 2023-04-16
Fixed connection issue and optimized `cpu`

jguddas:


icons/cpu.svg
```diff
@@ -9,14 +9,14 @@
...
+  <rect x="4" y="4" width="16" height="16" rx="2" />
+  <rect x="9" y="9" width="6" height="6" />
+  <path d="M15 2v2" />
+  <path d="M15 20v2" />
+  <path d="M2 14h2" />
+  <path d="M2 9h2" />
+  <path d="M20 14h2" />
+  <path d="M20 9h2" />
+  <path d="M9 2v2" />
+  <path d="M9 20v2" />
```
karsa-mistmere, line 21:
```suggestion
  <path d="M15 2v2" />
  <path d="M15 20v2" />
  <path d="M2 15h2" />
  <path d="M2 9h2" />
  <path d="M20 15h2" />
  <path d="M20 9h2" />
  <path d="M9 2v2" />
  <path d="M9 20v2" />
```

I think these two lines are probably also misaligned and not on purpose, so maybe fix these?

PR #1102, opened 2023-04-16
Optimized `rocket`

jguddas:
Not 100% lossless.


icons/rocket.svg
karsa-mistmere, line 1: Some sharp corners are introduced: 
image

karsa-mistmere (changes requested):

PR #1105, opened 2023-04-16
Optimized `instagram`

jguddas:


icons/instagram.svg
```diff
@@ -9,7 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect width="20" height="20" x="2" y="2" rx="5" ry="5" />
-  <path d="M16 11.37A4 4 0 1 1 12.63 8 4 4 0 0 1 16 11.37z" />
-  <line x1="17.5" x2="17.51" y1="6.5" y2="6.5" />
+  <circle cx="12" cy="12" r="4" />
+  <rect x="2" y="2" width="20" height="20" rx="5" />
+  <path d="M 17.5 6.5h0.01" />
```
karsa-mistmere, line 14: If we're to touch this in spite of #94 we should probably also make it stick to the guidelines more:
```suggestion
  <circle cx="12" cy="12" r="3" />
  <rect x="3" y="3" width="18" height="18" rx="5" />
  <path d="M17 7h.01" />
```

PR #1108, opened 2023-04-17
Add [`un`]`group` icons

danielbayley:
Alternatives:

Maybe the `selection` (dashed) version would be more appropriate…?


danielbayley:
> A few alternate drafts:

I quite like the first actually…

ericfennis (approved):

PR #1110, opened 2023-04-17
Add `markdown` icon

danielbayley:
Alternatives:

Untitled

Untitled


karsa-mistmere:
I think this is find, although if we were to make it a bit larger, to me, this would seem most logical:
markdown

(w/ the M and the arrow both being 8 by 8 pixels)

danielbayley:
Opinions on the boxed version? (more like the actual Markdown icon)

karsa-mistmere:
> Opinions on the boxed version? (more like the actual Markdown icon)

Yeah I can see where that's coming from, but I think it's just way too dense, especially the M, on 100% is looks more like an H.

danielbayley:
> (w/ the M and the arrow both being 8 by 8 pixels)

@karsa-mistmere `push`ed that version…

jguddas:
I think markdown needs the box to be identifiable, and the box version looks too cramped.

:-1: for adding a markdown icon.

danielbayley:
> I think markdown needs the box to be identifiable

Added a few alternatives [above](https://github.com/lucide-icons/lucide/pull/1110#issue-1671736297)…

Although to be honest I think the boxed version is ok, and obviously recognisable as what it is. Sure, ideally we’d have more space but…

tryoxiss:
I think this icon makes a lot of sense. Personally I think the M and arrow is recognisable enough on its own to not need the box, and the version with top and bottom lines helps without cramping it.

I also don't think markdown could be considered a brand icon like #94 is about, since markdown is less one homogonus thing and more a group of simillar markup languages. Just off the top of my head there is CommonMark, MarkdownExtra, Misskey Markdown, Github Flavoured Markdown, and I know there are many more. Unless being technechal like this, they are all just "markdown" and people don't bother to distingush.

SaadBazaz:
Need this. Probably a staple for devtools. Lucide seems to be used in many of them...

PR #1111, opened 2023-04-17
Persistent icon release data caching via workflow

karsa-mistmere:
With the growing number of icons, it's getting more and more costly to fetch the list of contributors for each icon when building the site, currently oftentimes resulting in some contributors missing from the site due to the exhaustion of GitHub API rate limits.

This PR does the following:
1) Creates a release metadata cache of all icon files as a JSON file in the root repository.
   The cache includes the following metadata about icon files:
   - release added
   - release when last modified
   - list of contributors
   - date when the icon was last modified (for checking cache up-to-dateness)
2) This cache has been initialized with all metadata about current icons, including the list of original Feather contributors (something we're unfortunately missing right now, even though credit should be given where credit is due).
3) Alters the GitHub CI workflow so that on changes to main not just a changelog is generated, but the metadata above is fetched and cached for any icons added or modified.
4) Adds a helper method to allow for manually updating the cache. This helper method uses `simple-git` to only update icons where the cache is not up to date (see 1)
5) Adds a pre commit hook to check the validity of the release metadata cache (no keys that do not have a matching SVG file).


ericfennis:
@karsa-mistmere Can you get this one rebased?

jguddas:
What about splitting the `icon-releases.json` up, we could even add it to the `icons/name.json` file?
Then we don't run into the same merge problems and general unwieldiness we had with `tags.json` and `categoriest.json`.

What about always using simple-git and dropping the GitHub API entirely, maybe later as a v2?

The fields could be consistent with the naming and have a `created` field as well.
```diff
  createdRealease
- changedRelease
+ updatedRelease
  updated
+ created
```

karsa-mistmere:
> What about splitting the `icon-releases.json` up, we could even add it to the `icons/name.json` file? Then we don't run into the same merge problems and general unwieldiness we had with `tags.json` and `categoriest.json`.

I'm not sure this will be necessary, since this file won't really be manually edited. We've deliberately chosen not to add these to the `icons/[icon-name].json` files for this very reason (so that people don't mistakenly fill this information out for new icons).
I guess we could certainly split this up into separate files in a new folder. What do you think @ericfennis ?

> What about always using simple-git and dropping the GitHub API entirely, maybe later as a v2?

Not an option, since we need to access the GitHub contributor usernames, which are only available through the API.

> The fields could be consistent with the naming and have a `created` field as well.

Consistent with what naming? 🤔 
We could certainly add a `created` timestamp, but `updated` is only there to help with invalidating cache to begin with. :)

karsa-mistmere:
> @karsa-mistmere Can you get this one rebased?

Done, also updated the cache.

jguddas:
What about generating the file during build time and using a `contributors.json` file that includes the `email` and `username`?

karsa-mistmere:
> What about generating the file during build time and using a `contributors.json` file that includes the `email` and `username`?
> 
> And if the build fails due to missing a contributor we get a `please run pnpm generate:contributors` error.

Um... I'm not sure I understand correctly, but if I do, then:
- Firstly, that's exactly what we're trying to avoid here (having to generate this at build time, because it keeps failing)
- We cannot create a cache of contributors that we would be able to match up with information available via `simple-git` because the GitHub API does not provide e-mail addresses for contributors: https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repository-contributors

jguddas:
> the GitHub API does not provide e-mail addresses for contributors
This is not perfect, so it might fail, but it is pretty good and has 500k+ weekly downloads.

https://www.npmjs.com/package/github-username

karsa-mistmere:
> > the GitHub API does not provide e-mail addresses for contributors
> 
> This is not perfect, so it might fail, but it is pretty good and has 500k+ weekly downloads.
> 
> https://www.npmjs.com/package/github-username

Yes, although to build a contributors cache, we would need to query the mail addresses by username and not vice versa but I guess we could use the same principle (https://api.github.com/search/commits?q=author:USERNAME)

ericfennis:
@karsa-mistmere @jguddas Hmm, yeah both have some drawbacks. probably.
These are I think the pros and cons for both solutions.

**Release info in [name].json file:**
Cons:
- People will maybe change this accidentally or will fill it in. 
Pro:
- One overview of the icon in one, you also see who It made. And when it was released.
- Nice for metadata improvement on the website

**Release info in separate JSON File**
Cons:
- Huge file, Not really human-readable.
- If two PRs are merged directly after each other, there is a chance that one will fail because the branch not in sync when it tries to update the main branch.
- Possible merge conflicts for people who are working in their own branches.
Pro:
- Keeps current JSON files simple
- People won't touch this metadata.

Please post more PROs and Cons, if you know more.

I maybe feel more for the [name].json, but not totally convinced yet. 🤔

jguddas:
> While so to speak this is of course public information, I feel quite a bit uneasy about adding contributors e-mail addresses to the repository in such an explicit way. 😕

Valid, we could hash them.

ericfennis:
@jguddas I think only username is fine. Why should we need the email?

jguddas:
> Jakob Guddas I think only username is fine. Why should we need the email?

We can get the email from the git history, so we don't need to fetch on every commit, only when there are new contributors.

image

Another cool part about using emails is that you can also include co-authors. https://regex101.com/r/RQ22hZ/1

karsa-mistmere:
@jguddas: Nice comparison chart, but I'm not sure where in your version do we build the release/contributors metadata for icons? 🤔

karsa-mistmere:
Closing this in favour of #1218  + we can fetch release data at build time using `simple-git` without having to use any GitHub API requests whatsoever.

PR #1114, opened 2023-04-18
Add `list-tree` icon

danielbayley:
Like `list-tree` from [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html), and a prominent feature in VS Code…

### DIFF
--- a/icons/list-tree.json
+++ b/icons/list-tree.json
@@ -1,0 +1,12 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "tree",
+    "browser"
+  ],
+  "categories": [
+    "files",
+    "text",
+    "layout"
+  ]
+}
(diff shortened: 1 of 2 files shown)


karsa-mistmere:
To be honest, I'm not quite sure about this one, it might work for codicons, but doesn't quite fit well with the Lucide style.
What do you think about something like this?
list-tree

It's basically a hybrid between `list` and `folder-tree`:
image

danielbayley:
> What do you think about something like this?

Like it!

icons/list-tree.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "tree",
+    "browser"
+  ],
+  "categories": [
+    "files",
+    "text"
```
karsa-mistmere, line 9:
```suggestion
    "text",
    "layout"
```

karsa-mistmere (approved):

ericfennis (approved):

PR #1121, opened 2023-04-20
Refine `list` icons

danielbayley:
Some of the existing `list` icons feel a bit off…

Some alternatives explored:

Untitled

### DIFF
--- a/icons/list-minus.svg
+++ b/icons/list-minus.svg
@@ -9,8 +9,8 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M11 12H3" />
-  <path d="M16 6H3" />
-  <path d="M16 18H3" />
-  <path d="M21 12h-6" />
+  <path d="M21 6H3" />
+  <path d="M13 12H3" />
+  <path d="M21 18H3" />
+  <path d="M21 12h-4" />
 </svg>
--- a/icons/list-plus.svg
+++ b/icons/list-plus.svg
@@ -9,9 +9,9 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M11 12H3" />
-  <path d="M16 6H3" />
-  <path d="M16 18H3" />
-  <path d="M18 9v6" />
-  <path d="M21 12h-6" />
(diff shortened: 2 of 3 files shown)


danielbayley:
> the plus sign is also way too small.

What about the approach in the 2nd row?

karsa-mistmere:
> > the plus sign is also way too small.
> 
> What about the approach in the [2nd row](https://github.com/lucide-icons/lucide/pull/1121#issue-1676223595)?

That would be an option, but we would need a matching design for `list-minus`

danielbayley:
> but we would need a matching design for `list-minus`

[Comment](https://github.com/lucide-icons/lucide/pull/1121#issue-1676223595) updated…

karsa-mistmere:
Still not entirely sure about this, the two sets feel quite different in meaning.

To me the current icons seem to be communicating the addition or deletion of a single item to a list, yours read much more like "create a new list", "remove list" etc.

ericfennis:
I'm not sure about this one either. The `X`, `minus`, and `plus` are less obvious now they are in the list icon. I think this breaks a bit the functionality of the icon.

I think we should keep it as it is.

danielbayley:
> the two sets feel quite different in meaning.
> To me the current icons seem to be communicating the addition or deletion of a single item to a list, yours read much more like "create a new list", "remove list" etc.

Well `list-x` in particular (within the context I am trying to use it) is meant to be “clear all”… Maybe more appropriate as (clear) `text-x`?

karsa-mistmere:
I vote not to move forward with these changes, I think the 100% view is much less clear with the new versions.

ericfennis:
I agree with @danielbayley, these changes break the function of these icons. by decreasing the size of the plus in `list-plus` the plus looks a bit like a dot on 100% scale. Please when designing icons try to look more on 100% scale. I've the feeling with a lot of PRs I need to mention this too many times.

Icons are mainly used and rendered on 100% scale in applications, so it is important that the icon is recognizable and tells the message of the use case of the icons. So sometimes icons maybe looks a bit weird designed on a zoomed-in scale, but when zoomed out the icon is telling the message perfectly.


I close this one for now.

PR #1009, opened 2023-04-03
Add `folder-git` icon

danielbayley:

PR #1128, opened 2023-04-20
Fix `folder-git` icons

danielbayley:
Oops, looks like some `class`es sneaked through, because of the issue (before it was) fixed in #1126…

But also, on a closely related note, what do we think of these? @karsa-mistmere @ericfennis @jguddas 

I’m looking for some kind of `folder-root` (`git` repo or otherwise) icon, for the file explorer/tree in my VS Code extension…

Untitled

Untitled

Related to #1009.


jguddas:
> I’m looking for some kind of folder-root (git repo or otherwise) icon, for the file explorer/tree in my VS Code extension…

Wouldn't drive(hard-drive (1)) cover most use cases?

jguddas:
@karsa-mistmere so cool :tada: 
image

danielbayley:
> Wouldn't drive(hard-drive (1)) cover most use cases?

Well the root would be one or more project folders, not the root of the hard drive…

jguddas:
I don't know what icon would mark something as project root, I don't associate anything with it :man_shrugging:

danielbayley:
> I don't know what icon would mark something as project root, I don't associate anything with it 🤷‍♂️

I mean in VS Code, and Atom, root folders have a different icon than normal/subfolders…

Untitled

Untitled

karsa-mistmere:
@danielbayley I don't really associate "project root" with any image, I guess `folder-dot` could work and still have generic usages, but let's make sure to use the same dot for all of these (`bell-dot`, `braces-dot` etc)

karsa-mistmere (changes requested):

karsa-mistmere (approved):

PR #1131, opened 2023-04-21
Rename `text-selection` to `text-select`

danielbayley:
This should have been called `text-select`, consistent with `box-select`…


ericfennis (approved):
Nice

PR #1063, opened 2023-04-13
Add/rename `braces`/`brackets`/`parenthesis` icons and improve `braces`/`json` metadata

danielbayley:

PR #1134, opened 2023-04-21
Add symbols (`constant`, `variable` etc…) icons

danielbayley:
This PR aims to cover relevant (we already have https://lucide.dev/icons?search=function-square, for example) programming `symbol-*` icons from [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html), used in VS Code breadcrumbs, Outline panel and other places…

`variable` related to #1063.

### DIFF
--- a/icons/constant.svg
+++ b/icons/constant.svg
@@ -1,0 +1,15 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect width="18" height="12" x="3" y="6" rx="2" />
+  <path d="M7 10h10" />
+  <path d="M7 14h10" />
+</svg>
--- a/icons/pi.svg
+++ b/icons/pi.svg
@@ -9,7 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M10 4v16" />
-  <path d="M4 9c0-2.8 2.2-5 5-5h9" />
-  <path d="M18 18a2 2 0 0 1-4 0V4" />
+  <line x1="9" x2="9" y1="4" y2="20" />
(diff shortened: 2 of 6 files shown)


icons/constant.svg
karsa-mistmere, line 1: Not sure about this one, it's a very generic shape and in and of itself it doesn't quite communicate anything specific, it's also a bit smaller than our similarly shaped icons.

What do you think of adding `pi` instead and using that to represent a constant?

danielbayley: It’s basically a carbon copy of the existing `symbol-constant`.

> What do you think of adding pi instead and using that to represent a constant?

Already had it as a draft, based off of https://lucide.dev/icons?search=pilcrow!

danielbayley:
Alternatively:

Untitled

danielbayley:
Alternatively:

Untitled

icons/pi.svg
```diff
@@ -0,0 +1,15 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M4 9c0-2.8 2.2-5 5-5h9" />
+  <line x1="10" x2="10" y1="4" y2="20" />
+  <path d="M18 18c0 1.1-.9 2-2 2s-2-.9-2-2V4" />
```
jguddas, line 14: Not a big fan of having letters as icons.
Anyway this can be optimized:
```suggestion
  <path d="M10 4v16" />
  <path d="M4 9c0-2.8 2.2-5 5-5h9" />
  <path d="M18 18a2 2 0 0 1-4 0V4" />
```

danielbayley: > Not a big fan of having letters as icons.

Is Pi technically a letter…?

karsa-mistmere: > > Not a big fan of having letters as icons.
> 
> Is Pi technically a letter…?

It sure it. 😄

karsa-mistmere (approved):

ericfennis (approved):

PR #1136, opened 2023-04-21
Add `monitor-`[`up`/`down`] icons

danielbayley:


jguddas (approved):

ericfennis (approved):
Nice work!

icons/monitor-up.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "tv",
+    "screen",
+    "display",
+    "upload",
+    "connect",
+    "remote"
```
karsa-mistmere, line 9:
```suggestion
    "remote",
    "screen share"
```

icons/monitor-down.json
```diff
@@ -0,0 +1,18 @@
...
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "tv",
+    "screen",
+    "display",
+    "desktop",
+    "download",
+    "clone",
+    "version control",
+    "git"
```
karsa-mistmere, line 11: I like the icon in general, but I'm not sure about these tags, I don't think I've ever seen this type of icon being used for git clone. 🤔

danielbayley: Hmm, I was going off this…

Untitled

karsa-mistmere: I think that's maybe more about GitHub _Desktop_ and less about `git clone`. 😄 

Same with the other "clone" icon in this dropdown, which stands in for "git clone" _using CLI_:

image

PR #1137, opened 2023-04-21
Changed `spline` start and end to be circular

jguddas:


ericfennis (commented):
I'm ok with this change. Can you also update `scale-3d`?

ericfennis (approved):
nice work!

PR #1138, opened 2023-04-21
feat/backdrop improvements

jguddas:
- fix: reduced z-fighting
- fix: fixed outline intensity issue

---
name: New icon
about: Add a new icon to the library
labels: "🎨 <icon"
---

- **Name of the icon** : 
- **Tags (alternative names for this icon)** (add them in as a separate json file using the same icon name) :
- **What is the purpose of this icon?** : 
- **100% scale preview** : 
- **Have you considered alternative possibilities** for its naming or design? :

### DIFF
--- a/site/src/pages/api/gh-icon/[...data].tsx
+++ b/site/src/pages/api/gh-icon/[...data].tsx
@@ -68,7 +68,7 @@ export default async function handler(req, res) {
               opacity={0.5}
               stroke="none"
             />
-            <g dangerouslySetInnerHTML={{ __html: src }} />
+            <g dangerouslySetInnerHTML={{ __html: src }} stroke="red" />
           </g>
           <rect
             x="0"
--- a/site/src/pages/api/gh-icon/[...data].tsx
+++ b/site/src/pages/api/gh-icon/[...data].tsx
@@ -47,11 +47,11 @@ export default async function handler(req, res) {
           </defs>
           <mask id="svg-preview-backdrop-mask-outline" maskUnits="userSpaceOnUse">
             <g stroke="#fff" dangerouslySetInnerHTML={{ __html: backdropString }} />
-            <g dangerouslySetInnerHTML={{ __html: src }} />
+            <g dangerouslySetInnerHTML={{ __html: src }} strokeWidth={2.05} />
           </mask>
           <mask id="svg-preview-backdrop-mask-fill" maskUnits="userSpaceOnUse">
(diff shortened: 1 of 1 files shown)


karsa-mistmere:
I think it's better not to leave any gap, I think something like this would probably be most logical (i.e. why add an extra outline for the _current_ icon?)

demo

jguddas:
ICAgPHBhdGggZD0iTTE0IDUuMjU2QTguMTQ4IDguMTQ4IDAgMCAwIDEyIDVhOS4wNCA5LjA0IDAgMCAwLTIgLjIyN00yMC4wOTggMTBjLjU3MiAxLjA2OC45MDIgMi4yNC45MDIgMy40NDRDMjEgMTcuODkgMTYuOTcgMjEgMTIgMjFzLTktMy05LTcuNTU2YzAtMS4yNTEuMjg4LTIuNDEuNzkyLTMuNDQ0IiAvPiAgIDxwYXRoIGQ9Ik0
PHN2ZwogIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIKICB3aWR0aD0iMjQiCiAgaGVpZ2h0PSIyNCIKICB2aWV3Qm94PSIwIDAgMjQgMjQiCiAgZmlsbD0ibm9uZSIKICBzdHJva2U9ImN1cnJlbnRDb2xvciIKICBzdHJva2Utd2lkdGg9IjIiCiAgc3Ryb2tlLWxpbmVjYXA9InJvdW5kIgogIHN0cm9rZS1saW5lam9

Before changing it to `2.05` you would see a slight red shimmer on some devices, it's not terrible, but it's there :man_shrugging:.

Now the axe has no red shimmer due to the `0.05px` gap.

karsa-mistmere:
Yeah, I guess anything hidden in there is probably too minor an issue to address.

I've been thinking, what if the coloured paths were translucent and the previous path was actually behind them? (maybe we could add that as a hover effect, it won't work on Github, but will once you click the image).
This might be a bit too much clutter though. 🤔 

demo2

jguddas:
Yeah, there is both `added` and `changed` that are currently not displayed, gonna have to think about it.

But I think the `deleted` is already quite nice and a good start.

jguddas:
Can we get this merged?

karsa-mistmere (approved):

karsa-mistmere:
I don't usually like to merge things without two approves, but this is a pretty minor change, so here goes. :)

PR #1142, opened 2023-04-22
Add geometry icons

danielbayley:
Related to #1134 and #1285. `cuboid` partially implements #1368 and #1369.

Edit: `area` inspired by the beautiful _[Things](https://culturedcode.com/things)_ app, and nicely matches  https://lucide.dev/icon/box /`package` etc…

Untitled


jguddas:
What about using dashed lines for the tangent / radius marker?

karsa-mistmere:
> What about using dashed lines for the tangent / radius marker?

Just a rough preview to give an overall impression:
image

jguddas:
I like those image

karsa-mistmere:
> I like those image

Yeah, they're not the cleanest looking, but at least match `spline` pretty well.

karsa-mistmere:
> @karsa-mistmere Wouldn’t the bottom row more represent `bezier-curve`?

I guess it could also represent that, although Bézier curve do have tangents, so... 😄 

`circumference` on the other hand is very definitely `diameter` (unless `radius` isn't radius after all).

jguddas:
Very pretty ❤️

danielbayley:
> `circumference` on the other hand is very definitely `diameter` (unless `radius` isn't radius after all).

Ha of course it is! Totally wasn’t thinking… Good catch.

icons/area.svg
karsa-mistmere, line 1: Like this as a general rule, but there should probably be better consistency between this and `layers`:
image

danielbayley: @karsa-mistmere Agreed. With that in mind, here are some options…

Untitled

Artboard 1@2x

icons/drafting-compass.svg
```diff
@@ -0,0 +1,17 @@
...
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="12" cy="5" r="2" />
+  <path d="M11 6.7 3 21" />
+  <path d="M14.9 10.1 13 6.7" />
+  <path d="M18.4 12.4c-3.5 3.5-9.2 3.5-12.7 0" />
+  <path d="m21 21-2.2-3.9" />
```
karsa-mistmere, line 16: Really nice icon, but once again, the optimisation script has been a bit overzealous:
```suggestion
  <circle cx="12" cy="5" r="2" />
  <path d="m3 21 8.02-14.26" />
  <path d="m12.99 6.74 1.93 3.44" />
  <path d="M19 12c-3.87 4-10.13 4-14 0" />
  <path d="m21 21-2.16-3.84" />
```

jguddas:
Those icons are super lovely.

The `area` icon might be a bit small tho.

The `layers` feels more like `layer` now, but it looks nice :+1:

icons/cylinder-volume.svg
```diff
@@ -0,0 +1,18 @@
...
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M14 7.9c4-.3 7-1.5 7-2.9 0-1.7-4-3-9-3S3 3.3 3 5c0 1.4 3 2.6 7 2.9" />
```
jguddas, line 12:
```suggestion
  <path d="M10 7.9C6 7.6 3 6.4 3 5a9 3 0 0 1 18 0c0 1.4-3 2.6-7 2.9" />
```

icons/diameter.svg
```diff
@@ -0,0 +1,17 @@
...
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M20.3 17.5A10 10 0 0 0 6.5 3.7" />
+  <circle cx="5" cy="5" r="2" />
+  <path d="m6.4 6.4 11.2 11.2" />
+  <circle cx="19" cy="19" r="2" />
+  <path d="M3.7 6.5a10 10 0 0 0 13.8 13.8" />
```
jguddas, line 16: Was not 100% circular before
```suggestion
  <circle cx="19" cy="19" r="2" />
  <circle cx="5" cy="5" r="2" />
  <path d="M20.3 17.5A9.95 9.95 0 0 0 6.5 3.7" />
  <path d="m6.4 6.4 11.2 11.2" />
  <path d="M3.7 6.5a9.95 9.95 0 0 0 13.8 13.8" />
```

jguddas:
What about something like this to represent area? circle (4)

jguddas:
Something like this feels more like layer**s**, but it sadly has less than 2px spacing :disappointed:  icon (50)

danielbayley:
> Something like this feels more like layer**s**, but it sadly has less than 2px spacing 😞

Yeah 😒 @jguddas See comparison table: https://github.com/lucide-icons/lucide/pull/1142#discussion_r1196665209

icons/area.svg
danielbayley: Presumably you mean `layers`… I think we should have the existing as `layers` and the variation underneath as `layers-2`. So are you thinking column 3? @karsa-mistmere

icons/cone.svg
```diff
@@ -0,0 +1,13 @@
...
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="m12 2 9 17c0 1.7-4 3-9 3s-9-1.3-9-3Z" />
```
karsa-mistmere, line 12: I think this could definitely have a taller base:
```suggestion
  <path d="M12 2 3 18A9 4 0 0 0 21 18Z" />
```

image
image

danielbayley: @karsa-mistmere That looks better, but shouldn’t it match `cylinder`/`database`…?

jguddas: I prefer @karsa-mistmere proposal.

jguddas:
> > Something like this feels more like layer**s**, but it sadly has less than 2px spacing disappointed
> 
> Yeah unamused @jguddas See comparison table: [#1142 (comment)](https://github.com/lucide-icons/lucide/pull/1142#discussion_r1196665209)

Can you share the third row as SVG?

icons/cone.svg
karsa-mistmere: > @karsa-mistmere That looks better, but shouldn’t it match `cylinder`/`database`…?

Ugh, I've completely forgotten about that... 😩 

In general, I guess yes, it should. In this instance however, I think it's better to make this more easily recognisable even if we lose some consistency, since without the added base height, it looks too much like a pie slice.

Side by side they don't look that bad, I think:
image

icons/area.svg
karsa-mistmere: `layers` is the strongest in column no. 3, for area and the alt it's column no. 4.
Why. Always. So. Many. Compromises. 😩 
I'd wait for @ericfennis on this.

danielbayley: > Why. Always. So. Many. Compromises. 😩

@karsa-mistmere Tell me about it! 🙈

Would it be a criminal offence to use both? (column 3 for `layers`, and col 4 for `layers-2` and `area`…)

danielbayley:
> What about something like this to represent area?

I like the isometric `area`, but other possible options I had…

Untitled
Artboard 1@2x
<br/>
<br/>

> Can you share the third row as SVG?

@jguddas Here you go…

area-layers

icons/cone.svg
danielbayley: Hmm, I much prefer the version I already `push`ed 😅 …It’s much more obviously a cone.

This is where we need that poll bot again…

icons/layers.svg
```diff
@@ -9,7 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <polygon points="12 2 2 7 12 12 22 7 12 2" />
-  <polyline points="2 17 12 22 22 17" />
-  <polyline points="2 12 12 17 22 12" />
+  <path d="M22 7c0-.3-.2-.6-.6-.8l-7.6-3.8a4.78 4.78 0 0 0-3.6 0L2.6 6.2a.94.94 0 0 0 0 1.6l7.6 3.8a4.77 4.77 0 0 0 3.6 0l7.6-3.8c.4-.2.6-.5.6-.8" />
+  <path d="m2 12.5 8.2 4.1a4.77 4.77 0 0 0 3.6 0l8.2-4.1" />
+  <path d="m2 17.5 8.2 4.1a4.77 4.77 0 0 0 3.6 0l8.2-4.1" />
```
jguddas, line 14: Yay? icon (55)
```suggestion
  <path d="m6 14.5-3.4 1.7a.83.83 0 0 0 0 1.6l7.6 3.8a4.14 4.14 0 0 0 3.58 0l7.62-3.8a.83.83 0 0 0 0-1.6L18 14.5" />
  <path d="m6 9.5-3.4 1.7a.83.83 0 0 0 0 1.6l7.6 3.8a4.14 4.14 0 0 0 3.58 0l7.62-3.8a.83.83 0 0 0 0-1.6L18 9.5" />
  <path d="M13.8 2.4a4.14 4.14 0 0 0-3.58 0L2.6 6.2a.83.83 0 0 0 0 1.6l7.6 3.8a4.14 4.14 0 0 0 3.58 0l7.62-3.8a.83.83 0 0 0 0-1.6Z" />
```

jguddas:
Area could also have more than 4 corners.

image

icons/layers.svg
danielbayley: Fine with me. I also had this in drafts, but figured maybe it might be better to leave the choice between styles… @karsa-mistmere What do you think?

danielbayley: > Yay?

@jguddas Added as `layers-3` (which actually works out nicely in this case, as `layers-2` is literally 2 layers, and `layers-3` has 3… 👌🏼)

danielbayley:
> Area could also have more than 4 corners.

@jguddas See https://github.com/lucide-icons/lucide/pull/1142#issue-1679447215

icons/area.svg
```diff
@@ -0,0 +1,14 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M21 16V8a2 2 0 0 0-1-1.73l-7-4a2 2 0 0 0-2 0l-7 4A2 2 0 0 0 3 8v8a2 2 0 0 0 1 1.73l7 4a2 2 0 0 0 2 0l7-4A2 2 0 0 0 21 16z" />
+  <path d="m3.3 7 7.7 4.4a1.82 1.82 0 0 0 2 0L20.7 7" />
```
jguddas, line 13: Version based on `layers-2` icon (58)

icons/layers-2.svg
```diff
@@ -0,0 +1,14 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M22 8c0-.4-.3-.8-.8-1.1l-7.6-4.5a2.93 2.93 0 0 0-3.1 0L2.8 6.9c-.5.3-.8.7-.8 1.1s.3.8.8 1.1l7.6 4.5a3.3 3.3 0 0 0 3 0L21 9.1c.7-.3 1-.7 1-1.1" />
+  <path d="m7.7 12-4.9 2.9c-.5.3-.8.7-.8 1.1s.3.8.8 1.1l7.6 4.5a3.3 3.3 0 0 0 3 0l7.6-4.5c.5-.3.8-.7.8-1.1s-.3-.8-.8-1.1L16.3 12" />
```
jguddas, line 13: What about calling this `two-layers`?

IMO we should just make it a hard requirement to come up with a good name before we merge even the cutest of icons.

icons/layers-3.svg
```diff
@@ -0,0 +1,15 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="m6 14.5-3.4 1.7a.83.83 0 0 0 0 1.6l7.6 3.8a4.14 4.14 0 0 0 3.58 0l7.62-3.8a.83.83 0 0 0 0-1.6L18 14.5" />
+  <path d="m6 9.5-3.4 1.7a.83.83 0 0 0 0 1.6l7.6 3.8a4.14 4.14 0 0 0 3.58 0l7.62-3.8a.83.83 0 0 0 0-1.6L18 9.5" />
+  <path d="M13.8 2.4a4.14 4.14 0 0 0-3.58 0L2.6 6.2a.83.83 0 0 0 0 1.6l7.6 3.8a4.14 4.14 0 0 0 3.58 0l7.62-3.8a.83.83 0 0 0 0-1.6Z" />
```
jguddas, line 14: What about using this as `layers`, refine the one we have instead of creating variants.

icons/plot.svg
```diff
@@ -0,0 +1,15 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M22 12c0-.4-.3-.8-.8-1.1l-7.6-4.5a2.93 2.93 0 0 0-3.1 0l-7.6 4.5c-.6.3-.9.7-.9 1.1s.3.8.8 1.1l7.6 4.5a3.3 3.3 0 0 0 3 0l7.6-4.5c.7-.3 1-.7 1-1.1" />
+  <path d="m17.5 8.7-11 6.6" />
+  <path d="m6.6 8.8 10.8 6.4" />
```
jguddas, line 14: Is the use case a plot of land, then maybe something like this could be added.
icon (59)

Naming it just plot is kinda unclear, maybe something like `land-parcel` is better in that case.

icons/area.svg
jguddas: Also not sure if `area` is the best name.

danielbayley: > Version based on `layers-2`

@jguddas Also nice! I’m fine with either, but maybe the cube version would be better for consistency… What do you think? @karsa-mistmere

icons/layers-2.svg
danielbayley: > IMO we should just make it a hard requirement to come up with a good name before we merge even the cutest of icons.

@jguddas Agree in general we should strive for better naming, but not sure about how _hard_ the requirement should be… In some cases—and specifically in this case—`-2` & `-3` actually work well as coincidentally matching the number of layers. cc: @karsa-mistmere @ericfennis 

> What about calling this `two-layers`?

I don’t see how `two` is preferable to `2` here… Typing `two` instead of `2` in search is less user friendly.

icons/layers-3.svg
danielbayley: > What about using this as `layers`, refine the one we have instead of creating variants.

@jguddas I’d be fine with that, but I think @karsa-mistmere want’s to keep the existing `layers` around, based on https://github.com/lucide-icons/lucide/pull/1142#discussion_r1196963327 …

> Not really a fan of changing the essence of `layers` so drastically. 😕

jguddas:
They all look cute.

But...

- `area` is not easy to understand
- `layers-2` is maybe not necessary
- `layer-3` should be renamed to `layers`

danielbayley:
> * `area` is not easy to understand

Good enough for _[Things](https://github.com/lucide-icons/lucide/pull/1142#issue-1679447215)_…

> * `layers-2` is maybe not necessary

@jguddas Are _any_ variants strictly necessary…?

I for one intend to use it (preferable to the triple version)…

> * `layer-3` should be renamed to `layers`

Fine with me, but as mentioned in https://github.com/lucide-icons/lucide/pull/1142#discussion_r1224465931 I think @karsa-mistmere had objections…?

jguddas:
> > * `area` is not easy to understand
> 
> Good enough for _[Things](https://github.com/lucide-icons/lucide/pull/1142#issue-1679447215)_…

What about just using `box` box (1) for things?

> @jguddas Are _any_ variants strictly necessary…?
> 
> I for one intend to use it (preferable to the triple version)…

I'm 100% against all `-number` versions, but if we give it a descriptive name `layers-two-of-em` or whatever, than I'm fine with it :man_shrugging:

danielbayley:
> What about just using `box` for things?

It’s meant to represent `area` not things 😆 (and I modelled it on `box`)—I suppose because the line is missing, that area—I meant good enough for them (very well designed app!)

> I'm 100% against all `-number` versions, but if we give it a descriptive name

What about `layers-double` and `layers-triple`?

jguddas:
`layers-double` sounds good.

Let's hope we can just drop the `layers-3` icon and update `layers` instead.

> It’s meant to represent area not things laughing (and I modelled it on box)

Can you explain what an area is, I'm not familiar with `things` and can't really deduct the use case from that?

danielbayley:
> Can you explain what an area is

@jguddas Screenshot above https://github.com/lucide-icons/lucide/pull/1142#issue-1679447215 probably explains better than I can… But basically in that context, meaning `area` of responsibility (like a category) used to group projects (which in turn contain todo lists).

But also more literally—this PR being for geometry icons—`area` represented by the horizontal surface area of the cube (with attention drawn by the missing vertical line, present in `box`…)

icons/weight.svg
karsa-mistmere, line 1: This one's from my drafts, I feel like a larger handle works a bit better:
weight

icons/drafting-compass.svg
karsa-mistmere, line 1: Seriously though, this is _exact the same icon to a tee_ as the one in my drafts. Just... how? 😆 

I'd show you the difference, but there literally isn't any:
image

ericfennis: Lol

icons/weight.svg
ericfennis: Is there still room left to increase the body with 1px?

ericfennis (approved):
Really lovely design. Great work @danielbayley !

icons/weight.svg
karsa-mistmere: Sure, we could bump up the size:
weight-old weight

ericfennis: Yeah I agree, this looks better. Can you provide a code suggestion for this icon?

icons/weight.svg
```diff
@@ -0,0 +1,14 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="12" cy="5" r="2" />
+  <path d="M4.7 9C4.9 7.9 5.9 7 7 7h10c1.1 0 2.1.9 2.3 2l1.4 10c.2 1.1-.6 2-1.7 2H5c-1.1 0-1.9-.9-1.7-2Z" />
```
karsa-mistmere, line 13:
```suggestion
  <circle cx="12" cy="5" r="3" />
  <path d="M6.5 8a2 2 0 0 0-1.905 1.46L2.1 18.5A2 2 0 0 0 4 21h16a2 2 0 0 0 1.925-2.54L19.4 9.5A2 2 0 0 0 17.48 8Z" />
```

karsa-mistmere: @ericfennis: here goes.

icons/weight.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
```
karsa-mistmere, line 3:
```suggestion
  "contributors": [
    "danielbayley",
    "karsa-mistmere"
  ],
  "tags": [
```

icons/cone.svg
karsa-mistmere, line 1: Not totally sure if this should be rounded or not, since `cube` is also rounded. 🤔 
image

icons/pyramid.svg
karsa-mistmere, line 1: Potential rounded version, see discussion on `cone`:
image

icons/cone.svg
danielbayley: I don’t think rounding makes sense on these.

karsa-mistmere: Well, I don't think it necessarily makes sense on cube either. 😄

karsa-mistmere: Dunno, at 100% they're almost indiscernible from the non-rounded versions:
image

There's also the option to add a 1px rounding to the top only and 2px to the base:
image

karsa-mistmere: (My main concern here is obviously `pyramid` being consistent with `layers`).

ericfennis: I like the second design from @karsa-mistmere

jguddas: I like the 2px pyramid and 1px cone.

danielbayley: I think 2px rounding on the base of `pyramid` to match `layers` makes sense, but especially since we already have sharp corners on other icons like e.g https://lucide.dev/icons/mouse-pointer, and all of the `arrow-*` icons, I don’t think the tip should be… Certainly no more than by 1px anyway.

danielbayley: For better comparison…

preview

icons/pyramid.svg
danielbayley: For better comparison…

preview

icons/cone.svg
ericfennis: I think second design looks a bit better with the other icons as well.

danielbayley:
> I think second design looks a bit better with the other icons.

danielbayley:
> I think second design looks a bit better with the other icons.

@ericfennis Ok, implemented…

ericfennis (approved):

icons/land-plot.svg
```diff
@@ -0,0 +1,16 @@
...
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="m12 8 6-3-6-3v10" />
+  <path d="m8 11.8-5.2 3.1c-.5.3-.8.7-.8 1.1s.3.8.8 1.1l7.6 4.5a3.3 3.3 0 0 0 3 0l7.6-4.5c.5-.3.8-.7.8-1.1s-.3-.8-.8-1.1l-5-3.1" />
+  <path d="m17.4 12.7-11 6.5" />
+  <path d="m6.5 12.7 11 6.5" />
```
karsa-mistmere, line 15: Looks cute, but it's pretty off-grid:

```suggestion
  <path d="m12 8 6-3-6-3v10" />
  <path d="m8 11.99-5.5 3.14a1 1 0 0 0 0 1.74l8.5 4.86a2 2 0 0 0 2 0l8.5-4.86a1 1 0 0 0 0-1.74L16 12" />
  <path d="m6.49 12.85 11.02 6.3" />
  <path d="M17.51 12.85 6.5 19.15" />
```

icons/layers-2.svg
karsa-mistmere: The corners are a bit wonky:
image

```suggestion
  <path d="m16.02 12 5.48 3.13a1 1 0 0 1 0 1.74L13 21.74a2 2 0 0 1-2 0l-8.5-4.87a1 1 0 0 1 0-1.74L7.98 12" />
  <path d="M13 13.74a2 2 0 0 1-2 0L2.5 8.87a1 1 0 0 1 0-1.74L11 2.26a2 2 0 0 1 2 0l8.5 4.87a1 1 0 0 1 0 1.74Z" />
```

danielbayley:
@ericfennis @karsa-mistmere @jguddas What do you guys think of completing the base of some of these shapes?

preview

jguddas:
> What do you guys think of completing the base of some of these shapes?

I like it more without the full base.

icons/layers-3.svg
karsa-mistmere: Also has wonky corners:
image

```suggestion
  <path d="m12.83 2.18a2 2 0 0 0-1.66 0L2.6 6.08a1 1 0 0 0 0 1.83l8.58 3.91a2 2 0 0 0 1.66 0l8.58-3.9a1 1 0 0 0 0-1.83Z" />
  <path d="m6.08 9.5-3.5 1.6a1 1 0 0 0 0 1.81l8.6 3.91a2 2 0 0 0 1.65 0l8.58-3.9a1 1 0 0 0 0-1.83l-3.5-1.59" />
  <path d="m6.08 14.5-3.5 1.6a1 1 0 0 0 0 1.81l8.6 3.91a2 2 0 0 0 1.65 0l8.58-3.9a1 1 0 0 0 0-1.83l-3.5-1.59" />
```

icons/layers.svg
karsa-mistmere: Same as `layers-3`:
```suggestion
  <path d="m12.83 2.18a2 2 0 0 0-1.66 0L2.6 6.08a1 1 0 0 0 0 1.83l8.58 3.91a2 2 0 0 0 1.66 0l8.58-3.9a1 1 0 0 0 0-1.83Z" />
  <path d="m22 17.65-9.17 4.16a2 2 0 0 1-1.66 0L2 17.65" />
  <path d="m22 12.65-9.17 4.16a2 2 0 0 1-1.66 0L2 12.65" />
```

icons/pyramid.svg
```diff
@@ -0,0 +1,14 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M12.6 2.3a.75.75 0 0 0-1.2 0L2.2 15.4c-.1.2-.2.4-.2.6 0 .4.3.8.8 1.1l7.6 4.5a2.94 2.94 0 0 0 3.1 0l7.6-4.5a1.35 1.35 0 0 0 .7-1.7Z" />
+  <path d="M12 2v20" />
```
karsa-mistmere, line 13: Also has some funky corners:
```suggestion
  <path d="M2.5 16.88a1 1 0 0 1-.32-1.43l9-13.02a1 1 0 0 1 1.64 0l9 13.01a1 1 0 0 1-.32 1.44l-8.51 4.86a2 2 0 0 1-1.98 0Z" />
  <path d="M12 2v20" />
```

icons/radius.svg
```diff
@@ -0,0 +1,16 @@
...
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M20.3 17.5a10 10 0 1 0-2.8 2.8" />
+  <circle cx="19" cy="19" r="2" />
+  <path d="m13.4 13.4 4.2 4.2" />
+  <circle cx="12" cy="12" r="2" />
```
karsa-mistmere, line 15: Over-optimised:
```suggestion
  <path d="M20.34 17.52a10 10 0 1 0-2.82 2.82" />
  <circle cx="19" cy="19" r="2" />
  <path d="m13.41 13.41 4.18 4.18" />
  <circle cx="12" cy="12" r="2" />
```

icons/tangent.svg
```diff
@@ -0,0 +1,16 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="17" cy="4" r="2" />
+  <path d="M15.6 5.4 5.4 15.6" />
```
karsa-mistmere, line 13: Over-optimised:
```suggestion
  <path d="M15.59 5.41 5.41 15.59" />
```

icons/triangle.svg
```diff
@@ -9,5 +9,5 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="m21.73 18-8-14a2 2 0 0 0-3.48 0l-8 14A2 2 0 0 0 4 21h16a2 2 0 0 0 1.73-3Z" />
+  <path d="M13.75 4a2 2 0 0 0-3.5 0L2.2 18A2 2.1 0 0 0 4 21h16a2 2 0 0 0 1.75-3Z" />
```
karsa-mistmere, line 12: This is actually slightly worse in precision and curvature than before. 😅
```suggestion
  <path d="M13.73 4a2 2 0 0 0-3.46 0l-8 14A2 2 0 0 0 4 21h16a2 2 0 0 0 1.73-3Z" />
```

karsa-mistmere (changes requested):

karsa-mistmere:
> @ericfennis @karsa-mistmere @jguddas What do you guys think of completing the base of some of these shapes?
> 
> preview

I think some of these - especially pyramid - are too dense.

ericfennis:
I think the `area` icon is a bit abstract for the name. Also the usecase or example you posted is also shown as a application specific usecase. Not sure if this fits in Lucide.

`cylinder-volume` is a bit too specific I think. And it bothers me that it's not exactly the same shape as the `cylinder` icon. Also, It raises the question of why we don't have a `box-volume` or `pyramid-volume` icon. But that can be my perfectionism 😄 

But can we try to wrap up this PR?

PR #1151, opened 2023-04-24
Add `venn` diagram icon

danielbayley:
Compare/diff/overlap etc…


jguddas:
What about using 3 circles?

danielbayley:
> That could be misconstrued as a pagan symbol. 🤔😛

Oh _that_ again… 😅

> What about using 3 circles?

@jguddas Yeah I tried that, didn’t think it worked very well…

Untitled
Artboard 1@2x

karsa-mistmere:
This is a similar use case, I had these rectangular "pathfinder" type icons in my drafts, although I guess they could also work with circles and cover the Venn diagram use case as well?

image

danielbayley:
> I had these … type icons in my drafts

@karsa-mistmere Oh snap! Again. 🙈

I also had similar in my drafts as `intersect`, `exclude`, `subtract`, `join`/`union`, `align-snap` etc…

Untitled

Artboard 1
<br/>

I think `venn` is a separate thing though, more related to the `*-chart` icons…

ericfennis:
This one feels a bit like a copy with the just merged `blend` icon. I will merge this one first to the lab branch

PR #1153, opened 2023-04-24
Add `radiation`/`biohazard` icons

danielbayley:
Alternatives:
Untitled


karsa-mistmere:
How about something like the 2nd row but with no rounding?
image

While in general, rounding fits the Lucide aesthetic more, I think in this instance it causes the well established shape to lose character.

jguddas:
I don't like the sharp corners and the big circle that doesn't have a 2px whole in it in the center.

jguddas:
I think this one is the best, but maybe a bit smaller with a less pronounced radius? image

karsa-mistmere:
> I think this one is the best, but maybe a bit smaller with a less pronounced radius? image

That's not bad, but yeah, maybe let's try a somewhat smaller corner radius.

danielbayley:
> larger gap between the segments and the central dot.

Yeah that was a rules violation. What do we think between these…?

Untitled

Untitled

jguddas:
I prefer the right one.

danielbayley:
> I prefer the right one.

Same…

karsa-mistmere:
I would love to also have a matching biohazard icon, but this is just not compliant, I'm afraid. 😅 
biohazard

jguddas:
Looks nice and compliant to me. icon (27)

PR #1160, opened 2023-04-26
Add `ligature` icon

danielbayley:
Or:

Untitled

Untitled


karsa-mistmere:
I think at 100% this looks too much like an uppercase letter A.

I'd make the `i` taller (same stem height in fact as our lower case `x` in super- and subscript), and push them closer apart, e.g.
ligature

karsa-mistmere:
Hmm... not sure about the sans serif, I think it makes the `f` look a bit out of place, and it's probably because while it's quite common for serif `f`s to have a hook that's significantly wider than the bar with sans serif fonts this is a much more unusual occurrence.

danielbayley:
> Hmm... not sure about the sans serif

Oops! My mistake… it wasn’t meant to be sans!

karsa-mistmere (approved):

ericfennis (approved):
Really clean, love it!

PR #1161, opened 2023-04-26
Add `ethereum` icon

danielbayley:
Can’t have BTC https://lucide.dev/icons?search=bitcoin without ETH… That’s just not on :)

Alts:

Untitled

Untitled

### DIFF
--- a/icons/ethereum.json
+++ b/icons/ethereum.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "currency",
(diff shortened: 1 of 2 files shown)


danielbayley:
> I'm afraid both are brand icons

@karsa-mistmere Oh, I didn’t think they’d fall under that (they certainly aren’t companies)…

Do they not count as https://lucide.dev/icons?search=currency, same as `£` `$` `€` etc…? I know which I’d rather hold! But that’s beside the point…

PR #1163, opened 2023-04-26
Add `eye-closed` icon

danielbayley:
Also [wal]nut/leaf…

### DIFF
--- a/icons/eye-closed.json
+++ b/icons/eye-closed.json
@@ -1,0 +1,25 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "view",
+    "watch",
+    "hide",
+    "hidden",
+    "nut",
+    "walnut",
+    "food",
+    "diet",
+    "leaf",
+    "autumn",
+    "fall",
+    "plant"
+  ],
+  "categories": [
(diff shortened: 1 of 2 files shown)


karsa-mistmere:
I don't find this particularly recognisable, the alternate use cases are also a bit dubious, I'd vote against merging this as it currently is.

ericfennis:
I agree with @karsa-mistmere, also the use case is missing.

danielbayley:
> the use case is missing.

@karsa-mistmere @ericfennis Sorry, too many PRs! 

Use case is situations like these being used as e.g watch/unwatch as a simple toggle, where this pair of icons are overlayed on top of each other…

Untitled

jguddas:
Another use case is for password fields and such.

What about adding a bit of curvature to the eyelid?

danielbayley:
> Another use case is for password fields and such.

Yeah, another good example!

> What about adding a bit of curvature to the eyelid?

I can’t `push`, since this is now closed (a bit of a silly limitation of GitHub I think…) and don’t really want to open a new PR for the same thing @ericfennis…

Untitled

Untitled

PR #1165, opened 2023-04-27
Add `scatter-chart` icon

danielbayley:


ericfennis:
Can you try a design with bigger dots?

danielbayley:
> Can you try a design with bigger dots?

`push`ed…

ericfennis:
this looks better!

PR #1169, opened 2023-04-27
[Packages][Lucide] Switch to `data-lucide`

### LINKED ISSUE #981: support HTML standard data-* attributes
Is it possible to add support for data-* attributes so that users who prefer to write standards-compliant code can use Lucide without having to add exceptions to code quality checking tools?

For example, instead of `<span icon-name='box'></span>` it should be possible to write `<span data-icon-name='box'></span>`.

The two options can co-exist if needed to maintain backward compatibility.

https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

karsa-mistmere:
closes #981


packages/lucide/src/lucide.ts
```diff
@@ -5,7 +5,7 @@ import * as allIcons from './icons';
  * Replaces all elements with matching nameAttr with the defined icons
  * @param {{ icons?: object, nameAttr?: string, attrs?: object }} options
  */
-const createIcons = ({ icons = {}, nameAttr = 'icon-name', attrs = {} } = {}) => {
```
ericfennis, line 8: We maybe should consider that we have an overlapping time, with changing this.
So maybe we should support the old attribute as well and console a warning, with saying we remove this in v1?

karsa-mistmere: Yeah, good idea, I've posted an implementation of this, also checking if the default attribute name was used.

packages/lucide/src/lucide.ts
```diff
@@ -17,10 +17,20 @@ const createIcons = ({ icons = {}, nameAttr = 'icon-name', attrs = {} } = {}) =>
...
-
   Array.from(elementsToReplace).forEach((element) =>
     replaceElement(element, { nameAttr, icons, attrs }),
   );
+
+  /** @todo: remove this block in v1.0 */
+  if (nameAttr === 'data-lucide') {
+    const deprecatedElements = document.querySelectorAll('[icon-name]');
+    if (deprecatedElements.length > 0) {
+      console.warn('Icons were found with the now deprecated icon-name attribute. These will still be replaced for backwards compatibility, but will no longer be supported in v1.0');
```
ericfennis, line 28: Maybe mention our package name somewhere. So devs know where this warning is coming from.

karsa-mistmere: Oh, sure, also, we should add the new attribute name anyway.

karsa-mistmere:
```suggestion
      console.warn('[Lucide] Some icons were found with the now deprecated icon-name attribute. These will still be replaced for backwards compatibility, but will no longer be supported in v1.0 and you should switch to data-lucide');
```

ericfennis: perfecto!

PR #1177, opened 2023-04-29
Fix children handling in lucide-react-native

quentin-sommer:
The react-native lib only accept several children on an icon, this PR makes it support single children. I also added regression tests.

The react lib was already [doing it properly](https://github.com/lucide-icons/lucide/blob/477f2b2affa629e3c713f9f752dea4fec4ef8a86/packages/lucide-react/src/createLucideIcon.ts#L40) and preact seems fine too using `toChildrenArray`


packages/lucide-react-native/src/createLucideIcon.ts
```diff
@@ -30,7 +30,9 @@ const createLucideIcon = (iconName: string, iconNode: IconNode) => {
             const upperCasedTag = (tag.charAt(0).toUpperCase() + tag.slice(1)) as keyof (typeof NativeSvg);
             return createElement(NativeSvg[upperCasedTag] as FunctionComponent<Record<string, string>>, attrs);
           }),
-          ...(children as ReactNode[] || []),
```
quentin-sommer, line 33: I dropped this type because children already has a type with `LucideProps`

ericfennis (approved):
@quentin-sommer Thanks for fixing this!

PR #1179, opened 2023-04-30
Add `square-asterisk` icon

danielbayley:
`key-asterisk` from [octicons](https://primer.style/design/foundations/icons)… Also has more general possible use as password/entry etc…


ericfennis (approved):
LGTM!

PR #1156, opened 2023-04-25
Optimise/add remaining `panel-*`/consolidate `layout` icons

danielbayley:
Following up from https://github.com/lucide-icons/lucide/pull/1003#issuecomment-1521387131 …

> For example `panel-top-left` is currently conflicting with layout icons.

I propose to rename `layout` as `panels-top-left` (leaving `layout` as an alias), to make it part of the `panel-*` set in #1003, and this PR… Which is meant to cover `layout-panel-*` [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html) for use in an upcoming VS Code _Lucide_ product and file icon theme…

`panels-top-left` (currently https://lucide.dev/icons?search=layout) is also `project-*` from [octicons](https://primer.style/design/foundations/icons), obviously as you can see above, right here on GitHub… So relevant `tags` also added, and `panels-top-left-dashed` as `project-template` from the same set.

Related to #1059, #1181, #1201, #1498 and #1500.

---

Edit: As per https://github.com/lucide-icons/lucide/pull/1156#issuecomment-1696013745, leaving these here for reference…

preview

PR #1003, opened 2023-03-30
Add `panel-*` icons

danielbayley:
This PR aims to complete the existing `sidebar-*` set of icons, cover the `layout`[`-panel`]`-*` [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html) and consolidate everything into a more coherent subset of icons. Closes #1059.

PR #1180, opened 2023-04-30
feat: added 100% preview with different stroke widths

jguddas:

PR #1184, opened 2023-04-30
Add/optimise `layout-*` alternate icons

danielbayley:
`layout-grid-2` is `layout` from [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html). These match existing https://lucide.dev/icons?search=layout- icons…

Also fixed `layout-template` to also use `rect`s, and rounded all the corners to 2px radius.

Semi-related to #1156 and #1003…


jguddas:
IMO we should add a radius to all of them network (3) -> network (2)

danielbayley:
> IMO we should add a radius to all the layout icons

Yeah already on it 😅

jguddas (approved):

karsa-mistmere:
As with other similarly themed recent PRs, I'm very unsure about taking this direction.
`grid` looks pretty decent, but anything with smaller rectangles than that is much less legible than it currently is.

karsa-mistmere:
As with #1188, we could maybe check things out with 0, 1px and 2px corners side by side.

jguddas:
I like those a lot better, when scrolling through the icons this just felt out of place.

danielbayley:
> As with #1188, we could maybe check things out with 0, 1px and 2px corners side by side.

Yep, and we should get #1180 merged as a priority, so we can automate more of this review process efficiently…

ericfennis:
I think the 1px rounding looks way better for this

jguddas:
What we should also check is if they feel consistent in combination with other icons, not just on their own.

danielbayley:
> I think the 1px rounding looks way better for this

`push`ed…

ericfennis (approved):

karsa-mistmere:
> What we should also check is if they feel consistent in combination with other icons, not just on their own.

Definitely! Here goes:

image

In my opinion:
* Row 1 (current): yeah, definitely not very consistent
* Row 2 (PR): Not _perfectly_ consistent, but definitely lightyears better than row 1.
* Row 3 (2px): Way too blobby.

---

@danielbayley: One final thing I'm not psyched about are the names.
I just don't find the names `layout-grid-2` and `layout-template-2` very fitting.

Maybe something like `layout-panel-left`, `layout-panel-top`? In either case, something that better describes the image at hand.

jguddas:
> Row 3 (2px): Way too blobby.

We should try to not just look at Icons in isolation, I personally don't mind the blobbyness and am more concerned about cohesion.

karsa-mistmere:
> We should try to not just look at Icons in isolation, I personally don't mind the blobbyness and am more concerned about cohesion.

I have done just that in the review apps, yes and I still find the 2px versions way too blobby. 🙂 
* **1px** https://lucide-cf2cmvhp2-lucide.vercel.app/
* **2px** https://lucide-r7j06odgz-lucide.vercel.app/

The 2px version isn't _that_ bad per se, it's just not my cup of tea.

danielbayley:
> Maybe something like `layout-panel-left`, `layout-panel-top`? In either case, something that better describes the image at hand.

@karsa-mistmere Done.

karsa-mistmere (approved):

PR #1187, opened 2023-04-30
Replace faces

jguddas:
I'm not good at this…


ericfennis:
Not sure about this change. From 100% the faces are not that recognizable.

ericfennis:
Sorry close this one see my previous comment, please if you want to improve/iterate on this please do this in an issue.

PR #1188, opened 2023-04-30
Changed `binary`

jguddas:
Felt to aggressive.


karsa-mistmere:
I find the rounding too aggressive. 😅

Can we see versions with 0, 1 and 2px rounded corners side by side for comparison?

jguddas:
icon (33) icon (32)

The 2px variant is also used here for arrow-down-0-1 arrow-down-0-1.

danielbayley:
> The 2px variant is also used here for arrow-down-0-1

Yeah I thought this would be a no brainer, for consistency…

karsa-mistmere:
I think consistency here might not be necessarily a must — `binary` has a much more monospace feel to convey that it's a "code", while numerical sorting isn't necessarily binary.

We could again check different roundings against each other to see if maybe one version works well for both use cases.

ericfennis:
I think people won't notice the difference between squared and rounded zeros if you look on 100%. For me, both are looking good, but it's good we keep consistency.

karsa-mistmere:
We also have `file-digit` using these, here's the comparison I meant:
image

I feel like the numerical sorting use case is a bit different, but a compromise for consistency would be using row #2 for all of them.

In general, for `binary` and `file-digit` row #2 looks the best to me, while for `arrow-down-0-1` it's row #3.

ericfennis:
Hmm ok, to me the third row looks best overall and matches the best with the lucide style on 100% scale.
But I get you binary maybe need to be looking a bit blocky looking.

However looking at some binary art, it's not always blocky 😛 
image

karsa-mistmere:
Yeah, of course, I'm also fine with row 3. :)

@jguddas: can you please also update `file-digit` to match?
(And possibly add the binary tag to that :D ).

ericfennis (approved):

PR #1190, opened 2023-04-30
Added round corners to `network`

jguddas:


karsa-mistmere:
Also not in favour of this change, I realise that at first glance these corners seem to conflict with the guidelines, but 2px brings these just bit too close to circles I think.

jguddas:
What about 1px https://github.com/lucide-icons/lucide/pull/1190/commits/59b95d60f556addb0f4c1d4fe5083c98a081a8de ?

ericfennis:
1px is better I think, Is nice if we make this consistent with #1184

ericfennis (changes requested):
Change to 1px corners?

ericfennis (approved):

PR #1203, opened 2023-05-02
Add `archive-x` (junk) icon

danielbayley:


jguddas:
Wouldn't trash-x based on the trash icon make more sens?

karsa-mistmere (approved):

jguddas:
The x is to small IMO

danielbayley:
> The x is to small IMO

@jguddas @karsa-mistmere These could stand to be slightly bigger anyway…

Untitled

karsa-mistmere:
> > The x is to small IMO
> 
> @jguddas @karsa-mistmere These could stand to be slightly bigger anyway…
> 
> Untitled

I'm not a big fan of making them off-centre though. 🤔

jguddas:
> This seems slightly taller than is ideal though.

I like the tall ones, when you do a google search for archive box you also see that they are commony quite tall.

danielbayley:
> I like the tall ones

I don’t think look quite as good, but `push`ed anyway…

karsa-mistmere:
I thought I would hate it 1px taller, but it's actually not bad at all, I think it actually even looks more consistent with the rest of the icons. One thing I would maybe also change is have the lid use 1px rounding to be more consistent with other smaller sized rectangles:
image
image

LMKWYT.

jguddas:
Looks great, don't care for the border radius either way.

ericfennis (approved):
I think this not a bad change. Looks good!

icons/archive-x.svg
```diff
@@ -0,0 +1,16 @@
...
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect width="20" height="5" x="2" y="3" rx="2" />
```
karsa-mistmere, line 12:
```suggestion
  <rect width="20" height="5" x="2" y="3" rx="1" />
```

icons/archive.svg
```diff
@@ -9,7 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect width="20" height="5" x="2" y="4" rx="2" />
-  <path d="M4 9v9a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V9" />
-  <path d="M10 13h4" />
+  <rect width="20" height="5" x="2" y="3" rx="2" />
```
karsa-mistmere, line 12:
```suggestion
  <rect width="20" height="5" x="2" y="3" rx="1" />
```

karsa-mistmere (changes requested):

karsa-mistmere (approved):

PR #1208, opened 2023-05-03
Refine/add `message`[`-*`] icons

danielbayley:
The existing `message-circle` is a bit small, for no obvious reason…

`message-square-warning` and `message-square-diff` are `report` and `request-changes` from [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html), respectively. The former is also part of [octicons](https://primer.style/design/foundations/icons).

Some other options… not sure what is desirable to include, so leaving them here:

Untitled

Update: Added `message-square-share` as part of the set ([`cross-reference`](https://primer.style/design/foundations/icons/cross-reference-16) from octicons, as seen above right here on GitHub…)

---

Edit: `message-bubble-*` variant (potential replacement for existing `message-circle-*` variants)…

preview


ericfennis (approved):
Nice LGTM!

icons/message-circle-question.svg
karsa-mistmere, line 1: The question mark is off the grid.

icons/message-square-diff.svg
karsa-mistmere, line 1: Not sure about this one, it doesn't look like a diff sign, more like the bottom half of the bubble is dashed for some reason. Can you try an alternate design here?

icons/message-square-warning.svg
karsa-mistmere, line 1: This one looks out of proportion, if we can't fit a taller exclamation mark somehow (e.g. file-warning) let's just drop it.

karsa-mistmere (changes requested):

jguddas:
How does it look compared to other circular icons?

icons/message-circle-question.svg
danielbayley: Ok, uses the existing `?` now. Although I think all the question marks might need looking at in a separate PR…

icons/message-square-diff.svg
danielbayley: Oops, meant to use the existing `path` from `file-diff`, this was a bit shorter for some reason…

icons/message-square-warning.svg
danielbayley: That seems drastic; I think it looks recognisable enough as what it is… But anyway, possible alternatives:

Untitled

Untitled

jguddas:
What about giving the first row a try, than we have more cohesion with the other circular icons?
 image

danielbayley:
> How does it look compared to other circular icons?

@jguddas Isn’t that what the new [Preview cohesion](https://github.com/lucide-icons/lucide/pull/1208#issuecomment-1532775971) is for…? :) Nice improvement by the way! Is it selecting icons at random to compare? (I haven’t looked over the code…) If not, I think `circle`[`-`something] should be in there, obviously based on this…

Anyway, the message `circle` is larger than it was before, but 1px less radius than `circle` to accommodate the pointer, bottom left…

jguddas:
> > How does it look compared to other circular icons?
> 
> @jguddas Isn’t that what the new [Preview cohesion](https://github.com/lucide-icons/lucide/pull/1208#issuecomment-1532775971) is for…? :) Nice improvement by the way! Is it selecting icons at random to compare? (I haven’t looked over the code…) If not, I think `circle`[`-`something] should be in there, obviously based on this…
> 
> Anyway, the message `circle` is larger than it was before, but 1px less radius than `circle` to accommodate the pointer, bottom left…

It still feels a bit small to me.

The cohesion preview has square icons in the first row and random icons in the last row.

danielbayley:
> What about giving the first row a try

Untitled

Untitled

> It still feels a bit small to me.

Well bigger than they were… I just think the pointer gets a bit too drowned out by the full 20x20 circle…

> And we should really take a look at the question mark for all icons where it is used.

Yeah, as mentioned https://github.com/lucide-icons/lucide/pull/1208#discussion_r1186709729, I think that’s for a separate PR though…

jguddas:
@karsa-mistmere you are the expert here, I prefer the bigger one, what do you think?

ericfennis:
@danielbayley Can you rebase the branch?

danielbayley:
> @danielbayley Can you rebase the branch?

Done…

ericfennis (approved):

danielbayley:
I also have these, which I think are nicer than the current `circle-message-*` variants…

preview

What do you guys think? @ericfennis @karsa-mistmere @jguddas

jguddas:
> I also have these, which I think are nicer than the current circle-message-* variants…

Yeah, those are awesome, could be the default so `message`, and the circle variant is `message-circle`.

danielbayley:
> Yeah, those are awesome, could be the default so `message`, and the circle variant is `message-circle`.

3 variants seems like overkill though… I would vote these as default `message` and `massage-square`, dropping the current `message-circle`s…

What do you guys think of that?

karsa-mistmere:
> What do you guys think? @ericfennis @karsa-mistmere @jguddas

To be honest, I very much prefer the existing ones - I think those two (message-square and message-circle) are pretty well-established shapes, and while this one does look nice it is much less universal and I think the reduced dimensions will eventually prove to be just a bit too tight (e.g. the back/forward variants are already pretty dense as compared to what would be ideal).

ericfennis:
I also prefer the existing ones. So I agree with @karsa-mistmere on this.

danielbayley:
> reduced dimensions will eventually prove to be just a bit too tight 

@karsa-mistmere There is actually _more_ space, horizontally… For example, I can’t fit the nice `voicemail`/note variant into `message-square`, or even the current iteration of `message-circle` (see also https://github.com/lucide-icons/lucide/pull/1208#issuecomment-1537171887.

What do we think of @jguddas suggestion to add them simply as `message`?

Seems a shame to waste them…

karsa-mistmere:
Sidenote: I did think of maybe "fixing" `message-square`'s visual weight and centre of gravity instead, but I'm not convinced of this either, the previous shape just feels better.
image
image

danielbayley:
> Sidenote: I did think of maybe "fixing" `message-square`'s visual weight and centre of gravity instead, but I'm not convinced of this either, the previous shape just feels better.

@karsa-mistmere Maybe slightly—they are nice—but also feels like it technically ought to be the same width as https://lucide.dev/icons/rectangle-horizontal. These don’t look bad… and the extra space does afford `message-square-voicemail`, which would be a nice addition. See what you guys think… cc: @jguddas @ericfennis 

preview

ericfennis (changes requested):
Nice I think this one is almost good to go.
I've doubts about use-cases for `message-dot`.
And `message-square-text-quote` is a bit hard to see on 100%. Probably `message-square-quote` is better suited for this usecase.

Can you remove `message-square-text-quote`? And provide a use-case for `message-dot`? I do not approve `active` states to use dots.

ericfennis (approved):

PR #1211, opened 2023-05-03
Optimise/add `scroll`[`-text`] (script) icon

danielbayley:
Use case is fairly self explanatory…


karsa-mistmere (approved):
Nice one, yeah, I definitely didn't know as much about SVG code back then as I do now. 😄

ericfennis (approved):
Awesome work, LGTM!

PR #1214, opened 2023-05-04
Adds `ferris-wheel` and `roller-coaster`

karsa-mistmere:
## Icon Request

* Icon name: `roller-coaster` & `ferris-wheel`
* Use case: to represent amusement parks, funfairs, tourist attractions in map / POI applications.
image

### DIFF
--- a/icons/ferris-wheel.json
+++ b/icons/ferris-wheel.json
@@ -1,0 +1,16 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "big wheel",
+    "daisy wheel",
+    "observation",
+    "attraction",
+    "entertainment",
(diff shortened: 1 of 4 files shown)


jguddas:
The straight lines don't feel like gondola to me, I also think we should make the radius 10px so it fits with the circle.

Here somethings I tried but am not super happy with either:

icon (40) circle (3)icon (41)

jguddas:
What about closing the gap so it is 2px? icon (43) → icon (42)

karsa-mistmere:
> What about closing the gap so it is 2px? icon (43) → icon (42)

Nice catch!

ericfennis (approved):
nice love it!

PR #1215, opened 2023-05-04
Add generic `folder-root`/`sync` icons

danielbayley:
Follow up to #1009 and #1128. See also: https://github.com/lucide-icons/lucide/pull/1128#issuecomment-1516560790 and https://github.com/lucide-icons/lucide/pull/1128#issuecomment-1518155990 for context…

> I guess folder-dot could work and still have generic usages

Current options for `folder-dot`:

Untitled

Untitled

I suppose middle row, right could be construed as kind of `git` specific, but also maybe interpreted as 'root folder' in general… I like it better than https://lucide.dev/icon/folder-git?search=folder-git even.

Thoughts on these?


ericfennis (approved):
LGTM!

danielbayley:
> I wish `folder-sync` matched `database-backup` and `list-restart` more

@karsa-mistmere See latest `push`…

PR #1216, opened 2023-05-04
Add `spell-check` icons

### LINKED ISSUE #928: Add `spell-check` icon
## Icon Request

* Icon name: spell-check
* Use case: to enable spell checking
* _Screenshots_ of similar icons: spell-check-2 spell-check

danielbayley:
See #928… with credit to @jguddas

Explored:

Untitled

Untitled

Closes #928.

### DIFF
--- a/icons/spell-check.svg
+++ b/icons/spell-check.svg
@@ -1,0 +1,15 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="m6 16 6-12 6 12" />
+  <path d="M8 12h8" />
+  <path d="m16 20 2 2 4-4" />
+</svg>
--- a/icons/spell-check-2.json
+++ b/icons/spell-check-2.json
@@ -1,0 +1,17 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "spelling",
+    "error",
+    "mistake",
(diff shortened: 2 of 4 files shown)


ericfennis:
Maybe we can try the wave from the waves icon?

karsa-mistmere:
> Maybe we can try the wave from the waves icon?

Yeah, I don't like this wavy line very much, I don't think it should be vertically symmetric, although the wave from `waves` doesn't work well either:
image
(1st alternative has a gap that's too little, 2nd alternative just doesn't look very nice or consistent with `baseline`).

Maybe something like the 3rd one here?
image
(`baseline` for comparison)

One addendum from my side: I think two spell check icons should be plenty (that's still one more than what's absolutely necessary :D ), let's keep just two out of the three, preferably the two with the capital letter As.

icons/spell-check.svg
```diff
@@ -0,0 +1,15 @@
...
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="m6 16 6-12 6 12" />
+  <path d="M8 12h8" />
+  <path d="m16 20 2 2 4-4" />
+</svg>
```
jguddas, line 15: If you want to give credit to someone, you can add them as a co-author.

`Co-authored-by: Jakob Guddas <github@jguddas.de`

https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

danielbayley: Will do!

danielbayley:
@karsa-mistmere `push`ed…

ericfennis (approved):

PR #1218, opened 2023-05-05
feat: update contributors on release

jguddas:
- Takes ~12 seconds for all icons.
- Gets all contributors, including co-authors.
- 0 caching, fast by default.
- Includes lost contributors from feather fork.
- Preserves contributors when renaming / copying in the future.

Possible Optimizations:
- Run only for changed icons.

Cons:
- Modifies the metadata file.
- `--follow` did not work with the `simple-git` log.

Issues:
- I can't test the release workflow.


scripts/updateContributors.mjs
```diff
@@ -0,0 +1,79 @@
+import simpleGit from 'simple-git';
+import { Octokit } from '@octokit/rest';
+import fs from 'node:fs';
+import path from 'path';
+import pMemoize from 'p-memoize';
```
karsa-mistmere, line 5: There are some dependencies missing from package.json.

jguddas: Oups

scripts/updateContributors.mjs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// get all contributors to be able to associate them when they are co authors
+await getContributors('icons');
+
+await Promise.all(
+  files.map(async (file) => {
+    const contributors = [
+      ...(LOST_CONTRIBUTORS[file.slice(6, -4)] || []),
+      ...(await getContributors(file, true)),
```
karsa-mistmere, line 70: We could also add the contributors for icon aliases if any:
```suggestion

const getAliasContributors = async (file) => {
  const aliases = (await JSON.parse(fs.readFileSync(file.replace(/.svg$/, '.json')))).aliases ?? [];
  const aliasContributors = [];
  for (const alias of aliases) {
    aliasContributors.push(...(await getContributors(`icons/${alias}.svg`)));
  }
  return aliasContributors;
}

const files = process.argv
  .slice(2)
  .map((file) => file.replace('.json', '.svg'))
  .filter((file, idx, arr) => arr.indexOf(file) === idx);

if (!files.every((file) => file.startsWith('icons/'))) {
  console.error("file path must start with 'icons/'");
  process.exit(1);
}

// get all contributors to be able to associate them when they are co authors
await getContributors('icons');

await Promise.all(
  files.map(async (file) => {
    const contributors = [
      ...(LOST_CONTRIBUTORS[file.slice(6, -4)] || []),
      ...(await getAliasContributors(file)),
      ...(await getContributors(file, true)),
```

scripts/updateContributors.mjs
```diff
@@ -0,0 +1,79 @@
...
+      repo: 'lucide',
+      ref: commit.hash,
+    });
+    return fetchedCommit?.author?.login;
+  },
+  { cacheKey: ([commit]) => commit.author_email, cache }
+);
+
+const getContributors = async (file, includeCoAuthors) => {
+  const { all: commits } = await simpleGit().log(['--reverse', file]);
```
karsa-mistmere, line 27: Failsafe to make sure that the file path is interpreted as a file and not a branch or tag:
```suggestion
  const { all: commits } = await simpleGit().log(['--reverse', '--', file]);
```

scripts/updateContributors.mjs
```diff
@@ -0,0 +1,79 @@
+import simpleGit from 'simple-git';
+import { Octokit } from '@octokit/rest';
+import fs from 'node:fs';
+import path from 'path';
+import pMemoize from 'p-memoize';
+import LOST_CONTRIBUTORS from './lost-contributors.json' assert { type: 'json' };
+
+const IGNORED_COMMITS = ['433bbae4f1d4abb50a26306d6679a38ace5c8b78'];
```
karsa-mistmere, line 8: We should probably extract this to a separate file.

scripts/updateContributors.mjs
```diff
@@ -0,0 +1,79 @@
+import simpleGit from 'simple-git';
+import { Octokit } from '@octokit/rest';
+import fs from 'node:fs';
+import path from 'path';
+import pMemoize from 'p-memoize';
+import LOST_CONTRIBUTORS from './lost-contributors.json' assert { type: 'json' };
```
karsa-mistmere, line 6: This is lost as in?... Not carried over from Feather?
If so, maybe let's call this `feather-contributors.json` or `original-contributors.json`.

karsa-mistmere (changes requested):
I've looked through the solution and it seems pretty neat, the only thing I'm missing is the ability to manually add contributors, we should definitely be able to do that.

scripts/updateContributors.mjs
jguddas: Is this only for moved files?
If we move them correctly, we could use `git log --follow` to include all commits of the file.

scripts/updateContributors.mjs
jguddas: Mostly for feather, but also if you mess up the history in some other way.

scripts/updateContributors.mjs
```diff
@@ -0,0 +1,79 @@
...
+
+const IGNORED_COMMITS = ['433bbae4f1d4abb50a26306d6679a38ace5c8b78'];
+
+const octokit = new Octokit({ auth: process.env.GITHUB_TOKEN });
+
+const cache = new Map();
+
+const getUserName = pMemoize(
+  async (commit) => {
+    const { data: fetchedCommit } = await octokit.repos.getCommit({
```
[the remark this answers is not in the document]
jguddas: It does 1 request per contributor, that's still quite a few requests so you might run int rate limiting issues at some point.

If you get issues, you can use `GITHUB_TOKEN="…" pnpm generate:contributors`.

scripts/updateContributors.mjs
jguddas: If it's just a couple, I don't think it's bad.

jguddas:
> I've looked through the solution and it seems pretty neat, the only thing I'm missing is the ability to manually add contributors, we should definitely be able to do that without the workflow overriding everything we do.

You can manually add contributors by adding them to the `lost-contributors.json` file.

jguddas:
Another thing we need to think about is what happens if people add new icons by copying existing icons.

scripts/updateContributors.mjs
karsa-mistmere: Yes, not that bad, but we could extend it automatically based on PR tags for example (marking certain PRs as "non-contributions").

ericfennis (commented):
Nice work, this looks promising!
Is it possible to add all the current contributors to all the icon files with this PR. This way we can also check if the script is working as expected.

icons/cog.json
```diff
@@ -10,5 +10,8 @@
   ],
   "categories": [
     "account"
+  ],
+  "contributors": [
```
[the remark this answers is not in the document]
jguddas: I only added the contributors that are not tracked in git, I can add them by running `pnpm generate:contributors` if you like.

What I don't know how to test is the release workflow.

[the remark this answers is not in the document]
karsa-mistmere: I've added all contributors, I think we definitely need to list everyone, even if it's "redundant" in a way so that people editing icons are not confused.

You can easily test the release workflow, it's just a bit tricky, you have to also update the release script in your own fork (and of course manually create a new release in your fork):
https://github.com/lucide-icons/lucide/blob/main/.github/workflows/release.yml#L31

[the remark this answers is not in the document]
karsa-mistmere: I've moved the target branch to `docs-vitepress`, I think it would make sense for this to go online when the new site does, what do you think @ericfennis & @jguddas?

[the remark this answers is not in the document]
ericfennis: @karsa-mistmere I think It's better to target the main branch, then we can already merge it and test it with the release flow.

scripts/updateContributors.mjs
```diff
@@ -0,0 +1,76 @@
...
+await Promise.all(
+  files.map(async (file) => {
+    const jsonFile = path.join(process.cwd(), file.replace('.svg', '.json'));
+    const json = JSON.parse(fs.readFileSync(jsonFile));
+    const contributors = [
+      ...(json.contributors || []),
+      ...(await getContributors(file, true)),
+    ].filter((contributor, idx, arr) => contributor && arr.indexOf(contributor) === idx);
+
+    fs.writeFileSync(jsonFile, JSON.stringify({ ...json, contributors }, null, 2));
```
ericfennis, line 74: Maybe it's better if we change the order of the object properties to:
```json
{
  "$schema": "..."
  "contributors": [],
  "tags": [],
  "categories": []
}
```
This way I think we prevent the issue with a possible merge conflict.

ericfennis (commented):
I think this is almost good to go. I really need this to complete the new website.
But I've a suggestion for the schema of the json files.

icons/cog.json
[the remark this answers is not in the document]
karsa-mistmere: OK, that also makes sense.

jguddas:
> I think this is almost good to go. I really need this to complete the new website. But I've a suggestion for the schema of the json files.
> 
> Jakob Guddas Can you also rebase this branch?

Absolutely swamped right now, but you should have access rights to my branch here.

scripts/updateContributors.mjs
karsa-mistmere: @ericfennis: I've rebased and added this reordering (as well as updated contributors to be up to date with latest changes as well as the reordering).

karsa-mistmere: And again. 😅

icons/accessibility.json
```diff
@@ -1,5 +1,6 @@
 {
   "$schema": "../icon.schema.json",
+  "contributors": [],
```
jguddas, line 3: This looks wrong.

ericfennis: hmm

karsa-mistmere:
I've also noticed that some commits we had were made with an incorrect user config (non-standard e-mail addresses etc), I've made a map of these and added them manually, except for commits that belonged to @delnyn, who is no longer on Github. :(

ericfennis:
@karsa-mistmere @jguddas I think we can merge this. But the workflow probably won't work. I tried to look into this but we have some problems with pushing from Github actions to protected branches.

Our current github actioins is failing at this step. I tried to find a solution to this. I came up with this thread: https://github.com/orgs/community/discussions/25305. Looks like it's only possible by using Personal Access Tokens. But that comes with some security risks.

karsa-mistmere:
Hi @jguddas,

Thanks for your yet again amazing work on this.

@ericfennis and I have talked in detail about what to do with this PR - especially in light of the GitHub workflow being unable to run on our protected main branch - and we've settled on adding these contributors but removing the workflow, since:

1) we haven't yet been able to find a sufficiently stable and secure way to enable running it that we're both happy with (in fact not even one that either of us are truly happy with)

2) both of us have expressed doubts about whether adding contributors automatically is even the way to go: giving credit for designing an icon is one thing, but we're not sure credit for adding tags or categories, or even doing trivial optimization on icons should be automatically warranted.

So after having settled on these, we've decided to merge this PR as is, again, thanks for your work, and one more thing: we would've liked to discuss this with you in detail in real time, but were unable to contact you, we would very much appreciate it if you contacted us via our official Lucide Discord channel so we could better keep in touch.

PR #1222, opened 2023-05-06
Add `font-size` icons

it-is-not:
image

Similar to what Microsoft Word does:
image

I haven't made any icons in a while so hopefully everything works properly with the new categories.

### DIFF
--- a/icons/a-arrow-down.json
+++ b/icons/a-arrow-down.json
@@ -1,0 +1,20 @@
+{
+  "$schema": "../icon.schema.json",
+  "contributors": [
+    "it-is-not",
+    "jguddas",
+    "danielbayley",
+    "ericfennis"
+  ],
+  "tags": [
+    "letter",
+    "font size",
+    "text",
+    "formatting",
+    "smaller"
+  ],
+  "categories": [
+    "text",
+    "design"
+  ]
+}
--- a/icons/a-arrow-down.svg
+++ b/icons/a-arrow-down.svg
@@ -1,0 +1,16 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M3.5 13h6" />
+  <path d="m2 16 4.5-9 4.5 9" />
+  <path d="M18 7v9" />
+  <path d="m14 12 4 4 4-4" />
+</svg>
--- a/icons/a-arrow-up.json
+++ b/icons/a-arrow-up.json
@@ -1,0 +1,21 @@
+{
+  "$schema": "../icon.schema.json",
+  "contributors": [
+    "it-is-not",
+    "jguddas",
+    "danielbayley",
+    "ericfennis"
+  ],
+  "tags": [
+    "letter",
+    "font size",
+    "text",
+    "formatting",
+    "larger",
+    "bigger"
+  ],
+  "categories": [
+    "text",
+    "design"
+  ]
+}
--- a/icons/a-arrow-up.svg
+++ b/icons/a-arrow-up.svg
@@ -1,0 +1,16 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M3.5 13h6" />
+  <path d="m2 16 4.5-9 4.5 9" />
+  <path d="M18 16V7" />
+  <path d="m14 11 4-4 4 4" />
+</svg>
--- a/icons/a-large-small.json
+++ b/icons/a-large-small.json
@@ -1,0 +1,19 @@
+{
+  "$schema": "../icon.schema.json",
+  "contributors": [
+    "it-is-not",
+    "jguddas",
+    "danielbayley",
+    "ericfennis"
+  ],
+  "tags": [
+    "letter",
+    "font size",
+    "text",
+    "formatting"
+  ],
+  "categories": [
+    "text",
+    "design"
+  ]
+}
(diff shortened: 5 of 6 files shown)


jguddas:
The hole inside the small `A` is less than the required 2px which makes it not work with bigger stroke width.

jguddas (changes requested):
The hole inside the small `A` of `font-size` is less than the required 2px which makes it not work with bigger stroke width.

icons/font-size-lower.svg
```diff
@@ -0,0 +1,13 @@
...
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="m5 18 4-8 4 8m-7-2h6M14 6l2 2 2-2" />
```
jguddas, line 12: What about something like this? baseline (1)
```suggestion
  <path d="m4 18 6-12 6 12" />
  <path d="M5.5 15h9" />
  <path d="M15.5 6 18 8.5 20.5 6" />
```

icons/font-size-higher.svg
```diff
@@ -0,0 +1,14 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="m4 18 6-12 6 12" />
+  <path d="M5.5 15h9M16 8l2-2 2 2" />
```
jguddas, line 13: What about something like baseline (2) or baseline?

icons/font-size-lower.svg
it-is-not: I don't mind, I think that the original looks a bit misaligned compared to yours so that looks good

icons/font-size-higher.svg
it-is-not: I prefer the second one as it looks less cluttered

jguddas: Another option so it doesn't look like the `^` character baseline (3)

it-is-not: Good point - I'm not sure about having two chevrons though. Maybe we could go with the one with the arrow, and also do that for the `font-size-lower` icon?

danielbayley:
@it-is-not I did have this in my drafts, but wasn’t really feeling it…

Artboard 1@2x

it-is-not:
Hi @danielbayley I think it's a little unclear what this icon means, but I think we could use it for something similar such as the `font-size-higher` and `font-size-lower` icons, maybe like @jguddas suggested? (https://github.com/lucide-icons/lucide/pull/1222#discussion_r1186708003)

danielbayley:
> I think it's a little unclear what this icon means

@it-is-not It was meant to be `font-`[`re`]`size`

> I did have this in my drafts, but wasn’t really feeling it

Actually, it looks much better like this:

Artboard 1
<br/>
Untitled

<br/>
<br/>

Also, for reference…

Artboard 1
Artboard 1@2
Artboard 1@2x
<br/>
Untitled

<br/>
<br/>

@jguddas What do you think?

jguddas:
Not sure about the resize variant, but I like the other two.

What about the font size icon, I do think that's quite useful to have?

danielbayley:
> This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 5 days.

Remove `stale` label.

it-is-not:
Hi @jguddas sorry for the very late reply, I agree that we should go for the `font-size-higher` and `font-size-lower` icons that @danielbayley suggested but the resize one doesn't seem quite as clear as the other ones to me (perhaps we could come back to it later). I also think that the `font-size` icon is useful but I think there are a few changes we might need to make to it as [it breaks the rules slightly](https://github.com/lucide-icons/lucide/pull/1222#pullrequestreview-1415784846) and, I don't think it quite looks right but I'm not quite sure why.

danielbayley:
@it-is-not Leaving these here in SVG format, if it helps…

preview

it-is-not:
Hi, I also redesigned the `font-size` icon so the small `A` meets the guidelines
image
As well as everything else, we could also have `font-size-plus` and `font-size-minus` icons? Let me know what you think

ericfennis (changes requested):
@it-is-not The arrows should be bigger. Hard to see on 100% scale. Maybe you could use the arrow from `download-cloud`. https://lucide.dev/icons/download-cloud.

For the `font-size` icon we should use the smaller "a" from [`arrow-down-a-z`](https://lucide.dev/icons/arrow-down-a-z)

And we should use generic naming for icons.
So rename this to: `a-arrow-down`, `a-arrow-up` and maybe `a-large-small`.

it-is-not:
Hi @ericfennis thanks for your feedback

- Good point on the naming conventions, I'll change that later

- Here's a draft I just made with your suggestions:
  image
  and here it is at roughly 100% scale:
  
  
  I'm not completely sure about the changes in the style of the letters as it might distract from the icon's purpose, but let me know what you think

ericfennis:
@it-is-not Thanks, this draft looks nice. The both `a-arrow-down, a-arrow-up` looks way better!
But I can imagine your doubts about the `Aa` icon, I think it's fine. But I'm curious what others think.

ericfennis:
@it-is-not Should we try wrap this up? And go for your last designs. We can always improve!

ericfennis:
@it-is-not Just found out we already have the `aa` icon. the [case-sensitive icon](https://lucide.dev/icons/case-sensitive)

it-is-not:
Good spot - what should we do with it in relation to the new font size icons? I don't think it quite has the same meaning as the  icon as they're two different characters (`A` and `a` as opposed to `A` and small `A`) so it probably won't be easily understood for font size, and it also doesn't follow the same naming conventions as you suggested earlier

(and yes, from your comment you marked as outdated, we should definitely try to wrap this up, I was just waiting for community feedback)

ericfennis:
@it-is-not Exactly I just found that one as well. And correct the naming is maybe not the best one. 
I think we should go for you latest `a-arrow-down` and  `a-arrow-up` icons.

ericfennis (approved):
LGTM for now. Maybe we should combine it eventually. But we can do that later

PR #1225, opened 2023-05-06
fix: cogs

jguddas:
- replace small cogs that don't work with 3px stroke width.
- changed cogs, so they are consistent and circularly symmetrical.


jguddas:
I rotated it for the `brain-cog` icon 33°, what about doing that for all icons?

ericfennis (commented):
Nice this looks better. Why is the `folder-cog-2` icon removed?

jguddas:
> Nice this looks better. Why is the `folder-cog-2` icon removed?

It doesn't fit, it would have to be something like this icon (74) to get it to work.

jguddas:
With them being rotated they are sadly less sharp, should I rotate them back?

ericfennis:
@jguddas Can you post a side-by-side comparison image?

jguddas:
> @jguddas Can you post a side-by-side comparison image?

image

ericfennis:
@jguddas Hmm to me actually your changes look the best

ericfennis (approved):

PR #1229, opened 2023-05-07
Add `squirrel` icon

danielbayley:
`squirrel` from both [octicons](https://primer.style/design/foundations/icons) and [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html)… Similar to https://lucide.dev/icon/rat.

Alts:
Artboard 1@2x

Untitled

Also refined `rat` to match…

Related to `rabbit` in #1234.


jguddas:
It feels quite big, maybe we can come up with a version without the outlined tail?

icons/squirrel.svg
```diff
@@ -0,0 +1,15 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M16 22v-2.3a2.7 2.7 0 0 1 2.7-2.7H20a2 2 0 0 0 2-2v-1c0-2.8-2.2-5-5-5V5a4 4 0 0 0-4 4v2a7 7 0 0 0-7 7V6a2 2 0 0 0 4 0 4 4 0 0 0-8 0v12a4 4 0 0 0 4 4h12" />
+  <path d="M15.6 22a3.5 3.5 0 0 0-3.2-5" />
+  <path d="M18 13h.01" />
```
jguddas, line 14: The eye needs to be moved to have 2px space in all directions.
```suggestion
  <path d="M16 22v-2.3a2.7 2.7 0 0 1 2.7-2.7H20a2 2 0 0 0 2-2v-1a5 5 0 0 0-5-5V5a4 4 0 0 0-4 4v2a7 7 0 0 0-7 7V6a2 2 0 0 0 4 0 4 4 0 0 0-8 0v12a4 4 0 0 0 4 4h12" />
  <path d="M15.6 22a3.5 3.5 0 0 0-3.2-5" />
  <path d="M 17.5 13h0.01" />
```
icon (44)

karsa-mistmere:
Very cute, but I feel like the tail is too geometric.

danielbayley:
> It feels quite big, maybe we can come up with a version without the outlined tail?

It’s recognisable as a squirrel as big tail is their main feature… Why have a 24x24px grid if we can’t use it? It was modelled on the `rat`… Also, not sure how else the tail could possibly look? Feel free to iterate though…

karsa-mistmere:
Here's my original idea (2nd icon), in practice it didn't quite work out, I feel like a more organic tail (3rd icon) would maybe look better:
image

I think the original tail looks more like a chinchilla's (curling towards the body) than a squirrel's (their tails typically curl away from the body and is much more bushy).

danielbayley:
> I feel like a more organic tail (3rd+4th icons) would maybe look better
> a squirrel's (their tails typically curl away from the body and is much more bushy).

Last one works well enough…

danielbayley:
@karsa-mistmere @jguddas Tweaked these a bit… You might like the little guy’s tail on the top row, but I think there is a 2px space violation between his ear and the tail…

Artboard 1

Artboard 1@2x

Untitled

jguddas:
> Tweaked these a bit… You might like the little guy’s tail on the top row, but I think there is a 2px space violation between his ear and the tail…

Maybe we should tweak the rules instead.

I don't think we are loosing detail when the gap closes, so it shouldn't be an issue @karsa-mistmere ?

karsa-mistmere:
I think in this instance the gap is fine, at least from my side.

danielbayley:
@jguddas @karsa-mistmere See updated https://github.com/lucide-icons/lucide/pull/1229#issue-1699172507

ericfennis (approved):

karsa-mistmere:
Nice work, @danielbayley!

PR #1232, opened 2023-05-08
Add `layout-list-check` icon

danielbayley:
`tasklist` from octicons: https://primer.style/design/foundations/icons/tasklist-16, used right here 👆🏼

Untitled


jguddas:
What about changing the layout list from 4 to 3 lines?

danielbayley:
> What about changing the layout list from 4 to 3 lines?

Fine with me, although it’s currently modelled on https://lucide.dev/icon/layout-list

jguddas:
> > What about changing the layout list from 4 to 3 lines?
> 
> Fine with me, although it’s currently modelled on https://lucide.dev/icon/layout-list

We would need to change it for all the layout lists, but then we have it consistent with the rest of the list icons.

danielbayley:
> We would need to change it for all the layout lists, but then we have it consistent with the rest of the list icons.

`push`ed…

jguddas (approved):

karsa-mistmere:
I think this use case is already covered by https://lucide.dev/icon/list-checks?search=Todo

karsa-mistmere:
@ericfennis: That one's maybe a bit too tiny though, maybe we could replace that with a pair of these (2nd & 3rd)?
image
`list-todo` & `list-checks`

These would maybe match `list-ordered` more.

danielbayley:
@karsa-mistmere @jguddas @ericfennis Implemented…

icons/layout-list.svg
karsa-mistmere, line 1: I'm not in favour of this change, I think `layout-list` should be consistent with other grid and layout icons, not the list icons and this would break that.

icons/list-todo.svg
```diff
@@ -0,0 +1,17 @@
...
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect width="7" height="7" x="3" y="3" rx="1" />
+  <path d="m4 17 2 2 4-4" />
+  <path d="M14 6h7" />
+  <path d="M14 12h7" />
+  <path d="M14 18h7" />
```
karsa-mistmere, line 16:
```suggestion
  <rect x="3" y="5" width="6" height="6" rx="1" />
  <path d="m3 17 2 2 4-4" />
  <path d="M13 6h8" />
  <path d="M13 12h8" />
  <path d="M13 18h8" />
```

icons/list-checks.svg
```diff
@@ -9,10 +9,9 @@
...
-  <line x1="10" x2="21" y1="12" y2="12" />
-  <line x1="10" x2="21" y1="18" y2="18" />
-  <polyline points="3 6 4 7 6 5" />
-  <polyline points="3 12 4 13 6 11" />
-  <polyline points="3 18 4 19 6 17" />
+  <path d="m4 7 2 2 4-4" />
+  <path d="m4 17 2 2 4-4" />
+  <path d="M14 6h7" />
+  <path d="M14 12h7" />
+  <path d="M14 18h7" />
```
karsa-mistmere, line 16:
```suggestion
  <path d="m3 17 2 2 4-4" />
  <path d="m3 7 2 2 4-4" />
  <path d="M13 6h8" />
  <path d="M13 12h8" />
  <path d="M13 18h8" />
```

karsa-mistmere (changes requested):
The other two should however be more consistent with other list icons, e.g.:
image

I also added these as suggestions.

ericfennis:
I'm don't we should change the layout-list icon. Layout list is used for card lists with images or something else. Now it is turned into a list with check boxes.

jguddas:
What about something like this for `layout-list`? circle (6)

karsa-mistmere:
> What about something like this for `layout-list`? circle (6)

@jguddas: This would still be very inconsistent with `grid` and other `layout-*` icons.

@danielbayley: Since Eric and I both agree, let's keep `layout-list` as it is, the others are good to go.

ericfennis (approved):

karsa-mistmere (approved):

PR #1236, opened 2023-05-08
Add `arrow-ligature` (function) icon

danielbayley:
Direct alternative to https://lucide.dev/icon/function-square, and a more recognisable, literal representation of function/method/etc…

See also other code/syntax related icons, e.g https://lucide.dev/icons?search=brackets

Related to #1063 and #1134.


ericfennis:
Can we have an use-case for this? I know the meaning, but never saw this used as an icon.

danielbayley:
> Can we have an use-case for this? I know the meaning, but never saw this used as an icon.

@ericfennis Sure, sorry forgot to add for this one… See https://github.com/lucide-icons/lucide/pull/1236#issue-1701062635

But more specifically, planning to use this in my complete VS Code _Lucide_ icons package, as `symbol-function`/`method`/`constructor` in the outline view (codicons `symbol-*` icons…)

I would have added it to #1134, but it already got `merge`d :)

jguddas:
should not be called `-function` thats a use case, not sure what we could call it, `arrow-big` maybe, idk.

danielbayley:
> should not be called `-function` thats a use case, not sure what we could call it, `arrow-big` maybe, idk.

@jguddas `arrow-fat` is aka…

jguddas:
> > should not be called `-function` thats a use case, not sure what we could call it, `arrow-big` maybe, idk.
> 
> @jguddas `arrow-fat` is aka…

`arrow-ligature` feels pretty good, what do you think?

jguddas (approved):

ericfennis:
The use case is a bit too specific for the lucide library.
I understand it is an arrow, but in the context of "no-code", I wouldn't use this kind of icon to indicate a function.
For this use case [function-square](https://lucide.dev/icons/function-square) will do the job fine.

PR #1172, opened 2023-04-28
Add `tab-*` icons

danielbayley:
`tab-arrow-up-right` is `tab-external` from [octicons](https://primer.style/design/foundations/icons).

Edit: _Possibly_ superseded by #1237.

Issue #1192, opened 2023-04-30
Some icons feel a bit small


jguddas:
Legend: ✅ Fixed | 🚧 Under construction | ❓ Unsure | 🟥 Rejected

|   | Name | Icon | Alternative | Addressed by |
| --- | --- | --- | --- | --- |
| 🟥 | banknote | ![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/square.svg)![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/banknote.svg) |   |   |
| ✅ | bot | ![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/square.svg)![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/bot.svg) |   | #934 |
| ✅ | bug | ![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/square.svg)![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/bug.svg) | ![](https://raw.githubusercontent.com/lucide-icons/lucide/13e4735d04606cd7418ea52237c2fbb11aac68e6/icons/bug.svg) | #988 |
| ❓ | case-lower | ![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/square.svg)![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/case-lower.svg) |   |   |
| ❓ | case-sensitive | ![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/square.svg)![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/case-sensitive.svg) |   |   |
| ❓ | case-upper | ![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/square.svg)![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/case-upper.svg) |   |   |
| ❓ | gamepad | ![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/square.svg)![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/gamepad.svg) |   |   |
| 🟥 | mailbox | ![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/square.svg)![](https://raw.githubusercontent.com/lucide-icons/lucide/main/icons/mailbox.svg) |   |   |
[...]

PR #1237, opened 2023-05-09
Optimise/rename `vote`/add `tab`[`-*`]/generic output icons

danielbayley:
Also covers `output` from [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html).

Untitled
<br/>
Artboard 1@2x
<br/>
<br/>

Possibly closes #1172…?


karsa-mistmere:
Not convinced about the names, I also wanted to submit a similar PR, in which I was planning to use the names `ballot`, `ballot-check`, `ballot-x`, by extension the extra icon in this PR would be `ballot-text`, I guess.

I'm open to other ideas, as long as the new names are more consistent.

jguddas:
Feels a bit like a browser-tab, probably due to not having a box the ballot is getting put in.

danielbayley:
> Not convinced about the names, I also wanted to submit a similar PR, in which I was planning to use the names `ballot`, `ballot-check`, `ballot-x`, by extension the extra icon in this PR would be `ballot-text`, I guess.

@karsa-mistmere Works for me…

> Feels a bit like a browser-tab

That’s what #1172 is for…

karsa-mistmere:
> > Feels a bit like a browser-tab
> 
> That’s what #1172 is for…

I do kind of agree with @jguddas on this, it does look conspicuously tab-like.

I did have a few alternate designs lying around in my drafts, but was OK with the one that got merged because mine all had some setbacks.
image

danielbayley:
> I do kind of agree with @jguddas on this, it does look conspicuously tab-like.

Hmm, but I do like the simplicity of these…

Maybe we could just have them all as a single set, with appropriate `tags` for both? (merge with #1172)

karsa-mistmere:
`tab-text` also kinda looks like a tombstone. 😅

danielbayley:
> `tab-text` also kinda looks like a tombstone. 😅

@karsa-mistmere I can add that to the metadata… :)

> What about making it bigger, #1192 ?

@jguddas See updated options: https://github.com/lucide-icons/lucide/pull/1237#issuecomment-1540182422

karsa-mistmere (approved):

ericfennis:
@danielbayley Can you rebase the PR?
I'm not sure about the use case for the `tab-slash` icon. Can we clear this up?

ericfennis:

PR #1244, opened 2023-05-09
Add `tally` icons

### LINKED ISSUE #752: tally and counter icons
## Icon Request

* Icon name: tally / counter
* Use case: a way to describe a total or a count of something, a tally or counter icon would make sense to display that something is be summed up or calculated.

image

image

danielbayley:
closes #752

Keeping count/score…

Alts:
Artboard 1@2x

Untitled


jguddas:
What about changing it so you can see the start and end of the first and last line?

danielbayley:
> What about changing it so you can see the start and end of the first and last line?

@jguddas Like that?

jguddas:
Not centered unless moved right by 0.5px.

karsa-mistmere:
This is connected to #752, I'm still not sure about centred vs left aligned.

danielbayley:
> Not centered unless moved right by 0.5px.

Yeah, that was intentional… ideally we don’t want things of grid, right? Problem is trying to fit 5 equal spaces into an even numbered grid…

> This is connected to https://github.com/lucide-icons/lucide/issues/752, I'm still not sure about centred vs left aligned.

Oh man, another can of worms I didn’t anticipate 🙈
I honestly haven’t gone through the [issues](https://github.com/lucide-icons/lucide/issues) (_or_ your offline drafts @karsa-mistmere 😅)

ericfennis:
If you have a list of items with these icons used, reading them in a list left aligned will be preferred. Also if you want to show 6 for example, you can combine `tally-5` and `tally-1` after eachother. So think left-aligned (current design) is the way to go.

ericfennis (approved):

PR #1253, opened 2023-05-10
Fix `text` icon alignment/add `text-keyword` icon

danielbayley:
Fixes https://lucide.dev/icon/text, and `text-keyword` like `symbol-keyword` from [codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html).

### DIFF
--- a/icons/text-keyword.svg
+++ b/icons/text-keyword.svg
@@ -9,8 +9,8 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect width="6" height="4" x="3" y="4" rx="1" />
-  <path d="M17 6h-4" />
-  <path d="M21 12H3" />
-  <path d="M15 18H3" />
+  <path d="M17 6H3" />
+  <rect width="8" height="4" x="3" y="10" rx="1" />
+  <path d="M21 12h-6" />
+  <path d="M15.1 18H3" />
 </svg>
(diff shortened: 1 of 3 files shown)


icons/text-keyword.svg
```diff
@@ -0,0 +1,16 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M17 6H3" />
+  <path d="M10 10c.6 0 1 .4 1 1v2c0 .6-.4 1-1 1H4c-.6 0-1-.4-1-1v-2c0-.6.4-1 1-1Z" />
```
jguddas, line 13:
```suggestion
  <rect x="3" y="10" width="8" height="4" rx="1" />
```

icons/text-keyword.svg
karsa-mistmere, line 1: Feels a bit `list-box-minus`-y, if you catch my drift. How about?
image

ericfennis:
A real use case is not provided. If you use this for example to search for keywords, you should use the search icon.
I also tried to find similar icons in other icon libraries. But I couldn't find any, except for your link to codicons.
So I close this for now.

danielbayley:
> A real use case is not provided.

> for example to search for keywords

The real use case is to represent keywords, in code symbol navigation, for example…

image

What do you suggest I use for `vscode-lucide-icons` instead?

@ericfennis Also, `text` still needs the fix in this PR, as for some reason it is currently misaligned… So if you reopen it I can revert the other `commit`, and just `merge` that for now…

PR #1256, opened 2023-05-10
Add `gauge` alternate icon

danielbayley:


jguddas (commented):
What about calling it circle gouge?

ericfennis (approved):

icons/gauge-circle.json
```diff
@@ -0,0 +1,18 @@
...
+    "meter",
+    "speed",
+    "pressure",
+    "measure",
+    "level"
+  ],
+  "categories": [
+    "account",
+    "transportation",
+    "sports",
```
ericfennis, line 15: why is this part of sports?

ericfennis: Oh I see was already part of `gauge` check.

danielbayley: > why is this part of sports?

> Oh I see was already part of `gauge` check.

@ericfennis Ah yeah, and in answer to the question, specifically, motorsports…

PR #1257, opened 2023-05-10
Add `key` variants

danielbayley:


jguddas:
The tip feels a bit heavy.

ericfennis (approved):
Nice!

danielbayley:
I think the existing `key` is a bit weak… What do you guys think of replacing it with `key-2`, and just having these 2 variants?

jguddas:
I quite like the current `key` icon, not a fan of those more realistic options, it's just not cute enough.

My vote is for keeping it like it is.

karsa-mistmere:
In hindsight, maybe we could do with lower density?
image
image

karsa-mistmere:
> > In hindsight, maybe we could do with lower density?
> 
> @karsa-mistmere That’s basically what I had originally! 😆

Similar, but not quite:
image

danielbayley:
> In hindsight, maybe we could do with lower density?

@karsa-mistmere `push`ed… Looking good now 👍🏼

ericfennis (approved):

PR #1234, opened 2023-05-08
Add `rabbit` & `turtle`/`snail` & `shell`/aminal icons

danielbayley:
Can represent speed, fast/slow (tortoise/hare…)

Untitled

Related to `squirrel` in #1229.

Alts:
Artboard 1@2x

Untitled

<br/>
<br/>

Artboard 1@2x
<br/>
<br/>
Untitled
<br/>
<br/>

Updated to better match the Lucide logo, and `lollipop` in #1149.

PR #1258, opened 2023-05-10
feat: added simple edit page

jguddas:
https://lucide-site-git-feat-added-simple-edit-page-jguddas.vercel.app/edit/icon


jguddas:
I build something to make it easier to submit icons.

I would love your feedback @g-angeles-aua @it-is-not @karsa-mistmere.

https://lucide-site-git-feat-added-simple-edit-page-jguddas.vercel.app/edit/icon

karsa-mistmere:
I really really like this as a concept (and actually use it a lot already when reviewing PRs), but the optimization script still needs some serious tweaking before we can publish this for public use, as it often results in really funky results that are way off the original.

And... it would also need to be able to run in the new, Vitepress based site (see the branch `docs-vitepress`). 🙈

it-is-not:
Hi @jguddas I like it a lot - I like how you can just copy and paste code from Figma (or other editors) into it without needing to worry about indentation, formatting and optimization.

I thought of a few ideas:

- In comments like [this](https://github.com/lucide-icons/lucide/pull/1229#issuecomment-1537506272), it shows the icons next to other icons, and different stroke widths and it at 24*24 resolution. Would it be possible to do that underneath the preview on the edit page as well?
- Syntax-highlight the code editor, so it's easily to read and debug the code (something like [this](https://css-tricks.com/creating-an-editable-textarea-that-supports-syntax-highlighted-code/)?)
- We should also make it so it follows many of the conventions we have such as how we code dots in svg.
- I've spend quite a bit of time in the past finding out how icons are stored and how the repository works, for example, there used to be a `tags.json` file with all the icons' tags in it. Nowadays, along with the `icon.svg` file there is also a `icon.json` file, for example in https://github.com/lucide-icons/lucide/pull/1234/commits/da11c96e2623375fa5053a1ccf5554da4ddedb20. Perhaps the edit page could put this file in the commit as well, and also allow the user to enter tags and categories?

jguddas:
> In comments like https://github.com/lucide-icons/lucide/pull/1229#issuecomment-1537506272, it shows the icons next to other icons

Great idea, I will see what I can come up with.

> Syntax-highlight

Is already complex enough IMO.

> We should also make it so it follows many of the conventions we have such as how we code dots in svg.

I already implemented a bunch of things including a `fixDots` function, but like @karsa-mistmere alluded too above, it isn't ready yet.

https://github.com/lucide-icons/lucide/blob/e009988c6696da05a6544eaf4bc256f9f49b48fe/site/src/pages/api/edit/optimize.ts#L851-L865

You can test it out by pressing optimize at https://lucide-site-git-feat-added-edit-page-jguddas.vercel.app/edit

Most things the user never needs to know about and can be automatically fixed, but some can either only be detected or need to be checked manually.

image

> Perhaps the edit page could put this file in the commit as well

We should create a `/edit/metadata` page and paste a link on the PR.  
We can't do it on the `/edit/icon` page due to GitHub only allowing you to edit one file at a time using the web interface.

jguddas:
-

PR #1268, opened 2023-05-11
Add `puzzle` (jigsaw) variants

danielbayley:
Leaving all explored options here for reference/comparison…

preview


jguddas:
The tabs of the normal puzzle icon feel a lot more puzzly to me.

danielbayley:
> The tabs of the normal puzzle icon feel a lot more puzzly to me.

@jguddas See latest `push`, and alts above… https://github.com/lucide-icons/lucide/pull/1268#issue-1706715572

karsa-mistmere:
The notches are too small and dense and the whole icon is a bit too bottom left heavy, maybe something like this?
puzzle-2

danielbayley:
> maybe something like this?

@karsa-mistmere Works for me…

jguddas:
Both feel a bit small.

For the naming, they could all be `something-puzzle-piece`.

The first one could be `irregular` and the second one could be `stack`

Scratch wiki: hat, stack, reporter, boolean, or cap

jguddas:
Hmm, not sure edge really describes it well.

danielbayley:
Looks like the [comment](https://github.com/lucide-icons/lucide/pull/1268#issuecomment-1544901058) bot has gone haywire again 🤪

danielbayley:
> Hmm, not sure edge really describes it well.

@jguddas Looks more like an “edge piece” to me, than anything else… Maybe `puzzle-piece-simple`?

jguddas:
> > Hmm, not sure edge really describes it well.
> 
> @jguddas Looks more like an “edge piece” to me, than anything else… Maybe `puzzle-piece-simple`?

For it to be an edge piece it would have to have one less flat side.

danielbayley:
> For it to be an edge piece it would have to have one less flat side.

@jguddas Technically yes… What do you think of `-simple` or `-minimal`?

ericfennis:
Can we have some use cases described why we should have another puzzle icon?

danielbayley:
> Can we have some use cases described why we should have another puzzle icon?

@ericfennis One obvious use case is to represent plugins/extensions (not just for browsers, but to lots of other apps/editors)…

The existing diagonal piece doesn’t necessarily fit so nicely into a toolbar, or alongside other, more squared, icons.

I’m personally less bothered about `puzzle-2` (if you guys were to insist on dropping one—although it looks more like a stereotypical puzzle piece…) But will definitely be using the nice and simple (so possibly the most Lucide appropriate) variant (currently named `puzzle-3`) for a bunch of things going forward… e.g to represent a plugins/extensions section on a portfolio site.

Also, I have since noticed these all over the place in the wild…

`puzzle-2` here for example in Chrome/Brave:

Untitled

And the simple `puzzle-3` here in Safari iOS, among other places…

IMG_7703E9C80CEB-1

karsa-mistmere:
I think the diagonal icon could work here just as well, I kind of feel like this is not really a separate use case, more of a personal preference.
image

That being said, as far as I'm concerned, we could add more variants, as long as have sufficiently large gaps between opposing sides of the notches and are well balanced in general, but let's keep things simple, two are probably already plenty, we don't need to cover all possible puzzle pieces in existence. :)

danielbayley:
> That being said, as far as I'm concerned, we could add more variants … but let's keep things simple, two are probably already plenty, we don't need to cover all possible puzzle pieces in existence. :)

Sure, I wasn’t planning on adding any more! :) Like I said, (currently named) `puzzle-3` trumps the others for me…

So you are saying I rename that (`-3` to `-2`) and we just go with those pair…? @karsa-mistmere cc: @ericfennis @jguddas

ericfennis (approved):
I'm ok with these changes in this PR

karsa-mistmere:
I'm not so sure. 🤔

What exactly does the alternate bring to the table apart from being very vertically off centre?
image

jguddas:
> I'm ok with these changes in this PR

I'm against adding this variant.

ericfennis:
With the comments from @karsa-mistmere and @jguddas I think it's better we close this one and first discuss if we need these variants.

PR #1270, opened 2023-05-12
Optimise `grid`/add [`layout-`]`grid` `move`/`list` improvements + metadata

danielbayley:
Related to #1184 and #1498.

### DIFF
--- a/icons/crop.json
+++ b/icons/crop.json
@@ -7,7 +7,13 @@
     "karsa-mistmere"
   ],
   "tags": [
-    "image"
+    "image",
+    "frame",
+    "area",
+    "select",
+    "scene",
+    "size",
+    "proportions"
   ],
   "categories": [
     "photography",
--- a/icons/crop.json
+++ b/icons/crop.json
@@ -7,11 +7,12 @@
     "karsa-mistmere"
   ],
   "tags": [
-    "photo",
     "image"
   ],
   "categories": [
     "photography",
-    "design"
+    "design",
+    "layout",
+    "tools"
   ]
 }
--- a/icons/frame.json
+++ b/icons/frame.json
@@ -6,11 +6,12 @@
     "ericfennis"
   ],
   "tags": [
-    "logo",
-    "design",
+    "grid",
+    "guidelines",
     "tool"
   ],
   "categories": [
+    "layout",
     "design",
     "photography"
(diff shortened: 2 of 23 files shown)


karsa-mistmere:
We do have a separate # icon, so I don't think adding those use cases to `grid-2` is necessary, I definitely do not wish for us to encourage people to use this as a hashmark.

Also, maybe we could call it `grid-lines` or `guidelines` instead? I really hate numbering our alternate designs, don't get me wrong, I'm definitely guilty of having done the same, but I think a qualifier of what makes them unique is much more helpful.

danielbayley:
> maybe we could call it `grid-lines`

@karsa-mistmere Done.

I’ve noticed that occasionally I go to add something I think is missing, then later find it accidentally, having missed it because the search didn’t bring it up due to lacking metadata… For example `frame` here, although different enough in this case.

> I think a qualifier of what makes them unique is much more helpful.

Yeah agreed, where possible…

danielbayley:
> maybe we could call it `grid-lines`

@karsa-mistmere Done.

I’ve noticed that occasionally I go to add something I think is missing, then later find it accidentally, having missed it because the search didn’t bring it up due to lacking metadata… For example `frame` here, although different enough in this case.

> I think a qualifier of what makes them unique is much more helpful.

Yeah agreed, where possible…

ericfennis:
hmm, the plus from `layout-grid-plus` is not aligned on the grid.
This will result in a blurry plus on low res screens. Not sure what we should do. @karsa-mistmere what do you think?
image

danielbayley:
> on low res screens

We need this kind of low-res preview added to the bot!

@ericfennis Not sure either in this case! Is supporting low res screens a higher priority than having certain icons at all? Surely there aren’t too many around by now… Also, technically we ‘support’ 3px stroke width, but many of the icons look like shit at 3px…

jguddas:
> > on low res screens
> 
> We need this kind of low-res preview added to the bot!

https://github.com/jguddas/lucide/commit/fa7623c2f433230cecd5ef740784e08b2eff3bc7

Sorry for it just laying there, it got lost in the migration to the new site.

I don't have time to look at it, so if someone wants to copy it over please do 🙏

ericfennis:

PR #1271, opened 2023-05-12
Fix Cursor for Customize and Small Icon Buttons; Reverse Sidebar Open/Close Icon; Remove Unnecessary Customize Button on Mobile Header

ciruz:
Hey,

the `Customize` Button had not the right cursor (same for the small Icon Button).

Before:
image
image

After:
image
image

And I reversed the Sidebar Open/Close Icon, since it makes more sense this way.

**Before:**

To open the sidebar:
image
To close the sidebar:
image

**After:**

To open the sidebar:
image
To close the sidebar: 
image

Before/After:
before
after

I also think that the `Customize` Button in the `Header` is not necessary (on mobile), because only the Icon is shown with a different button height, this looks weird and there is also a second `Customize` Button just below, with the same functionality.

Before:
image

After:
image

### DIFF
--- a/site/src/components/Header.tsx
+++ b/site/src/components/Header.tsx
@@ -1,4 +1,4 @@
-import { Button, Flex, Link, WrapItem, Text, Wrap, Heading, Box } from '@chakra-ui/react';
+import { Button, Flex, Link, WrapItem, Text, Wrap, Heading, Box, Hide } from '@chakra-ui/react';
 import download from 'downloadjs';
 import { Download, Github } from 'lucide-react';
 import NextLink from 'next/link';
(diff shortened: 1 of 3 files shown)


karsa-mistmere:
Nice work, I think the icons were probably mixed up, because on page init the wrong sidebar icon is shown until it's been toggled off and on again, which is still the case with this PR: the sidebar is open but `panel-left-open` is shown.

ciruz:
@karsa-mistmere you are right, because `sidebarOpen` is initally `undefined`. To address this, I updated the code to check for `undefined` so that the Icon is initally correct too. But since the `Sidebar` is open by default on desktop, but not on mobile, the Icon now is not correct on Mobile.

ciruz (commented):
It's done via ``useBreakpointValue`` https://github.com/lucide-icons/lucide/blob/main/site/src/components/IconCategoryDrawer.tsx#L60.

I implemented it the same way for the Icon now.

Initial on mobile now:
image

Initial on desktop now:
image

ericfennis:
Thanks for you contribution, we appreciate this! 
But we are happy to announce a new website, see #1275. These fixes are now outdated.

PR #1273, opened 2023-05-12
add missing key for default menu items

ciruz:
The external `Link` has a `key` property ( https://github.com/lucide-icons/lucide/blob/main/site/src/components/MobileMenu.tsx#L49 ), but not the default `Link`, this can cause a warning (like: `Warning: Each child in a list should have a unique "key" prop.`).

### DIFF
--- a/site/src/pages/docs/index.tsx
+++ b/site/src/pages/docs/index.tsx
@@ -42,7 +42,7 @@ const DocPage = ({ doc, data }) => {
         <Box maxWidth="xl" width='100%' marginX='auto'>
           <MDXRemote {...doc} data={data} components={mdxComponents} />
 
-          { nextPage.map((page, i) => (
+          { nextPage.map((page) => (
             <Link href={`docs/${page}`}>
               <Button
                 variant="ghost"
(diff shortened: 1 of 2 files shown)


site/src/pages/docs/index.tsx
```diff
@@ -42,7 +42,7 @@ const DocPage = ({ doc, data }) => {
         <Box maxWidth="xl" width='100%' marginX='auto'>
           <MDXRemote {...doc} data={data} components={mdxComponents} />
 
-          { nextPage.map((page) => (
+          { nextPage.map((page, i) => (
```
jguddas, line 45: > We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state.

https://legacy.reactjs.org/docs/lists-and-keys.html#:~:text=We%20don%E2%80%99t%20recommend%20using%20indexes%20for%20keys%20if%20the%20order%20of%20items%20may%20change.

ciruz: you absolutely right

ericfennis:
Close to outdated code, see comment in #1271

PR #1274, opened 2023-05-12
remove console.log statements

ciruz:
Not sure if they are left by mistake.


ericfennis:
Close to outdated code, see comment in #1271

PR #1028, opened 2023-04-09
Sorting out our `sort` icons

karsa-mistmere:
This pull request originates from #1004 by @danielbayley

During its review I noticed what I believed were inconsistencies about arrow directions in our current sort icons ("surely, sort-asc should point _downwards_"), after doing some research, I believe these perceived inconsistencies probably stem from cultural differences, a summary of what I've found:

Icon sets where `asc=up`, `desc=down`:
- Lucide
- [Octicons](https://primer.style/design/foundations/icons/sort-asc-24)
- [css.gg](https://css.gg/app) [basically, although implicitly]
- [Remix icon](https://remixicon.com/)

Icon sets where `asc=up`, `desc=down`:
- [Fontawesome v4](https://fontawesome.com/v4/icon/sort-alpha-asc) 
- [Phosphor Icons](https://phosphoricons.com/)
- [Tabler Icons](https://tablericons.com/)

Icon sets where the order is not stated, but instead only the direction is used (up or down):
- [Fontawesome v6](https://fontawesome.com/search?q=a-z&o=r) 
- [Bootstrap](https://icons.getbootstrap.com/?q=sort)
- [Unicons](https://iconscout.com/unicons/explore/line?redirect=%2Funicons%2Fexplore%2Fline%3Fcolor%3D%2525230092E4%26name%3Dsort-amount-down)
- [Zwicon](https://www.zwicon.com/cheatsheet.html)

---

In this PR I propose to follow in the footsteps of the "mainstream" and renamed our sort icons to not obviously reflect any order so that the exact usage is up to the end users (aliases were added for backwards compatibility):
e.g. `sort-asc` becomes `arrow-up-narrow-wide` and `sort-desc` becomes `arrow-down-wide-narrow`

Further icons supporting ascending and descending numerical and alphabetical order were also added:
image

As a side effect, the now incorrectly named `arrow-up-down` and `arrow-left-right` were made consistent with the sort icons and matching pairs were added (`arrow-down-up` and `arrow-right-left`).
(We may choose to retain the legacy design under a different name, as they have a significantly different feel).

PR #1275, opened 2023-05-12
New site! 🚀

ericfennis:
image

# New site! 🚀

Lucide is moving to use [Vitepress](https://vitepress.dev/) for the new marketing/documentation site.
@karsa-mistmere and I worked on this for the last weeks.

TLDR;

The current website does not fit the requirements anymore. With 10000+ visitors each month we need to have something good. And with the current stack we won't achieve this with the time we have. For Lucide we need something that is a good foundation for the next documentation we want to write for Lucide. Current website feels a bit buggy, has a lot of performance issues and build times are getting out of hand. To have a fast website have chosen a new stack: Vitepress + Nitro (for API) and Vue and plain css for the rest.

You all probably have some questions regarding this change. Here are some we try to answer.

## What was wrong with the old site?

The old site was a custom build docs site using React, NextJS, CharkraUI and custom MDX loaders for documentation part. This was great for the start of Lucide and it worked pretty well but now the project has been grown and the current experience of the website is lagging and not match the quality we want to show. It is struggling with performance issues, web vitals are scoring low [69](https://pagespeed.web.dev/analysis/https-lucide-dev/p7qorrlqqe?form_factor=mobile) and things don't work really well.

## Why not fix the old site?

[ChakraUI](https://chakra-ui.com/) is great for get something quick out of the ground and it has an excellent theming system, but it is not best choice to make a fast website. Chakra is using emotion under the hood and this really hard to get it fast, because CSS styling is injected on runtime. When using dark mode the whole website is flashing because the styles need to be injected first. Not the best experience. 

We maybe can fix everything if we spend a lot of time in it but maybe we will score 80 in web vitals, we think this is not worth the effort. Also we already spend a lot of time in improving the site, but it didn’t make big steps.

Other thing is that we don't have the full time to work on the website, the current website was really custom meaning you need to build a lot of things your own. In the beginning we think we will do that but we don't.

## Why Vitepress?

Vitepress is made with performance in mind and has excellent markdown features build in. And is made for open source dev projects. Some other open source projects you maybe know are also using Vitepress as there docs website, like: Rollup & Vite, Vitest and Vue . And I think they are really well made. Because the foundation is really solid.

Some features from Vitepress:

- Markdown extensions: like code groups, line highlighting, alerts, tables etc. [[See all here](https://vitepress.dev/guide/markdown)](https://vitepress.dev/guide/markdown)
- Table of contents sidebar.
- Next previous buttons
- Full site search with Angolia.

And hopefully more will come, Vitepress has a very active community.

## Did you consider other markdown site generators

Yes we did, vitepress matched our requirements.

## Are we still able to make REST APIs?

Yes, we currently have some rest APIs in Next.js. One of them is showing icons previews in PRs, helping us reviewing icon PRs. This one will still work with our new setup.

We successfully integrated [[Nitro](https://nitro.unjs.io/)](https://nitro.unjs.io/) together with Vitepress when deploying to Vercel. Plus it now runs on the Vercel Edge servers, meaning that we don’t have cold starts and everything will working a bit faster as well.

APIs source files will be located in: `docs/.vitepress/api`

## What are the performance gains?

We are now scoring a solid [[95](https://pagespeed.web.dev/analysis/https-lucide-docs-vercel-app/ytkptbmy7r?form_factor=mobile)](https://pagespeed.web.dev/analysis/https-lucide-docs-vercel-app/ytkptbmy7r?form_factor=mobile) on performance, which is really impressive. And we have build times for 1000+ pages around 4 minutes (current build times are 8+ min.).

## Can I see a preview?

Yes you can!

Preview of the new site:

[https://lucide-docs.vercel.app/](https://lucide-docs.vercel.app/)

## So what’s next?

We want to first merge the new foundation of website, which includes all the features we currently have on the website + a new landing page and icon detail pages. Which has code examples quick copy buttons previews and related icons.

Other things we want to include in the website:

- New improved documentation
    - More detailed guides about using lucide icons
    - Code and project examples with e.g. Codesandbox
    - Improved design guides
    - and more.
- Figma plugin showcase page
- Icon detail with icons in action

And suggestions are welcome!

## Bugs and feedback.

Please let us know what you think of the new website.
If you find bugs or have feedback for improvements let us now.


jguddas:
Changing color stroke-width or size using the customizer is laggy on my machine.

scripts/writeReleaseMetadata.mjs
```diff
@@ -0,0 +1,142 @@
...
+const tags = await fetchAllReleases();
+
+const comparisonsPromises = tags.map(async (tag, index) => {
+  const previousTag = tags[index - 1];
+
+  if (!previousTag) return undefined;
+
+  const diff = await git.diff(['--name-status', '--oneline', previousTag, tag]);
+  const files = diff.split('\n').map((line) => {
+    const [status, file, renamedFile] = line.split('\t');
```
karsa-mistmere, line 43: I understand the motivation behind switching to renames from aliases support, however, this is by no means foolproof and will only work as intended if the renames were done in an exact manner.

ericfennis: No that's not exactly true. When squashing PRs to main branch, Git recognizes the renames, it is sometimes not a fully matching. But Git has a special status for this, it is: `R{percentage of the match}`. So he knows it renamed but also 50 matching, This way we can check on this see: line 75 from this file.
The line here above can be undefined, if not set renamed file will not be used.

karsa-mistmere: Yes, but it's still not foolproof though; I've seen false positive renames in PRs when an icon was added and another renamed and the wrong icon (the new one) was detected as the original since the SVG code was deemed to be closer.

karsa-mistmere: E.g. with https://github.com/lucide-icons/lucide/pull/1028
`git diff --name-status --oneline v0.173.0 v0.174.0` will output:
```
R062    icons/sort-asc.svg      icons/arrow-down-up.svg
R051    icons/sort-asc.json     icons/arrow-down-narrow-wide.json
R062    icons/sort-desc.svg     icons/arrow-down-narrow-wide.svg
R051    icons/sort-desc.json    icons/arrow-up-wide-narrow.json
```

Which is a complete and utter mess, and is quite weird, since I did in fact use `git mv` for these. 🤷‍♂️

ericfennis: But wait is it a problem? We only need to check when the version is changed right? So if the renames are not right but are visible in the version change then still the version tag is valid for that specific icon right?

ericfennis: Ok I think I got it sorted.

I don't use the rename filename value to detect the renames but only create a object map with all the icon names with created dates and changed date. After that we loop over all the icons and also check the aliases. If aliases exist we apply the created version tag of the alias as well. So I use aliases again.

The metadata is still not 100% correct if we use aliases. There where some renames before we started using aliases and some aliases are intentionally not applied because the old name was used again for other icon design. But I think 97% correctness is ok, let's move on.

BTW, the builds are now respectively 3:30 min, so that pretty neat

jguddas:
What about moving the whole thing into `apps/docs` so we can later add things like `apps/editor` and such?

PR #1278, opened 2023-05-16
Add [`square-`]`pi`/`sigma` icons

danielbayley:


karsa-mistmere (approved):
LGTM!

(Boy am I happy we went with this pi shape, some others would not have fit so snugly within square. 😄 )

ericfennis (approved):

PR #1281, opened 2023-05-16
Improve `qr-code` icon metadata

danielbayley:


danielbayley:
> The alt doesn't really look like a QR code to me

I actually came across it in the wild, and thought it was neat… But I do like the existing one also, and we probably don’t need 2.

Just the optimisation then…

PR #1283, opened 2023-05-16
Add `play-square` icon

danielbayley:
Generic ‘play video’ icon, especially since https://lucide.dev/icon/youtube will be [removed soon](https://github.com/lucide-icons/lucide/issues/94)…

### DIFF
--- a/icons/play-square.json
+++ b/icons/play-square.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "../icon.schema.json",
+  "tags": [
+    "music",
+    "audio",
+    "video",
+    "start",
+    "run"
+  ],
+  "categories": [
+    "arrows",
+    "multimedia"
+  ]
+}
--- a/icons/play-square.svg
+++ b/icons/play-square.svg
@@ -1,0 +1,14 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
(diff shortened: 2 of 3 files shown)


karsa-mistmere (approved):
Nice one!

jguddas:
A bit left heavy.

danielbayley:
> A bit left heavy.

@jguddas Maybe, but shifting it right 1px also looks off. The only other alternative being to shift it right by half a pixel, knocking it off-grid… Not sure that’s desirable?

Untitled

Artboard 1

karsa-mistmere:
Yeah, being off grid is out of the question.

1st row: originals+PR
2nd row: shifted 1px to the right
3rd & 4th row: size bump proportional to play icon w/ & w/o added rounding as per #1043

image
image

karsa-mistmere:
> What about using circumcenter?
> 
> image

Well, if you can illustrate to me how exactly we would use this while also sticking to a 24px grid, then sure. 😄

jguddas:
> > What about using circumcenter?
> > image
> 
> Well, if you can illustrate to me how exactly we would use this while also sticking to a 24px grid, then sure. 😄

Since when are we so strict on sticking to the grid, why not just make it align well even if it is not snapped to the grid?

jguddas:
> > Since when are we so strict on sticking to the grid, why not just make it align well even if it is not snapped to the grid?
> 
> It's always been and always will be a concern, because not sticking to the grid will result in blurry lines such as this: image

Ahh, good point, for this I don't think it's that bad.

That preview is super helpful, we should think about adding a preview like this to the PR comment.

ericfennis (approved):

PR #1285, opened 2023-05-17
Improve 3D icons metadata

danielbayley:
Related to #1142.

### DIFF
--- a/icons/beef.json
+++ b/icons/beef.json
@@ -8,7 +8,9 @@
     "meal",
     "meat",
     "bbq",
-    "steak"
+    "steak",
+    "3d",
+    "perspective"
   ],
   "categories": [
     "food-beverage"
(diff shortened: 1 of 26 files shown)


karsa-mistmere:
I think these changes would be counterproductive in the long run.

We could maybe add a 3D icons category, but tagging seemingly random icons with perspective and isometric is not helpful when easing icon search.

Unfortunately, to me even a category seems unwarranted: it would seem to suggest that these icons are "a set" that is to be used together, which is not the case - all our icons should aim to be consistent with each other.

danielbayley:
> tagging seemingly random icons with perspective and isometric is not helpful when easing icon search.

@karsa-mistmere Simply just a `3d` tag then?

karsa-mistmere:
> > tagging seemingly random icons with perspective and isometric is not helpful when easing icon search.
> 
> @karsa-mistmere Simply just a `3d` tag then?

Also definitely not in favour _unless_ the icon actually has as 3D use case (e.g. our 3d cursors).

PR #1286, opened 2023-05-17
Sort out `pen`[`cil`]/`edit` icons/add `pencil-ruler` (stationery) icon/improve metadata

danielbayley:
Meant to represent tools/utilities/design/illustration/technical drawing/sketch etc…

Alternatives (including possible alterations to [`pencil`](https://lucide.dev/icon/pencil)…)

Untitled

Artboard 1@2x


jguddas:
Maybe a bit big and maybe a bit hard to point out what the icon is supposed to convey.

danielbayley:
> Maybe a bit big

It’s using exactly the proportions of `ruler` and `pen`[`cil`]…

> a bit hard to point out what the icon is supposed to convey.

> Meant to represent tools/utilities/design/illustration/technical drawing/sketch etc…

ericfennis (commented):
Nice icon! 
Not sure about the name: `tools`, I think we can better name it as `ruler-pencil` or `pencil-ruler`.

danielbayley:
> Not sure about the name: `tools`, I think we can better name it as `ruler-pencil` or `pencil-ruler`.

@ericfennis I think `pencil-ruler`, since the pencil is the front/overlapping element… A similar approach to the naming in #1156.

jguddas:
What about naming it `stationery`?

danielbayley:
> What about naming it `stationery`?

I think `pencil-ruler` is more in line with the general approach to naming. But that should definitely be added to `tags`…

jguddas:
IMO `pencil-eraser` should become the new `pencil` and have an alias as `edit-2`.

`pencil-eraser` and `edit-2` could then be removed.

jguddas:
`pencil-line` should also have the eraser IMO.

danielbayley:
> I think we can have both `pencil` and `pencil-eraser`

Surely it should be `pen` and `pencil` then…?

> > pencil-line should also have the eraser IMO.

> Yes, I wholeheartedly agree, without the eraser, it looks a bit too much like an airplane.

I think @jguddas was referring to what is currently named `edit-3`…?

danielbayley:
@karsa-mistmere @jguddas @ericfennis See latest changes…

ericfennis (approved):

icons/ruler.json
```diff
@@ -8,9 +8,13 @@
...
   "tags": [
     "measure",
+    "centimetres",
     "meter",
     "foot",
-    "inch"
+    "feet",
+    "inches",
+    "units",
+    "stationery"
```
karsa-mistmere, line 17: Maybe this (and also the pencil one?) could have some more tags, i.e. length, width, depth, breadth, dimensions, size, extent, measurements.

karsa-mistmere, line 17:
```suggestion
    "stationery",
    "length",
    "width",
    "depth",
    "breadth",
    "dimensions",
    "size",
    "extent",
    "measurements"
```

icons/pencil-ruler.json
```diff
@@ -0,0 +1,30 @@
...
+    "writing",
+    "stationery",
+    "artist",
+    "measure",
+    "centimetres",
+    "meter",
+    "foot",
+    "feet",
+    "inches",
+    "units"
```
karsa-mistmere, line 22:
```suggestion
    "units",
    "length",
    "width",
    "depth",
    "breadth",
    "dimensions",
    "size",
    "extent",
    "measurements"
```

karsa-mistmere (approved):

karsa-mistmere:
Nice work, @danielbayley!

PR #1293, opened 2023-05-18
Add `squircle`

jguddas:
The now so fashionable shape between a square and a circle.

I made it 19px in size, this makes it .5px off grid, which might be a problem.

Use: Icon shape chooser.

image


karsa-mistmere:
Yeah, on a high DPI display it looks neat, but it's pretty damn blurry at 1x.

image
image

icons/squircle.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "../icon.schema.json",
```
jguddas, line 2:
```suggestion
  "$schema": "../icon.schema.json",
  "contributors": [
    "jguddas"
  ],
```

jguddas:
We could consider unifying the `badge`, `circle` and `square`, currently we have a bunch of different icons for each shape.

icon - 2023-10-28T154433 964icon - 2023-10-28T154711 419check-square (1)check-circle (1)

ericfennis (approved):
I'm ok with this shape icon. 
Looking at the 24px DPI preview it looks fine to me.

jguddas:
> I'm ok with this shape icon. Looking at the 24px DPI preview it looks fine to me.

I change it so it is the same size as the square, before it was 0.5px bigger.

ericfennis:
@jguddas Awesome! 
@karsa-mistmere you also ok with this?

PR #1297, opened 2023-05-19
Optimise/add missing `check`[`-*`] icon/s

danielbayley:
`check-square`[`-2`] was missing, plus these were inconsistent with other `check-`[`*`/`2`] icons, in that the `-2` variation is for when the check extends outside of it’s shape, as opposed to the original where it should be contained inside…

### DIFF
--- a/icons/check-circle-2.svg
+++ b/icons/check-circle-2.svg
@@ -9,6 +9,6 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M12 22c5.523 0 10-4.477 10-10S17.523 2 12 2 2 6.477 2 12s4.477 10 10 10z" />
+  <circle cx="12" cy="12" r="10" />
   <path d="m9 12 2 2 4-4" />
 </svg>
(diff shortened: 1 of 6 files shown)


ericfennis (commented):
Nice icons!

Not sure if we need to switch the `check` and `check-2`. I agree your change should be the how it should be. But the people who are using Lucide upgrading to a new version are getting surprised with a new check icon, not sure if we should do this.

danielbayley:
> Not sure if we need to switch the `check` and `check-2`. I agree your change should be the how it should be. But the people who are using Lucide upgrading to a new version are getting surprised with a new check icon, not sure if we should do this.

So many inconsistencies when you really look at the whole set :/

@ericfennis Ok, switched them back now…

karsa-mistmere:
I think we need better names instead of the `-2` suffixes to begin with. 🤔

danielbayley:
> I think we need better names instead of the `-2` suffixes to begin with. 🤔

@karsa-mistmere Agreed where possible, but that particular can of worms isn’t specific to this PR. I think that should be addressed separately, because a lot of these kind of variants already have this numbered naming scheme as it stands…

PR #1300, opened 2023-05-21
Sort out `sliders`/`settings`/`cog`/`gear` icons

danielbayley:
Following on from #231 and https://github.com/lucide-icons/lucide/pull/987#issuecomment-1515121353 …

Also related to #1225.


danielbayley:
> My two cents:
> 
> * `settings` should be `gear`
> * `sliders` should be `sliders-vertical`
> * `settings-2` should be something like `sliders-large-vertical`
>   (but I still prefer your design, it makes it more obvious that they're sliders, not circles on sticks)
> * `cog` should either remain `cog` or if we want to be semantically correct, it could become `cogwheel`

@karsa-mistmere This should be good to go now… @ericfennis

karsa-mistmere:
I still prefer your original redesign for these. 😅
image

Really sells the `sliders` use case as opposed to "`lollipops-left-right`".

danielbayley:
> I still prefer your original redesign for these. 😅

@karsa-mistmere Ha well I now prefer the simplicity of the original 😅 …and have since noticed it all over the place!

Maybe we should put that to a vote (need that bot again…) cc: @ericfennis @jguddas

PR #990, opened 2023-03-28
Add `open-ai` icon

danielbayley:
open-ai

PR #1301, opened 2023-05-21
Move brand icons to separate package

danielbayley:
See  #94 and #670.

Closes #990.
Closes #1161.


karsa-mistmere:
Yeah, we will probably not move these to a new package, doing so would not solve the underlying issues, nor people keeping to request brand icons because "you also have Facebook and YouTube and Instagram, so why not this?!", nor any guideline and licensing issues. Not to mention, why not also have lucide-brands-react, lucide-brands-vue-next etc. So from my side it's a very definite no.

danielbayley:
I’ll probably just maintain my own then, as some of these I’d want to keep. Maybe the same fork I’m already using…

ItsReddi:
@danielbayley could you explain how to use your fork?
If i am trying it, the brand icons are not included in the build.

Thanks :)

PR #1305, opened 2023-05-23
Optimise `gem` icon/improve metadata

danielbayley:
Explored `gem` variants, for reference…

preview

### DIFF
--- a/icons/gem.json
+++ b/icons/gem.json
@@ -2,12 +2,22 @@
   "$schema": "../icon.schema.json",
   "tags": [
     "diamond",
+    "crystal",
+    "ruby",
+    "jewellery",
     "price",
     "special",
-    "present"
+    "present",
+    "gift",
+    "ring",
+    "wedding",
+    "proposal",
+    "marriage",
+    "rubygems"
   ],
   "categories": [
     "gaming",
-    "money"
+    "money",
+    "development"
   ]
 }
--- a/icons/gem.svg
+++ b/icons/gem.svg
@@ -9,8 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <polygon points="6 3 18 3 22 9 12 22 2 9" />
-  <path d="m12 22 4-13-3-6" />
-  <path d="M12 22 8 9l3-6" />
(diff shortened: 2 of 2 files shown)


ericfennis (commented):
not sure about the change in the icon. There is a reason why the upper inner triangle is not perfect. When making it a perfect triangle, the sharp corner is more below and adding an extra pixel, making the icon denser on 100%. So I'm not in favour of this change.

danielbayley:
> not sure about the change in the icon. There is a reason why the upper inner triangle is not perfect. When making it a perfect triangle, the sharp corner is more below and adding an extra pixel, making the icon denser on 100%. So I'm not in favour of this change.

@ericfennis I see the reason now. Fixed…

icons/gem-2.svg
karsa-mistmere, line 1: Not in favour of adding this, in general, I think we should aim to perfect the icons we have with a singular vision instead of adding similar looking alts, too many of those makes it feel like we can't make up our minds which do we prefer in Lucide.

danielbayley: @karsa-mistmere Fair enough. I left them here for reference: https://github.com/lucide-icons/lucide/pull/1305#issue-1722457803

icons/gem.json
```diff
@@ -2,12 +2,22 @@
...
+    "wedding",
+    "proposal",
+    "marriage",
+    "rubygems"
   ],
   "categories": [
     "gaming",
-    "money"
+    "money",
+    "development"
```
ericfennis, line 21: why is gem part of development?

danielbayley: > why is gem part of development?

@ericfennis [RubyGems](https://rubygems.org)…

ericfennis: aahh

ericfennis (changes requested):
We should keep the design of the gem as it is. The lines should not be parallel, if you look at some diamond images you can see why.

danielbayley:
> We should keep the design of the gem as it is. The lines should not be parallel, if you look at some diamond images you can see why.

@ericfennis Ok, just optimised/added metadata then…

ericfennis (approved):
@danielbayley Thanks, good to go!

PR #1307, opened 2023-05-24
Improve `disc` icons metadata/add alternate CD icon

danielbayley:


danielbayley:
> I'm not quite sure, why not keep `disc-2` as it was and add the new one as `disc-3`?

@karsa-mistmere Done.

ericfennis (approved):

PR #1292, opened 2023-05-18
Increases `ticket` in size as per #1192

karsa-mistmere:
Now has the same base dimensions as `credit-card`.

PR #1312, opened 2023-05-29
Add `ticket-*` icons/improve metadata

danielbayley:
Following on from #1292. Related to #1192.

Use cases should be fairly obvious, or at least discernible from the metadata… But a summary anyway:

* `ticket-check` is booked.
* `ticket-plus` is book/add to basket.
* `ticket-minus` is remove from basket.
* `ticket-x` is cancelled/invalid/expired.
* `ticket-slash` is marked/redeemed/used.
* `ticket-percent` is discount/voucher/offer.
* `ticket-text` is T&Cs (terms and conditions/small print) or just a nice variant, matching the many other existing `*-text` icons… Also seen in the wild.


jguddas (approved):

ericfennis:
hmm not sure about these icons, Do we have real-life usecases?

danielbayley:
> hmm not sure about these icons, Do we have real-life usecases?

I came across `ticket-x` in the wild this weekend on either Booking.com or Airbnb (can’t remember which)… `ticket-slash` (redeemed) and `ticket-check` (booked) seem like a no brainer also. The others less so… We could drop `ticket-plus`/`minus` if you want @ericfennis ?

jguddas:
Adding and removing tickets is also a common use case.

I vote for keeping them all.

icons/ticket-text.svg
karsa-mistmere, line 1: Any use cases for this one?

danielbayley: @karsa-mistmere Forgot to add that to the list above… https://github.com/lucide-icons/lucide/pull/1312#issue-1731133830

> * `ticket-text` is T&Cs (terms and conditions/small print) or just a nice variant, matching the many other existing `*-text` icons… Also seen in the wild.

ericfennis: This is not a solid use case. Also the argument of  just a nice variant, matching the many other existing *-text icons" is to me not a good direction we should go. We should always think of real life use cases for new icons and not add random slashes and text lines because similar icons have that.

PR #1313, opened 2023-05-31
docs: add nuxt-lucide-icons package

JaZo:
We've published [nuxt-lucide-icons](https://www.npmjs.com/package/nuxt-lucide-icons) and it would be nice if it was listed on the packages page.


ericfennis (approved):
Nice cool!

PR #1315, opened 2023-05-31
Add icon headset

KayleyHill:
callcenter-headset

I am probably putting this in the complete wrong area, sorry it is my first pull request. I made this one using the headphones icon as a basis.


ericfennis:
@KayleyHill It looks like something went wrong. There are no changes. You sure you pushed the icon?

KayleyHill:
This was my first attempt before I made the topic in Issues channel. Not really sure how to do this properly yet. Also tbh @ericfennis this version is a bit poor, the one that @jguddas made was much better and should be the one that goes live.

ericfennis:
Check will wait for @jguddas to create a PR then.

PR #995, opened 2023-03-29
Add missing `git` icons

danielbayley:
* Added `git-commit-vertical` (and subsequently renamed `git-commit`[`-horizontal`])
* Added `git-pull-request-create`
* Refined `git-fork` to better match [`network`](https://lucide.dev/icons/network)
* Refined `git-pull-request-closed` to better match other `*-x` icons (as most `x`es are 5x5 px, not 6x6)
* Added `git-`[`pull-request-`[`create`/`compare`]`-arrow`/`s` variants, only making them slightly wider for better consistency than just scaling them up on both axes.
* Added `git-graph` icon (maybe should be `git-commit-graph`?) like in VS Code…

Untitled
<br/>
<br/>

Full preview of changes in https://github.com/lucide-icons/lucide/pull/995#issuecomment-1509285280.

PR #1319, opened 2023-05-31
Add browser/dev tools icons

danielbayley:
Inspired by Mozilla/Firefox/dev tools. For reference:

Screenshot 2023-06-01 at 7 42 07 pm
<br/>
<br/>

Related to #1039.

Edit: Leaving removed `applications` and `computer-chip` icons here for reference:

preview


icons/memory.svg
karsa-mistmere, line 1: Not in favour of this, as it doesn't match our CPU icon very well, and if we add something like this it should probably have a more generic name, e.g. `computer-chip`.

icons/vpn.svg
karsa-mistmere, line 1: Doesn't look very VPN-y to me, and I'm not sure about use cases for this either, can you provide some?

icons/layout-grid-2.svg
karsa-mistmere, line 1: Not very consistent with our other layout icons, I think this is simply not feasible in Lucide.

icons/inspector.svg
karsa-mistmere, line 1: Not sure about the name for this; it's a pretty generic shape with a very specific name (our current `inspect` would also fare better as `mouse-pointer-square`), I think even something like `inspect-square` would be somewhat better, but I'm not sold on that either.

icons/inspect-2.svg
karsa-mistmere, line 1: Back in a while, we've very deliberately removed the stem to be more consistent with icons where it wasn't feasible, not sure about adding it back as an alt.

jguddas: I like the simpler version without a stem.

icons/inspector.svg
```diff
@@ -0,0 +1,17 @@
...
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect width="14" height="14" x="5" y="5" rx="2" />
+  <path d="M12 5V2" />
+  <path d="M19 12h3" />
+  <path d="M12 22v-3" />
+  <path d="M2 12h3" />
```
jguddas, line 16: What about just calling it `scope` and adding a dot in the middle?

icon

```suggestion
  <path d="M2 12h3" />
  <path d="M12 12h0.01" />
```

icons/tablet-smartphone.svg
```diff
@@ -0,0 +1,15 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M5 4a2 2 0 0 1 2-2h12a2 2 0 0 1 2 2v16a2 2 0 0 1-2 2h-2" />
+  <path d="M8 18h.01" />
+  <path d="M11.5 8c.9 0 1.5.6 1.5 1.4v11.2a1.37 1.37 0 0 1-1.4 1.4H4.4A1.37 1.37 0 0 1 3 20.6V9.4A1.37 1.37 0 0 1 4.4 8Z" />
```
jguddas, line 14:
```suggestion
  <rect x="3" y="8" width="10" height="14" rx="2" />
  <path d="M5 4a2 2 0 0 1 2-2h12a2 2 0 0 1 2 2v16a2 2 0 0 1-2 2h-2.4" />
  <path d="M 8 18h0.01" />
```

icons/vpn.svg
jguddas: We could call this `path` or something and have the use case include routing for maps and stuff.

danielbayley: > Doesn't look very VPN-y to me, and I'm not sure about use cases for this either, can you provide some?

> We could call this `path` or something and have the use case include routing for maps and stuff.

Yeah it’s meant to represent indirection/rerouting/proxy… Loosely based on  https://lucide.dev/icons/lock?search=network/`share-2`/`workflow`… I guess maps is another use case as well.

We could call it `indirect` or `reroute`? @karsa-mistmere @jguddas

icons/inspector.svg
danielbayley: > What about just calling it `scope` and adding a dot in the middle?

Already something similar in #1338.

icons/layout-grid-2.svg
danielbayley: @karsa-mistmere I get that it breaks perfect consistency with other `layout-grid-*` icons, but it does actually fit the constraints in general… Could call it `applications`, or something else… or just discard it completely?

icons/memory.svg
danielbayley: > it should probably have a more generic name, e.g. `computer-chip`

Renamed.

> it doesn't match our CPU icon very well

@karsa-mistmere Hmm, some alternatives, based more on [`cpu`](https://lucide.dev/icons/cpu)…

preview

danielbayley:
The preview bot seems to have stopped updating 😕 https://github.com/lucide-icons/lucide/pull/1319#issuecomment-1571040840

ericfennis (changes requested):
I think some icons we already have. Like: `cpu` and `applications`. Also `crosshair` or `locate` could work.

So I think for this PR we should go for:
- `mouse-pointer-square`
- `mouse-pointer-square-dash` 
- `tablet-smartphone`
- `waypoints`

Can you describe the use case for `waypoints`?

danielbayley:
> I think some icons we already have. Like: `applications`.

Where? https://lucide.dev/icons/grid?search=app

> Also `crosshair` or `locate` could work.

Moved to #1338.

> Can you describe the use case for `waypoints`?

@ericfennis As mentioned https://github.com/lucide-icons/lucide/pull/1319#discussion_r1220075343, it was meant to represent indirect/rerouting/proxy/VPN… and is loosely based on https://lucide.dev/icons?search=network /`share-2`/`workflow`… But is also versatile enough to include `maps`/`navigation`/orienteering/path etc… Especially if e.g there are multiple stops on the route, the icon would be perfect to represent, say a journey planner… It would go nicely with https://lucide.dev/icons/git-commit (and `git-commit-vertical`) from https://github.com/lucide-icons/lucide/pull/995#issuecomment-1509285280.

karsa-mistmere (approved):

jguddas (approved):

ericfennis (approved):

PR #1320, opened 2023-06-01
Add `candlestick-chart` icon/`bar-chart-big` variants

danielbayley:


jguddas (commented):
Love the candlestick-chart, not convinced by the big variants.

danielbayley:
> Nice icons @danielbayley ! Can you post a design of the barcharts with the smaller bar reduced in size. Maybe it makes it more readable on 100%.

Done…

karsa-mistmere:
I think it's better, but maybe we could easily go even shorter?
image

danielbayley:
> I think it's better, but maybe we could easily go even shorter?

@karsa-mistmere Ok, so now they proportionally match the https://lucide.dev/icons/?search=align icons…

karsa-mistmere:
> > I think it's better, but maybe we could easily go even shorter?
> 
> @karsa-mistmere Ok, so now they proportionally match the https://lucide.dev/icons/?search=align icons…

Ooooh, nice catch on consistency! Yes, it's perfect this way. 😊
image

karsa-mistmere (approved):

PR #1326, opened 2023-06-01
Add `percent-*` (discount price tag) icons/improve metadata

danielbayley:
Came across `percent-circle` in the wild the other day… Although others seem more appropriate representations of discount/price  tag…

### DIFF
--- a/icons/badge-percent.json
+++ b/icons/badge-percent.json
@@ -8,12 +8,17 @@
     "verified",
     "unverified",
     "sale",
-    "marketing"
+    "discount",
+    "offer",
+    "marketing",
+    "sticker",
+    "price tag"
   ],
   "categories": [
     "account",
     "social",
     "money",
+    "shopping",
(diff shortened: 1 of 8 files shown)


jguddas:
The percentage sign inside tag isn't centered.

ericfennis:
I'm not sure about this one, I get sort of the use case, it's the same as `percent-badge`. But I'm why we should bloat the library with all kinds of percent icons. I think we should a bit more selective.

danielbayley:
> all kinds of percent icons. I think we should a bit more selective.

@ericfennis Yeah maybe we don’t need them all, but most symbols in the library have `-square` and `-circle` variants… Although I think `percent-diamond` is the most relevant one. So since `percent-tag` doesn’t really sit very well, as @jguddas pointed out, I have dropped it…

ericfennis (approved):

PR #1327, opened 2023-06-01
Add `presentation`/`projector` icons

### LINKED ISSUE #1303: Whiteboard icon
## Icon Request

* Icon name: Whiteboard
* Use case: To use in apps with a whiteboard or drawing canvas
* _Screenshots_ of similar icons:  https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcR8qaDhTkDlrOMdt4MRYsWpX_x7HxKt7ExLDA&usqp=CAU

danielbayley:
Closes #1303.

See also alts:

preview


jguddas:
Okay, now those are some cute icons ♥️

karsa-mistmere:
Love the projector!

For presentation screen I prefer the second one, I think if we add rounding, it would feel more natural to have it on the bottom side:
image

danielbayley:
> For presentation screen I prefer the second one, I think if we add rounding, it would feel more natural to have it on the bottom side:

@karsa-mistmere Well the second one is also meant to be a whiteboard… where the little shelf for the marker pens is, hence that being at the bottom. The other one is the pull-down screen to go with the `projector`, based on https://lucide.dev/icons/blinds.

karsa-mistmere:
> Well the second one is also meant to be a whiteboard… where the little shelf for the marker pens is, hence that being at the bottom.

I think it's fine to cover both use cases in a single icon.

> The other one is the pull-down screen to go with the `projector`, based on https://lucide.dev/icons/blinds.

Yeah, that one I'm not totally sure about, there's not enough gap between the handle and the screen and this makes it looks really dense.

> @karsa-mistmere Are you going to add those, based on these in a separate PR?

Sure. :)

danielbayley:
> there's not enough gap between the handle and the screen and this makes it looks really dense.

> cover both use cases in a single icon.

@karsa-mistmere I think it was fine, but merged anyway…

PR #1331, opened 2023-06-03
Sort out `loader`/cursor icons

danielbayley:


icons/loader-circle.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "../icon.schema.json",
```
jguddas, line 2: I think you need to add this if you move things but @karsa-mistmere probably knows more.
```suggestion
  "$schema": "../icon.schema.json",
  "aliases": [
    "loader-2"
  ],
```

karsa-mistmere:
Not a fan of adding all kinds of states of the loader, we might as well add different rotations of the circle loader, but there are much better ways of animating than changing icons.

icons/loader-pinwheel.json
```diff
@@ -0,0 +1,19 @@
+{
```
jguddas, line 1: I think it should be `pinwheel-loader`.

jguddas: No, I just checked, the group goes in the front, so you are 100% correct.

icons/loader-circle.json
danielbayley: > I think you need to add this if you move things but @karsa-mistmere probably knows more.

@jguddas I thought about that, but not sure in cases where the alias name is another existing icon…? @karsa-mistmere

danielbayley:
> Not a fan of adding all kinds of states of the loader, we might as well add different rotations of the circle loader, but there are much better ways of animating than changing icons.

@karsa-mistmere `revert`ed…

icons/loader-circle.json
karsa-mistmere: @jguddas is right, we need aliases for backwards compatibility. And you've renamed `loader-2` to `loader-circle`, so `loader-2` used to be _this_ icon, not another one. :)

icons/loader-circle.json
```diff
@@ -7,10 +7,21 @@
...
   ],
   "categories": [
+    "cursors",
     "multimedia",
-    "layout"
+    "layout",
+    "design"
+  ],
+  "aliases": [
+    "loader-2"
```
ericfennis, line 25: I'm in renaming this one and adding an alias. But we can get in to some problems with renaming icons with `-2` on the end. If we add it as an alias we can never make a `loader-2` icon, because the name is already occupied by an alias. This is something we can think of if go to v1.

danielbayley: @ericfennis Ok, so keep it as `loader-2` for now…?

ericfennis: Yess unfortunately the best here.

danielbayley: @ericfennis Ok, done.

karsa-mistmere:
I'm not a big fan of `loader-circle-big`.

`loader-pinwheel` could maybe be `circle-pinwheel` instead? 🤔 

`mouse-pointer-ban` is super useful though!

danielbayley:
> `loader-pinwheel` could maybe be `circle-pinwheel` instead? 🤔

@karsa-mistmere Shouldn’t `loader-*` icons be grouped by prefix like others? Or maybe `cursor-pinwheel`? If not, maybe even `spinning-pinwheel`? See: https://en.wikipedia.org/wiki/Spinning_pinwheel.

jguddas:
- circle-big: weird name but okay with me.
- pinwheel: needs some path optimization

PR #1332, opened 2023-06-03
Add `wallet-cards` icon/`wallet` variant/optimise `credit-card` icon/improve metadata

danielbayley:
`wallet-cards` curvature matches the `inbox-*` icons in #998.

Related to #1350.


ericfennis (approved):
Nice work!

icons/wallet-2.svg
karsa-mistmere, line 1: Not sure about this one, the sizes should be consistent between the two wallets:
image

Obviously, the current `wallet` is the one that's too small, I'd say we'd better fix that.

danielbayley: > the sizes should be consistent between the two wallets:

@karsa-mistmere Not sure how it could have been any more consistent than this…

Untitled

> Obviously, the current `wallet` is the one that's too small, I'd say we'd better fix that.

Fixed them both…

karsa-mistmere (approved):

Issue #1334, opened 2023-06-05
Add cryptocurrencies icons


AirOne01:
Related PR at #1335

## Icon Request

Icons of some of the most popular cryptocurrencies.
**I already made those icons**. This is just an Issue to keep track of them and get feedback. Feel free to tell me anything that can be enhanced or changed. If this issue is useless, ping me and I'll remove it 👍.

* Icon names:

| Codename | Name |
| --- | --- |
| ADA | Cardano |
| ETH | Etherum |
| MATIC | Polygon |
| SOL | Solana |
| USDC | USD Coin |
| USDT | Tether |
| XRP | XRP |

* Use case: Anything involving the most popular cryptocurrencies
* _Screenshots_ of similar icons:

| | |
| --- | --- |
| ![image](https://github.com/lucide-icons/lucide/assets/21955960/8daa8d0f-5a78-4860-8fa0-e7c35c1d2649) | ![image](https://github.com/lucide-icons/lucide/assets/21955960/64fd9646-21e7-45bd-8262-b23abfd5cd61) |
| ![image](https://github.com/lucide-icons/lucide/assets/21955960/fab60806-9811-4e8d-83b2-92921a3077a4) | ![image](https://github.com/lucide-icons/lucide/assets/21955960/44e82e88-3c71-4584-a46c-d45e7cf2c6ba) |
| ![image](https://github.com/lucide-icons/lucide/assets/21955960/e39f4159-31d9-4273-9e08-c7bb9b0d6d4f) | ![image](https://github.com/lucide-icons/lucide/assets/21955960/84a355f8-3bd7-4923-a652-367486a89030) |
| ![image](https://github.com/lucide-icons/lucide/assets/21955960/230a127d-4743-4d08-8b8b-a127021a42eb) | |

## What I made:

![My designs on Figma](https://github.com/lucide-icons/lucide/assets/21955960/5815c867-3086-47c4-97cd-014c7c205a49)

### Notes:

- The dogecoin sucks, and it will not be part of my PR.
- USD Tether uses the subtract tool in Figma (results in `clip-rule="evenodd"`). There is no mention of this being bad for what I know but I just thought I'd mention it.
- Polygon (MATIC) is a normal english word used for a cryptocurrency. Should I make it more obvious that we're not talking about the shape ?
- Cardano uses `fill="none" stroke="currentColor"` which does not respect the [icon design guide](https://lucide.dev/guide/design/icon-design-guide). I kept it because it can arguably be ok.

[...]

PR #1335, opened 2023-06-05
Add cryptocurrencies icons

AirOne01:
Solves #1334 

Worth noting:

- The dogecoin sucks, and it will not be part of my PR.
- USD Tether uses the subtract tool in Figma (results in `clip-rule="evenodd"`). There is no mention of this being bad for what I know but I just thought I'd mention it.
- Polygon (MATIC) is a normal english word used for a cryptocurrency. Should I make it more obvious that we're not talking about the shape ?
- Cardano uses `fill="none" stroke="currentColor"` which does not respect the [icon design guide](https://lucide.dev/guide/design/icon-design-guide). I kept it because it can arguably be ok.

You can find my Figma board [here](https://www.figma.com/file/cWJVxf9Kc19LcGpANpj5Ah/Lucide?type=design&node-id=0%3A1&t=UI5ZhfFc5LPw5XMh-1).


AirOne01:
Tether and Cardano both look glitched on my setup. I'll take a look when I come back home. Ping me for anything. 👍

AirOne01:
#670

PR #1339, opened 2023-06-05
Improve `undo-2` metadata

danielbayley:


jguddas:
How is this different from a flipped undo?

karsa-mistmere:
Yes, this is very similar to https://lucide.dev/icons/undo-2

danielbayley:
> Yes, this is very similar to https://lucide.dev/icons/undo-2

@karsa-mistmere @jguddas Oh I didn’t see that! Nothing for `back` shows up in search…

It is a different, more squared design… But if you prefer, this PR could instead fix the missing tags that led to it…?

icons/undo-2.json
```diff
@@ -8,7 +8,10 @@
   "tags": [
     "redo",
     "rerun",
-    "history"
+    "history",
+    "backwards",
```
jguddas, line 12: `back`, `revert`, `u-turn`

danielbayley: > `back`, `revert`, `u-turn`

@jguddas I've been operating under the assumption that e.g a tag of `backwards` would negate the need for the substring `back` as separate tag, in the search…? @karsa-mistmere

karsa-mistmere: For searching, yes, although I think actually simply `back` would suffice, since it's the tag that all other left pointing arrow icons have.

ericfennis (commented):
@danielbayley Can you remove the `arrow-back` icon in this PR. Since we already have `undo-2`. Then it is good to go.

ericfennis (approved):

PR #1341, opened 2023-06-05
Add TV/controller/`connector` icons

danielbayley:
`triangle-right` related to #1142.

Edit: Leaving other `cable`[`-*`] icons here, for reference…

preview


jguddas:
I like the `antenna` :+1:

icons/antenna.json
```diff
@@ -0,0 +1,29 @@
...
+    "live",
+    "frequency",
+    "tune",
+    "scan",
+    "channels",
+    "aerial",
+    "receiver",
+    "transmission",
+    "transducer",
+    "terrestrial"
```
jguddas, line 22: I'm new to tagging, maybe `"satellite"` could be added as well?

danielbayley:
Wondering if `cord-*` would be a better name than `connector-*`?

karsa-mistmere:
> Wondering if `cord-*` would be a better name than `connector-*`?

Definitely, or maybe even `cable-*`

> @karsa-mistmere Nice, I didn’t realise that was on the table… pushed.

It isn't, it's just something that I wish was. 😆

ericfennis:
@danielbayley I think it's not really clear what the exact purpose is of the `cable-audio-video`. I think it's better to keep It more generic. On 100% scale, it is really to distinguish them from each other.

And I'm not sure about renaming de `tv-2` to `tv-flatscreen`. Yes, this is maybe a better name. But we can't really add an alias for this because then we can never add a `tv-2` icon in the future. And I think we should be more carefull with renaming things without adding "backward compatibility". I think we should keep it as it is.

icons/hdmi-port.svg
```diff
@@ -0,0 +1,14 @@
+<svg
```
jguddas, line 1: Feels a bit small and is hard to identify as a port.
What about something like this?
icon (87)

icons/hdmi-port.svg
```diff
@@ -0,0 +1,14 @@
...
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M22 9c0-.6-.4-1-1-1H3c-.6 0-1 .4-1 1v4c0 .6.4 1 1 1h1l2 2h12l2-2h1c.6 0 1-.4 1-1Z" />
```
jguddas, line 12:
```suggestion
  <path d="M22 9a1 1 0 0 0-1-1H3a1 1 0 0 0-1 1v4a1 1 0 0 0 1 1h1l2 2h12l2-2h1a1 1 0 0 0 1-1Z" />
```
`C`urves should be `A`rcs.

icons/ratio.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "../icon.schema.json",
+  "contributors": [
+    "danielbayley"
+  ],
+  "tags": [
+    "screens",
```
jguddas, line 7:
```suggestion
    "screens",
    "rotate",
    "rotation",
```

icons/triangle-right.json
```diff
@@ -0,0 +1,22 @@
+{
```
jguddas, line 1: :-1: don't like it.

icons/tv-flatscreen.json
```diff
@@ -0,0 +1,29 @@
+{
+  "$schema": "../icon.schema.json",
+  "contributors": [
```
jguddas, line 3:
```suggestion
  "aliases": [
    "tv-2"
  ],
  "contributors": [
```

icons/tv-flatscreen.json
```diff
@@ -0,0 +1,29 @@
...
+    "ericfennis"
```
jguddas, line 4: Are the contributors correct?

jguddas:
My comment about the cables got lost, in short, I vote for only keeping the `cable-plugs` and naming the icon `cable`.

icons/tv-flatscreen.json
danielbayley: > Are the contributors correct?

@jguddas No idea 🤷🏻‍♂️😅 I think `contributors` really needs to be something (possibly fully?) automated. Which I think #1343 is tracking…

icons/triangle-right.json
danielbayley: > 👎 don't like it.

@jguddas Maybe useful to have _slightly_ more constructive feedback than that 😅

Do we not want to cover all base `shapes`…? Should be part of #1142?

danielbayley:
> not sure about renaming de `tv-2` to `tv-flatscreen` … I think we should keep it as it is.

@ericfennis `revert`ed.

icons/tv-flatscreen.json
ericfennis: Yes this one is correct, I created the `tv-2` icon originally.

ericfennis:
I agree with @jguddas . Maybe one `cable` icon will fulfill the use-case. On 100% scale it is hard to see if it is a power cable or USB. For example if you want to indicate that something is going on with your power cable I think `plug` or `plug-zap` is a better fit for that.

danielbayley:
> I vote for only keeping the `cable-plugs` and naming the icon `cable`.

> Maybe one `cable` icon will fulfill the use-case.

@jguddas @ericfennis `push`ed…

karsa-mistmere (approved):

PR #1343, opened 2023-06-05
Optimizing contributors

karsa-mistmere:
Our current contributors format doesn't differentiate between designers, optimizers and people who edited the metadata (tags, categories etc) who are currently (mostly) uncredited.

I've done some very heavy, mostly manual lifting on the current metadata, including hunting down original designer usernames from Lucide and Feather PRs and issues and I split the current contributors lists up into three categories as stated above.

I don't think using this data as it currently is would be very feasible, as updating it would be terribly cumbersome, both manually (these new JSONs are quite a handful) and automatically (adding to the metadata contributors would add you to the metadata contributors 🤯 ).

I see three possibilities to make things more manageable:
1) we take the designers and drop all other contributors and keep things tidier in the future (in practice a single designer could be manually added every time an icon is added)
2) we drop just the metadata contributors to avoid clutter and having to take on the task of keeping score of _every single commit_ that's a non-contribution (metadata JSON format refactors, batched upkeep tasks, editing of contributors etc that are still contributions, obviously, just not directly related to the icon itself).
3) we move all of these to some kind of separate file, e.g. `[iconName].contributors.json`, that way automatically updating them would result in less conflicts and maintenance, as changes made to these files could be ignored.

Let me know, what you think, at the present moment I think I'm mostly in favour of solution 1, as it would result in the least amount of manual maintenance needed.

---

**Sidenote:** this whole issue stems from me wanting to remove myself from the contributors list of a lot of icons that I did not in fact contribute to, but was nonetheless added, e.g. all icons in #1042, just because I added `interpunct` as a tag to a single icon in this PR which according to the original plans shouldn't even have warranted any contribution whatsoever. 

I could remove myself manually, but every time we'd update the contributors, I'd be added back in.


jguddas:
I still vote for keeping the status quo.

ericfennis:
This one is a bit outdated

PR #1344, opened 2023-06-06
Add VR/AR icons

danielbayley:
`focus-box` also doubles as `symbol-variable` from VS Code/[codicons](https://microsoft.github.io/vscode-codicons/dist/codicon.html).


ericfennis:
I agree with @karsa-mistmere on this one. Without knowing the context it is hard to see what icon is trying to tell.

danielbayley:
They look obvious enough to me, especially the controller! But then maybe I am heavily biased, as a long time Oculus Quest user… I have some variations I’ll post…

xnousnow:
What about making vr headset design like Apple Vision pro?

karsa-mistmere:
> Example icons at #1421 would be better to understand more.

I think those icons all look very much like scuba diving masks, personally, VR would never come to mind.

ericfennis:
I agree with @karsa-mistmere and @xnousnow. These need some refinement.
Please continue in thread #1421 for feedback on designs.

PR #1346, opened 2023-06-06
Lock icons

cd16b:


danielbayley:
@cd16b Nice! I had something very similar drafted… Just with a slightly more keyhole shape:

lock-2
unlock-2

ericfennis (approved):
Nice like these!

jguddas:
What is the use case for this icon that is currently not covered by `lock`?

karsa-mistmere:
Not really in favour of adding these as separate icons. Either let's change the OG lock icons or drop these, I think they're way too similar to be a separate set.

karsa-mistmere:
@ericfennis: I've reviewed this again, upon further reflection I'm okay with adding these as alts just as long as the base lock shape is the same in both icons (i.e. the core icons are also updated), although I think I'd rather stick to either `lock-dot` or `lock-keyhole` not both:
image

ericfennis:
@karsa-mistmere awesome. I think we should go for simple and go for the `lock-dot`. Since on 100% scale you probably won't see the keyhole that well.

icons/lock-dot.svg
```diff
@@ -0,0 +1,15 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="12" cy="16" r="1"/>
+  <rect width="18" height="12" x="3" y="10" rx="2"/>
+  <path d="M7 10V7a5 5 0 0 1 10 0v3"/>
```
jguddas, line 14: Formatting
```suggestion
  <circle cx="12" cy="16" r="1" />
  <rect x="3" y="10" width="18" height="12" rx="2" />
  <path d="M7 10V7a5 5 0 0 1 10 0v3" />
```

icons/unlock-dot.svg
```diff
@@ -0,0 +1,15 @@
...
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <circle cx="12" cy="16" r="1"/>
+  <rect width="18" height="12" x="3" y="10" rx="2"/>
+  <path d="M7 10V7a5 5 0 0 1 9.33-2.5"/>
```
jguddas, line 14: Formatting
```suggestion
  <circle cx="12" cy="16" r="1" />
  <rect x="3" y="10" width="18" height="12" rx="2" />
  <path d="M7 10V7a5 5 0 0 1 9.33-2.5" />
```

karsa-mistmere:
> @karsa-mistmere awesome. I think we should go for simple and go for the `lock-dot`. Since on 100% scale you probably won't see the keyhole that well.

image

I think it works OK, definitely sells the illusion of a keyhole, and even scaled up it looks pretty decent:
image

ericfennis:
@karsa-mistmere True, I'm okay with both ways. I think the dot is a more simplistic design that fits with the lucide style. But the keyhole is more recognizable and more useful when scaled up.

ericfennis:
@cd16b Can you update the names to `lock-keyhole`? probably a more logical name.
And I'm still open to changing back to the keyhole designs instead of the dot.

PR #1348, opened 2023-06-07
General improvements in documentation

ericfennis:
I did some overall improvements to the documentation, which make it a bit better to read. I also used more of the Vitepress Markdown extensions to improve the pages.


karsa-mistmere:
Angular was left out, I added the same changes there.

PR #1224, opened 2023-05-06
feat: defention of done

jguddas:

PR #1349, opened 2023-06-07
Optimise/add `shield-*` icons/improve metadata

danielbayley:


ericfennis (changes requested):
Hmm I not sure about lot of them. They look really nice but I think some of them are a bit out of scope for Lucide. The different shields patterns are too specific for the lucide icons set. Also, some of them have religious symbols and I think we should not add those to the Lucide library.

For the others, I placed some question marks because it is not clear what the use cases are.
image

jguddas:
> Also, some of them have religious symbols and I think we should not add those to the Lucide library.

Every day there is something new to add to https://github.com/lucide-icons/lucide/pull/1224.

danielbayley:
Some of the ones you marked already exist! 😅 I optimised and added fairly comprehensive metadata (which ideally would explain use cases…)

@ericfennis So I removed the sigil ones—I thought they might be useful for `gaming`… Which is a similar use case for `shield-plus`/`minus` (up/downgrade armour etc.), but also to indicate increased protection (security settings/antivirus etc…)

* `shield-question` is unverified (and already exists)
* `shield-half` (could have called it `shield-2`, but trying to go for better names…) is a common variant of just `shield`
* `shield-more` is scanning (for malware/antivirus etc…)
* `shield-ban`/`slash` is a similar subset of use cases as https://lucide.dev/icons/ban, https://lucide.dev/icons/circle-slash, but in more of a security context, e.g removed/banned for security breaches, virus blocked etc…

jguddas (approved):

ericfennis (commented):
What's the usecase for `shield-minus` and `shield-more` ?

ericfennis (approved):

PR #1350, opened 2023-06-08
Add format/data/`card` icons

danielbayley:
Related to #1332.


karsa-mistmere:
1. Use cases are missing for all added icons, especially for `slot-*` and floppy disk alts.
2. Not a fan of the `card-*` naming scheme, since these items have practically nothing to do with each other apart from having a similar shape, since we definitely do not want to group them, I think `credit-card`, `sd-card` and `sim-card` should be fine as  names.

ericfennis:

PR #1354, opened 2023-06-10
Add `barcode`/`scan` icons/optimise/improve metadata

danielbayley:
Loosely related to #1344.


jguddas (approved):
It's a violation :disappointed:, but it looks sooooooo good.

danielbayley:
> It's a violation 😞, but it looks sooooooo good.

Oh no! 😭 Is it really? How?

karsa-mistmere:
> > It's a violation 😞, but it looks sooooooo good.
> 
> Oh no! 😭 Is it really? How?

Yeah, I'm afraid the 1px wide rectangles aren't something we have in any other icons, and they don't scale well:
image

danielbayley:
Such a shame! Fixed now anyway… @karsa-mistmere @jguddas

jguddas:
not sure `scan` is the right word

danielbayley:
> not sure `scan` is the right word

@jguddas  It’s the [existing](https://lucide.dev/icons/?search=scan) naming… What else should it be?

ericfennis (approved):
Nicee ones!!

jguddas (approved):

Gobot1234:
Hi sorry for the slight necro but didn't want to open a new issue. After staring at the barcode scan it seems that the margin isnt the same on either side with the lines. Was this intentional?

danielbayley:
> After staring at the barcode scan it seems that the margin isnt the same on either side with the lines. Was this intentional?

@Gobot1234 Of course. The uneven gap between lines is what makes it idetifiable as a barcode… Not ideal in this case that means the margins also are not even, but constraints are tight!

PR #1356, opened 2023-06-10
Add `activity-square` icon

danielbayley:


ericfennis (approved):
Nice icon!

jguddas (approved):
Love it :heart:

karsa-mistmere:
It feels a bit skewed, maybe something like this would have been better?
image
image

danielbayley:
> It feels a bit skewed, maybe something like this would have been better?

@karsa-mistmere I can’t tell, is that .5px difference? Tweak it in another PR if you like…

PR #1359, opened 2023-06-12
Add VS Code Lucide SVG snippets

danielbayley:
Helps with creation/optimisation of icons in VS Code, by adding some common basic shapes—with proper rounding—as snippets. Also to encourage best practices, like e.g using `path` vs [`poly`]`line`. This could be further expanded with other common elements, e.g `arrow`/`chevron` etc…

Related to #1224.


ericfennis (approved):
LGTM!

PR #1289, opened 2023-05-17
Add fast/junk/food/snack icons

danielbayley:
Related to #1149, and #1306.

PR #1149, opened 2023-04-24
Add `candy-cane`/`lollipop`/`popsicle`/`dessert`/`popcorn` icons

danielbayley:
Related to #1289, and #1306.

Alts:
Untitled

`cake-slice` alts:

preview

PR #1372, opened 2023-06-17
Add shipping/`container` (Docker) icons

danielbayley:
Shipping/Docker `container` like from [octicons](https://primer.style/design/foundations/icons/container-16). Reuses `cuboid` (brick) from #1142.

`arrows-up-from-line` commonly found on boxes to indicate package orientation. Often paired with a cracked glass icon to indicate fragile… See also `goblet-crack` in https://github.com/lucide-icons/lucide/pull/1374#issuecomment-1595896191.

### DIFF
--- a/icons/arrows-up-from-line.json
+++ b/icons/arrows-up-from-line.json
@@ -1,0 +1,23 @@
+{
+  "$schema": "../icon.schema.json",
+  "contributors": [
+    "danielbayley",
+    "ericfennis"
+  ],
+  "tags": [
+    "direction",
+    "orientation",
+    "this way up",
+    "vertical",
+    "package",
+    "box",
+    "fragile",
+    "postage",
+    "shipping"
+  ],
+  "categories": [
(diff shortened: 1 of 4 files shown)


jguddas:
What about something like this? icon (90)

danielbayley:
> What about something like this? icon (90)

Kind of looks like a cupcake? 😅

jguddas:
> > What about something like this? icon (90)
> 
> Kind of looks like a cupcake? sweat_smile

Yeah :laughing:, if you have a cool cupcake design, we can add it to #1289.

danielbayley:
> if you have a cool cupcake design, we can add it to #1289.

I actually don’t! But that would be more at home in https://github.com/lucide-icons/lucide/pull/1149#issuecomment-1520066481 …maybe you could adapt this and `commit` there? @jguddas

ericfennis (approved):
Nice I like them!

PR #1378, opened 2023-06-21
Refine `file`/`sticky-note`[`-*`] icons/add `file-dot` (dotfile) icon

danielbayley:
Current `file`[`-*`] icons don’t make any sense, in that while the fold is a sharp corner, the rest of the file shape has 2px rounding. Also, the folded corner is currently 0.5px off-grid… This PR fixes these issues, while also freeing up a little extra space internally for other elements.


karsa-mistmere:
When scaled up the corners look a bit funky, although the current icon also suffers somewhat from this issue.
Not totally sure, but maybe we could also add a 2px rounding to the top right corners?

image

danielbayley:
> When scaled up the corners look a bit funky, although the current icon also suffers somewhat from this issue. Not totally sure, but maybe we could also add a 2px rounding to the top right corners?

@karsa-mistmere I would say that if you were to fold it in real life, the corner will be straight across… But as you’ve illustrated, none of them are currently strictly accurate…

Not sure what happened to the preview bot here 🤷🏻‍♂️

ericfennis:
@danielbayley @karsa-mistmere I redesigned these ones before and made the little triangle not align on the corners on purpose. That was because the little triangle looked too dense on 100% scale, so shifting it a bit It looked a bit bigger. But rounding the corners could be a nice solution.

Looks like the preview bot has a max of icons he can process 🤔

karsa-mistmere:
image
image
image

100% & 200% views of file outlines w/ 5×5px corner and 2px rounding at each point.

icons/file-badge-2.svg
```diff
@@ -9,7 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M14.5 2H6a2 2 0 0 0-2 2v16a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V7.5L14.5 2z" />
-  <path d="M12 13a3 3 0 1 0 0-6 3 3 0 0 0 0 6Z" />
+  <path d="M15 2H6a2 2 0 0 0-2 2v16a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V7Z" />
+  <circle cx="12" cy="10" r="3" />
```
jguddas, line 13:
```suggestion
  <circle cx="12" cy="10" r="3" />
  <path d="M14 2v4a2 2 0 0 0 2 2h4" />
```

ericfennis:
Can we have the use-case for `file-dot`?

ericfennis (approved):

danielbayley:
> Can we have the use-case for `file-dot`?

@ericfennis Missed this comment! Use case is to represent a `.`dotfile, of course…

PR #1013, opened 2023-04-04
Add `arrow-*-circle` icons (including `esc`ape key)

danielbayley:
<kbd>esc</kbd>

Related to #1002.

PR #1383, opened 2023-06-22
Add missing `arrow-*-from-square`/optimise `external-link` icon/s

danielbayley:
Following on from #1351, and closely related to #1013.


ericfennis (approved):
@danielbayley Nice ones !

karsa-mistmere (approved):

jguddas:
Should we remove the -from, and make the name shorter?

karsa-mistmere:
> Should we remove the -from, and make the name shorter?

I vote against, I think we should leave space for matching `arrow-down-left-to-square` icons, should the need arise:
image

danielbayley:
> Should we remove the -from, and make the name shorter?

> I vote against

Same, as I was also going off the existing naming of other icons… See https://lucide.dev/icons/?search=from.

PR #1390, opened 2023-06-23
Add navigation icons

danielbayley:
`signpost` modelled on https://lucide.dev/icons/milestone. Maybe they are both slightly too small…?

`waypoints` moved from https://github.com/lucide-icons/lucide/pull/1319#issuecomment-1600786999.

> Can you describe the use case for `waypoints`?

@ericfennis As mentioned https://github.com/lucide-icons/lucide/pull/1319#discussion_r1220075343, it was meant to represent indirect/rerouting/proxy/VPN… and is loosely based on https://lucide.dev/icons?search=network /`share-2`/`workflow`… But is also versatile enough to include `maps`/`navigation`/orienteering/path etc… Especially if e.g there are multiple stops on the route, the icon would be perfect to represent, say a journey planner… It would go nicely with https://lucide.dev/icons/git-commit (and `git-commit-vertical`) from [#995](https://github.com/lucide-icons/lucide/pull/995#issuecomment-1509285280).


karsa-mistmere:
I have a very similar map pin icon in my drafts, I think a more pronounced perspective as in this one would maybe look better (i.e. more easily recognisable)?
map-pin

danielbayley:
> I have a very similar map pin icon in my drafts

@karsa-mistmere Nice, like it also! I had a couple of alts… So I guess it’s between these:

preview

jguddas (approved):

ericfennis (approved):

danielbayley:
@jguddas @ericfennis Do you guys have a preference between these https://github.com/lucide-icons/lucide/pull/1390#issuecomment-1609853830? options for `map-pinned`?

jguddas:
+1 for NR.2, it does feel a bit like an alien with shoulders, but I don't mind :+1:

danielbayley:
@ericfennis @karsa-mistmere @jguddas I changed the rounding to 1px as a nicer compromise…

PR #1394, opened 2023-06-24
Add wellbeing icons

danielbayley:
We probably don’t need them all, but the `candle*` variants are a nice alternative light/dark mode toggle.

`yin-yang` is an excellent way of representing contrast, or even dark/light mode toggle.

`triangle-stripes` to represent class/hierarchy.

`flower-lotus` alts…

Untitled
<br/>
<br/>

Related to #1398.


danielbayley:
I also had these in drafts… But that was before I knew about the

> Icons do not include religious or political imagery

…rule https://github.com/lucide-icons/lucide/pull/1224#issue-1698657210, so just leaving them here for reference:

preview

ericfennis:
Can we have some use cases for this one?
The lotus I get: "digital wellbeing", the others not.

For your other drafts you posted, We should not add religious or cultural symbols, this is out of scope of the Lucide library. Lucide should be neutral of religion and culture. This because lucide is world wide used by all kinds of people with their own background, if we considering adding these symbols we also should be open for all other symbols in religion and culture and that are a lot. And a lot of religion and culture symbols having conflicts of meaning with each other. I really want to avoid any discussions about religion or culture in this project.

ericfennis:
Use cases are not really solid.
I never saw a candle used for light/mode dark mode toggles. We shouldn't add that one.
`triangle-stripes` is a bit abstract to use for hierarchy, [network](https://lucide.dev/icons/network) is the one that fits better for that.

I think these are the ones we should add.
`yin-yang` is nice, and represents "balance".
`lotus` is nice, and could be used for "digital wellbeing".

ericfennis:

PR #1411, opened 2023-06-29
Optimise `send` icon/add variant

danielbayley:
Leaving alt here for reference…

preview


karsa-mistmere:
The central gap is very dense and the icon seems a bit too large in the cohesion preview.

danielbayley:
> The central gap is very dense and the icon seems a bit too large in the cohesion preview.

@karsa-mistmere Fixed now, matches the other `send` when rotated…

jguddas:
I'm not a fan of introducing variants, why add this?

danielbayley:
> I'm not a fan of introducing variants, why add this?

@jguddas We already have variants. People have preferences, and maybe their ideal set of common icons is _almost_ covered by Lucide, but just not _quite_… very frustrating. That said, we obviously shouldn’t be going crazy…

This one is commonly found in the wild, and specifically looks better on a button say, where it is quite likely to be used.

karsa-mistmere:
While I do see the rationale behind adding a horizontal alt, I think a horizonal send icon should look more like this, to keep the style consistent:
send-horizontal

danielbayley:
> I think `send-horizonal` (note name) should look more like this, to keep the style consistent:

@karsa-mistmere `push`ed…

karsa-mistmere (approved):

ericfennis (approved):

Issue #1367, opened 2023-06-16
lucide.eot file missing in lucide-static


schmidt-oliver:
## Prerequisites

* Version: V. 0.244.0
* Are you running from source/main: no
* Are you using a released build: lucide-static, import using parceljs
* Operating system: Unix
* Bits: 64 bits

## Step to reproduce

- `npm install lucide-static`
- Go to `node_modules/lucide-static/font/.`
- See that `lucide.eot` is missing.
- Try to `@import "~lucide-static/font/Lucide.css";`
- Getting error

### Actual behavior
- `npm install lucide-static`
- Go to `node_modules/lucide-static/font/.`
- Find the file `lucide.eot`
- Try to `@import "~lucide-static/font/Lucide.css";`
- File found, so no error occurs

## Any message or error

`@parcel/resolver-default: Cannot load file '../node_modules/lucide-static/font/lucide.eot' in './src'.`

## Resources

- No resources

##  Comment
Hello together, other packages like those from @fontsource only serve woff2 and woff files in their css, isn't that enough? Or is it possible to provide a css file with only the icon classes and their unicode content, so that you can add the rest yourself? Thank you.

ericfennis:
@schmidt-oliver Thanks for pointing this out!

schmidt-oliver:
I found out that you are using [fontcustom](https://github.com/FontCustom/fontcustom) based on ruby and python to create the font files. There is an interesting node package called [svgtofont](https://github.com/jaywcjlove/svgtofont). Eventually it is a great alternative. I will take a look and try it out.

ericfennis:
Thanks for pointing out. Currenty I'm nog really happy with fontcustom. It's also not very well maintained. We currently have some open issues with the font. We really need stable hex values for the icons. We currently have some open issues about it: #514, #926 and #538. So a custom icon-codepoints map for the font bundler is a must.
I was looking at [fantasticon](https://github.com/tancredi/fantasticon)  but maybe svgtofont is also a good one to look at.

schmidt-oliver:
Thank you and good to know. I will take this into account in my tests.

schmidt-oliver:
[...]

Issue #514, opened 2022-02-22
Stable Font Codepoints


lilithium-hydride:
## Prerequisites

* Version: Releases v0.17.0 - v1.17.1
* Running from: `lucide.ttf`
* Operating system: Artix Linux x86_64

## Step to reproduce

1. Use an older release's codepoints in something.
2. Make note of the icons that are rendered from those codepoints.
3. Update the font file to the latest release without adjusting the codepoints.
4. Observe that the rendered icons are different.

### Actual behavior

1. Use an older release's codepoints in something.
2. Make note of the icons that are rendered from those codepoints.
3. Update the font file to the latest release without adjusting the codepoints.
3. Observe that the rendered icons do not change between releases.

## Any message or error

N/A

## Resources

![Left: new, Right: old](https://user-images.githubusercontent.com/78992082/155236408-900e6bdb-7d01-49c0-82f5-1f26257995d7.png)

## Context

It's currently necessary to pay close attention to the font version when using it outside of a web environment. I understand that this would make it practically impossible to keep icon locations neatly organized, but this shouldn't hurt the web development side of things at all, and would make it far less painful for font users to keep up with releases.

ericfennis:
@lilithium-hydride Tnx for reporting not sure why this happens, we need to check the build.

ericfennis:
Is fixed in #1413 !

agorgl:
Commenting here to let the authors know that this issue is NOT fixed, and has occurred so far 2 times in a period of some months I've been using lucide font. More details on specific versions that this can be reproduced in issue https://github.com/lucide-icons/lucide/issues/3619

ericfennis:
Duplicate of #3619

PR #1413, opened 2023-06-29
Changed font creation process

schmidt-oliver:
Moved font creation to node library [svgtofont](https://github.com/jaywcjlove/svgtofont).

Fixes these Issues:
- #1367
- #514 

Because of a problem with [path directions](https://stackoverflow.com/questions/72379177/auto-colour-fill-when-generate-font-from-svg-figma-icon-using-icomoon/72451714#72451714) in [outline-stroke](https://github.com/elrumordelaluz/outline-stroke) (see [#5](https://github.com/elrumordelaluz/outline-stroke/issues/5)) I moved the creation of outline SVGs to [svg-fixer](https://github.com/oslllo/svg-fixer).

There are some options which could be changed in both build scripts. Please test yourself and change them if needed. Tried to adapt your code style. Updated the github workflow. I hope I didn't missed something.

New process:
- Build outline icons with [svg-fixer](https://github.com/oslllo/svg-fixer)
- Build font with [svgtofont](https://github.com/jaywcjlove/svgtofont)
- ~~Do not delete the created ./lucide-font folder. The info.json file that is created the first time contains the code points that will be used in the next execution. Existing code points will be reused. New Icons will be added after existing.~~

Tested the whole process myself. But you should test it yourself before bringing it to production.


ericfennis (commented):
Fantastic job, I'm also impressed with the amount of output and different files it has for using the font. And JSON file is really useful for the site.
I noticed that the CSS classes are now prefixed with `lucide-` instead of `icon-`, maybe we should keep the old class prefix for backward compatibility.

The lucide-font.yml workflow has currently `container` option declared for the ruby environment but if we switch to this flow we can remove that one.

schmidt-oliver:
Thank you for pointing this out. Fixed both.

- [x] I noticed that the CSS classes are now prefixed with `lucide-` instead of `icon-`, maybe we should keep the old class prefix for backward compatibility.
- [x] The lucide-font.yml workflow has currently `container` option declared for the ruby environment but if we switch to this flow we can remove that one.

ericfennis:
@schmidt-oliver Thanks for your changes, this looks really nice!
One thing I'm not sure about and that the `info.json` file. When releasing a new version the workflow need to have this file to know the old assigned codepoints.
So I thought maybe we can think of something to define those unicodes without the need of the info.json file. And the best way I think is counting up the unicode number from when the icon was added to the library.


So for the website we already fetching this data with `scripts/writeReleaseMetaData.mjs`. This outputs a file with these contents:
```json
{
  "accessibility": {
    "createdRelease": {
      "version": "0.33.0",
      "date": "2022-04-24T12:12:44+02:00"
    },
    "changedRelease": ...
  },
  "activity-square": {
    "createdRelease": {
      "version": "0.244.0",
      "date": "2023-06-12T21:36:07+02:00"
    },
    "changedRelease": ...
  },
...
```
We could use `createdRelease` to sort the whole list of icons and then use these order to define the unicodes:
So if we start counting from `&#57400;` we can count up all icons indexes up to this unicode number. This way we don't need to share this `info.json` between builds. And another thing is that we probably can use this really easy for the website as well. (code examples, copy character code, etc)

Let me know what you think of this idea

schmidt-oliver:
Interesting idea. You can fetch the json file, sort the list by `createdRelease.date` and add the index of the icon to the unicode start number. Sounds great, but the `writeReleaseMetaData.mjs` does not contain modules to import. Should we assume that the JSON file `releaseMetaData.json` already exists in `.vitepress/data` while creating the font? If so, I can adjust the process accordingly.

ericfennis:
@schmidt-oliver Yeah, the easiest way is simply to run the `scripts/writeReleaseMetaData.mjs` before we run the build font script. So this way you can import the JSON file and do the mapping in your script.

schmidt-oliver:
Well, I will probably update the build script within the next week.

ericfennis:
@schmidt-oliver yeah, take your time. Thanks for the work you already did!

ericfennis:
Yeah you can ignore these warnings. These icons are created before the fork (forking from feather). They should have the date of the fork.
I think if multiple icons have the same date, then they should be sorted by name.

schmidt-oliver:
Hey @ericfennis, found two icons where `icon.releaseDate` is missing: `ban` and `panel-left`. Can you please check why and fix that issue?

ericfennis:
@schmidt-oliver I Found the issue, fixed it on the main branch, and rebased this PR.
Can you update the `pnpm-lock.yaml`?

schmidt-oliver:
@ericfennis I have changed the process from using the `info.json` file to using the version information. I set `startUnicode = 57400;` as you suggested. Since you fixed the missing information of the two icons, everything works fine. I've checked the sort order twice to make sure it's correct. The code is linted.

I have a suggestion regarding the `releaseMetaData.json` file. My function `convertReleaseMetaData` converts the content of this file to the following notation (array, name as attribute) and adds the `index`  according to the defined sort order and the `unicode` according to the `startUnicode` and `index`.
```js
[
  {
    createdRelease: { version: '0.0.0', date: '2020-06-08T16:39:52+0100' },
    changedRelease: { version: '0.244.0', date: '2023-06-12T21:36:07+02:00' },
    name: 'activity',
    index: 0,
    unicode: 57400
  },
  {
    createdRelease: { version: '0.0.0', date: '2020-06-08T16:39:52+0100' },
    changedRelease: { version: '0.18.0', date: '2022-04-06T11:13:36+02:00' },
    name: 'airplay',
    index: 1,
    unicode: 57401
  },
  {
    createdRelease: { version: '0.1.0-alpha.0', date: '2020-06-08T16:39:52+0100' },
    changedRelease: { version: '0.152.0', date: '2023-04-16T14:17:17+02:00' },
    name: 'alarm-clock',
    index: 2,
    unicode: 57402
  },
  {
    createdRelease: { version: '0.11.0', date: '2020-06-08T16:39:52+0100' },
    changedRelease: { version: '0.133.0', date: '2023-04-09T11:46:30+02:00' },
    name: 'album',
    index: 3,
    unicode: 57403
  },
  // ...
]
```
As we want to display the Unicode numbers on the website, it would be smart to modify the data directly where it is created. After that we could use the prepared data for the font creation. What do you think about that?

tools/build-font/main.mjs
```diff
@@ -5,45 +5,58 @@ import path from 'path';
...
+  return comparison;
+}
+
+function sortByCreatedReleaseDate(a, b) {
+  const dates = [a, b].map((value) => new Date(value.createdRelease.date).valueOf());
+  return (dates[0] > dates[1]) - (dates[0] < dates[1]);
+}
+
+function sortByName(a, b) {
+  return new Intl.Collator().compare(a.name, b.name);
```
ericfennis, line 55: I think It would be best to pass a default locale here. So we are always sure it will have the same output in all environments.

ericfennis (commented):
@schmidt-oliver Fantastic work. Looks really good.
I don't fully understand the last part of your last comment, do you mean to run the website builds first and then use the created "data" for the font builds?

tools/build-font/main.mjs
schmidt-oliver: Ahh, sure. Which one do you prefer?

ericfennis: I assume it doesn't really matter that much for English, so go for `en-US`?

ericfennis:
@schmidt-oliver Yeah that would be hard to accomplish because website is built on an other cloud environment then Github actions.

schmidt-oliver:
@ericfennis Ahh, I see. I added the default locale. But I have some conflicts with the `pnpm-lock.yaml` file. Is it possible for you to merge my changes locally and recreate `pnpm-lock.yaml' yourself?

ericfennis:
I think best way is to accept all incoming changes and run `pnpm install` again. That will do the trick.

schmidt-oliver:
Checked files, they look fine now, but github checks fail.

schmidt-oliver:
@victorbnl I only had to fix the `pnpm-lock.yaml` conflict again.
@ericfennis Project is ready to use, only the merge is missing. Hope this time it will be merged before any other conflicts occur ^^. Would be nice if it gets merged soon. I also want to use it in production without my own fork.

Have a nice day =)

ericfennis:
@schmidt-oliver Awesome! I will review and test this.
But let's get this one merged this week!

schmidt-oliver:
@ericfennis Awesome, I re-checked it today and it's working well on my local machine. If any issues arise in the future, I will try to resolve them.

@victorbnl, you are welcome to review it as well. The project installation is based on `pnpm`. After package installation, two scripts are required to generate the font: `build:outline-icons` and `build:font`.

- The `build:outline-icons` script uses `oslllo-svg-fixer` to convert the icons to their outline version. This is necessary for TTF conversion. The previously used library, `outline-stroke`, did not [work as intended](https://github.com/lucide-icons/lucide/pull/1413#issue-1780771528) and had to be replaced.

- The `build:font` script uses `svgtofont` to transform the outlined version of the icons into a font. It takes the release metadata of each icon and its name to order them. Based on that sort order the unicode code points are created. If `svgtofont` becomes unsupported in the future, `fantasticon` could serve as an alternate solution, since both tools rely on the [same dependencies](https://github.com/lucide-icons/lucide/issues/1367#issuecomment-1609413219).

I thought a little summary would be a good idea.

victorbnl:
Nice, thank you for the summary! Looks quite good to me, but two things:

- Why use two different scripts for `build:outline-icons` and `build:font`? As far as I know `build:outline-icons` is only used in order to use `build:font` after, and `build:font` requires `build:outline-icons`. Thus separating them just makes an additional unnecessary script one has to run for no reason.

- Are you planning to automate the font build and publish it somewhere, say for each version as a GitHub release? That would be very nice as well.

schmidt-oliver:
Hi @victorbnl,

> * Are you planning to automate the font build and publish it somewhere, say for each version as a GitHub release? That would be very nice as well.

https://github.com/lucide-icons/lucide/blob/12d612d7b298581cab0f7c98c52a9fbcdcffb243/.github/workflows/lucide-font.yml#L39-L49

Both steps are part of the GitHub workflow and the final font is released in each package. They are not intended to be run by an end user.

> * Why use two different scripts for `build:outline-icons` and `build:font`? As far as I know `build:outline-icons` is only used in order to use `build:font` after, and `build:font` requires `build:outline-icons`. Thus separating them just makes an additional unnecessary script one has to run for no reason.

https://github.com/lucide-icons/lucide/blob/12d612d7b298581cab0f7c98c52a9fbcdcffb243/package.json#L18-L19

Each script does it's own job and as long as they aren't intended to be run by an end user it's totally fine to separate them in my opinion. This way it is a bit clearer and easier to identify problems in the code and to write tests. But sure, it's also possible to combine all the steps in a single task:

```json
 "build:font": "pnpm --filter outline-svg start && pnpm --filter docs prebuild:releaseJson && pnpm --filter build-font start", 
```

ericfennis:
@schmidt-oliver Awesome work. I tested this with an old HTML file from the old font and it all looks great.

1 change I'm not totally convinced yet is chang in the `outline-icons` step. Previously Github Actions took ~1m 45s to complete, with the changes in this PR the build time will increase to +6min (according to the Lucide font checks workflow). To me this is a bit too long, Can you explain why we should move forward with the new workflow?

schmidt-oliver:
@ericfennis Thank you.

> 1 change I'm not totally convinced yet is chang in the `outline-icons` step. Previously Github Actions took ~1m 45s to complete, with the changes in this PR the build time will increase to +6min (according to the Lucide font checks workflow). To me this is a bit too long, Can you explain why we should move forward with the new workflow?

> * The `build:outline-icons` script uses `oslllo-svg-fixer` to convert the icons to their outline version. This is necessary for TTF conversion. The previously used library, `outline-stroke`, did not [work as intended](https://github.com/lucide-icons/lucide/pull/1413#issue-1780771528) and had to be replaced.

No problem, the reason is described [in this issue](https://github.com/elrumordelaluz/outline-stroke/issues/5). But I will try to break it down. The problem is based on the way how the outlines are created. The outline paths can be clockwise or counter-clockwise. And there are rules to determine whether the paths are filled or cut out. The [even-odd rule](https://en.wikipedia.org/wiki/Even%E2%80%93odd_rule) is used in the `outline-stroke` library and is not compatible with the `svgtofont` library and its dependencies `svgicons2svgfont`, `svg2ttf`. To create the font with these libraries, which are the only JS-based libraries I've found, you need to have outlined icons based on the older [non-zero rule](https://en.wikipedia.org/wiki/Nonzero-rule). The `oslllo-svg-fixer` takes this rule into account and respects the path directions. It seems that outlining with this rule takes more time. It's possible to reduce the time by reducing the precision of the outlining process. But I don't recommend it.

More information about the the two rules:
- https://www.sitepoint.com/understanding-svg-fill-rule-property/
- https://oreillymedia.github.io/Using_SVG/extras/ch06-fill-rule.html

victorbnl:
> Both steps are part of the GitHub workflow and the final font is released in each package. They are not intended to be run by an end user.

My bad, as I saw there was no action running on your fork I thought you hadn’t made one. Apparently you just didn’t enable them. That’s nice!

> Each script does it's own job and as long as they aren't intended to be run by an end user it's totally fine to separate them in my opinion. This way it is a bit clearer and easier to identify problems in the code and to write tests. But sure, it's also possible to combine all the steps in a single task

Given it’s ran in an action, it doesn’t matter much indeed.

victorbnl:
Why run the workflow on pull requests though? The font should be built every time Lucide icons change, right?

ericfennis:
@schmidt-oliver Thanks makes sense! The process probably can speed up if the processes are done in worker pools. Looks like SVGFixer doesn't do that yet. Could be a nice improvement.

But that's aside I think it's ready to go then!

@victorbnl As a maintainer I like to have workflows run as fast as possible. During a release, the font workflow is dependent on two other workflows that build packages. So that's why I'm not happy with this increase in time. But I will see if we can speed things up later.

Also, the Font workflow is not running on all PRs, only if there are changes in: icons, font build scripts, and the package-lock file. Why? Because then we can know if the font process is still working correctly or if something has changed.

ericfennis (approved):

victorbnl:
> @victorbnl As a maintainer I like to have workflows run as fast as possible. During a release, the font workflow is dependent on two other workflows that build packages. So that's why I'm not happy with this increase in time. But I will see if we can speed things up later.

I didn’t discuss the process speed, was it me you wanted to mention?

> Also, the Font workflow is not running on all PRs, only if there are changes in: icons, font build scripts, and the package-lock file. Why? Because then we can know if the font process is still working correctly or if something has changed.

Well from what I understand the font created in this action is not directed towards end users but Lucide maintainers for testing then. In that case, where is the user who wants to use the font supposed to get it from?

schmidt-oliver:
> @schmidt-oliver Thanks makes sense! The process probably can speed up if the processes are done in worker pools. Looks like SVGFixer doesn't do that yet. Could be a nice improvement.

Interesting idea. I have never worked with worker pools before. The `oslllo-svg-fixer` allows a single SVG as parameter instead of a folder path. So it's possible to handle the worker pools without touching the library. Let me see if I can do something about that.

ericfennis:
@schmidt-oliver I'm already working on it.
The first tests are really promising.

See my PR: https://github.com/oslllo/svg-fixer/pull/89

danielbayley:
@ericfennis @schmidt-oliver @victorbnl Do we need to run the entire set of SVGs through the svg-fixer outline step on every build? I wonder if we could [cache](https://github.com/actions/cache) the intermediate outlined SVGs, and only rerun the outline step over those that have changed since the last build…?

schmidt-oliver:
@danielbayley Sure, we could reuse them. Then we need to store the date or version number of each outlined icon during the outline step and compare it to its `changedRelease.date` or `changedRelease.version.` If the release is newer, we need to rebuild it.

Finii:
@ericfennis wrote here https://github.com/lucide-icons/lucide/pull/1413#issuecomment-1625291792:

> One thing I'm not sure about and that the `info.json` file. When releasing a new version the workflow need to have this file to know the old assigned codepoints.

I just wanted to note that the `info.json` approach was for superiour and did in fact solve #514. The new approach does shift all codepoints if an old icon is dropped as the vacant slot can not be left empty and/or be filled with a new glyph. If a icon is renamed the confusion will be more local possibly shifting only some icons, but still.

Whatever the problem is with also fetching the `info.json` file and commiting it back on icon additions, I seem to not get it. Maybe it would be worthwhile to solve the problems that Eric saw with that file, to really solve 514. Other icons collections do it that way, its not that it would be impossible? The now deserted but in former times often used svg-to-font generator [FontCustom](https://github.com/FontCustom/fontcustom/) automagically did it...

ericfennis:
@Finii So you're telling us that #514. Is not fixed? That's not what we wanted.
Unfortunately `info.json` approach is not feasible. That would mean we need to keep that file constantly updated when new icons are added, and that is just too much work. And automatically pushing it to the main is not possible, due to security reasons (main is a protected branch). 

But It is a bummer to hear that this is still not working correctly as we expected.
Can you create an issue for this?

ericfennis:
@danielbayley Yeah I thought of this idea as well. This great solution for icons that will be renamed. But the problem is what to do with new icon PRs. If we have multiple PRs open which icon will get which `unicode`. If some other PR is merged the other one need to update it. So that's not very practical. And you can't simply just fill in something. Unicode positions are limited. 😞

danielbayley:
> the problem is what to do with new icon PRs. If we have multiple PRs open which icon will get which `unicode`. If some other PR is merged the other one need to update it.

@ericfennis Would this not cease to be a problem if we could make sure every PR was first `rebase`d onto the last `merge`d? New icons would simply take up next available slots, no?

> Unicode positions are limited.

Isn’t the limit something like ~150,000 though? Not even I am going to churn out _that_ many icons… 😅

Finii:
Just as idea ...

> That would mean we need to keep that file constantly updated when new icons are added, and that is just too much work. And automatically pushing it to the main is not possible, due to security reasons (main is a protected branch).

* A new icon does not get any codepoint associated in the PR (as it is impossible to say which  it would get, as you pointed out)
* A workflow is triggered on changes in the svg directory just on pulls to master/main. That workflow checks all svgs that are existing in that workflow and adds new icons to the `icons.json` at the bottom (i.e. next free codepoint).
* If the `icons.json` is changed by the workflow the workflow itself commits that changes to master/main (in fact you can always commit it back, if there is no change nothing will happen anyhow)
* If an existing icons is removed or renamed the PR for that needs to change the `icons.json` accordingly

That means almost no manual attention is needed.

And I can not see the security reasons you mention. The workflow that is triggered on pulls to master can only be triggered by persons/processes that are allowed to pull to master. The workflow inherits that right, and can add one more commit on the fly.

At Nerd Fonts for example we
* on pushes to master that change any relevant files
* we create a zip file from all files that are needed for custom patching
* commit the new version of the zip back automatically

ericfennis:
@Finii We had something similar before, it was a workflow that updated our packages.json files after our releases in our GitHub repo. 
That used the GitHub action token to push commits to our main branch. That worked fine for a long time. But we hadn't turned on branch protection rules on our `main` branch. We realized this was needed to secure our project, but that made it impossible to push back from GitHub actions because Github Actions is not a user you can give access to your branch. See https://github.com/orgs/community/discussions/13836
And I'm not in favor of using PAT tokens, because that is a big possible security risk.

If you are doing it with the use of branch protection rules and without PATs, I really want to know how. I've searched a lot Internet for a solution but never found a solution.

schmidt-oliver:
@Finii @ericfennis @danielbayley: We are creating a releaseJson file that contains all the information about icon releases and changes, including their timestamps. If it is possible to find out when an icon was deleted or renamed (a renamed icon could be handled like a deleted and a new one), it should be possible to take this into account when creating codepoints. But it's much more complex than the current algorithm. That approach would not need to store the codepoints. But every deletion have to be taken into account. Is it possible to find out when an icon was deleted?

PR #1430, opened 2023-07-14
Add `shopping-basket` icon

danielbayley:


jguddas:
The `shoping-bag` looks cute and a bit big?

danielbayley:
> The `shoping-bag` looks cute and a bit big?

@jguddas It [already exists](https://lucide.dev/icons/shopping-bag), I just optimised it… Do we think it should be slightly smaller? @karsa-mistmere @ericfennis

karsa-mistmere:
While it is in fact a bit too large, I don't think making it more square-ish is the way to go, since then it looks less and less like a shopping bag. I tried to make it narrower (the same dimensions as `tablet`), but that doesn't feel great either:
image

I think we should maybe just keep it as it is.

karsa-mistmere (approved):

ericfennis (approved):
@danielbayley Nice work!

PR #1436, opened 2023-07-17
Fix fills in smaller dots

karsa-mistmere:
The current `palette`1 icon renders incorrectly in various software including Figma:
image

This PR fixes the issue and improves the pixel perfection of the outline shape.


ericfennis:
hmm

danielbayley:
Outline shape looks better 👍🏼 But I prefer the larger globs of colour.

karsa-mistmere:
> Outline shape looks better 👍🏼 But I prefer the larger globs of colour.

Those are quite buggy though. 😅

danielbayley:
> Those are quite buggy though. 😅

Strange, is that the case with all `circle`s of that size? @karsa-mistmere

karsa-mistmere:
> Strange, is that the case with all `circle`s of that size? @karsa-mistmere

Yeah, I'm afraid it is.

karsa-mistmere:
> They work in Figma if we provide the `fill` attribute.

Something that we do not and cannot actually support out of the box (not yet anyway). 😅

And it's not just Figma that's affected either, Chrome can also render these circles funkily:
image

---

Alternatively we could do a rather ugly hack like this:
palette-uglyhack

```svg
  <circle cx="13.5" cy="6.5" r=".5" />
  <circle cx="17.5" cy="10.5" r=".5" />
  <circle cx="8.5" cy="7.5" r=".5" />
  <circle cx="6.5" cy="12.5" r=".5" />
```
=>

```svg
  <path d="M13.5 7a.5.5 0 0 1 0-1 .5.5 0 0 1 0 1.01" />
  <path d="M17.5 11a.5.5 0 1 0 0-1 .5.5 0 0 0 0 1.01" />
  <path d="M8.5 8a.5.5 0 1 0 0-1 .5.5 0 0 0 0 1.01" />
  <path d="M6.5 13a.5.5 0 1 0 0-1 .5.5 0 0 0 0 1.01" />
```

ericfennis:
I normally am a bit against hacky things.
But still not sure about the design of this icon with smaller circles. 
🤔

ericfennis:
@karsa-mistmere I've added the `fill="currentColor"` to the remaining icons that use these circles as we discussed in Discord.

PR #1439, opened 2023-07-17
Update createLucideIcon.ts

D3OXY:
Added Missing ref prop

### DIFF
--- a/packages/lucide-react/src/createLucideIcon.ts
+++ b/packages/lucide-react/src/createLucideIcon.ts
@@ -1,4 +1,4 @@
-import { forwardRef, createElement, ReactSVG, SVGProps, ForwardRefExoticComponent, RefAttributes, RefObject } from 'react';
+import { forwardRef, createElement, ReactSVG, SVGProps, ForwardRefExoticComponent, RefAttributes } from 'react';
 import defaultAttributes from './defaultAttributes';
 
 export type IconNode = [elementName: keyof ReactSVG, attrs: Record<string, string>][]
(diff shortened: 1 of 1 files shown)


packages/lucide-react/src/createLucideIcon.ts
```diff
@@ -8,9 +8,10 @@ export type SVGAttributes = Partial<SVGProps<SVGSVGElement>>
 export interface LucideProps extends SVGAttributes {
   size?: string | number
   absoluteStrokeWidth?: boolean
+  ref?: RefObject<SVGSVGElement>;
 }
 
-export type LucideIcon = ForwardRefExoticComponent<LucideProps & RefAttributes<SVGSVGElement>>
+type LucideIcon = ForwardRefExoticComponent<LucideProps & RefAttributes<SVGSVGElement>>;
```
ericfennis, line 14: Sorry maybe I was not clear, I mean like this. We should keep the export.
```suggestion
export type LucideIcon = ForwardRefExoticComponent<LucideProps>;
```

ericfennis: Or is this not working?

ericfennis (dismissed):
awesome work!

ericfennis (changes requested):
Suggestion mode was not working, but I think this is the way:

`RefAttributes` includes the ref property.
```ts
export type SVGAttributes = Partial<SVGProps<SVGSVGElement>>

type ComponentAttributes = RefAttributes<SVGSVGElement> & SVGAttributes

export interface LucideProps extends ComponentAttributes {
  size?: string | number
  absoluteStrokeWidth?: boolean
}

export type LucideIcon = ForwardRefExoticComponent<LucideProps>;
```

ericfennis (approved):

PR #1440, opened 2023-07-17
Move dynamicImportFile to dedicated file

### LINKED ISSUE #1437: All icons are loaded since 0.260.0
## Prerequisites

* Version: 0.260.0
* Are you running from source/main: no
* Are you using a released build: from npm 0.260.0
* Operating system: Windows, Mac OSX, Ubuntu 22.04 LTS (Jammy Jellyfish)
* Bits: 64

## Step to reproduce

Using Remix 1.18.1 server-side-rendering, only icon-specific imports are used, no general `icons` import, also no `lazy` or `dynamicIconImports` commands used:

With version **0.259.0** I see a page loading 40 (chunks) in network panel.

After updating to version **0.260.0** (and changing nothing else) more than 1200 network requests cause a substantial increase in page loading time (~30seconds).

ericfennis:
Fixes slow compilation on VIte and Remix with the new `dynamicIconImports` file. Moded it to a separate import.
Closes #1437


papsko:
Looks like moving `dynamicIconImports` to separate file resolved #1437 for regular use case (staticaly importing specific icons), however the problem still exists while trying to use these dynamic imports, right?

I'm using Vite. I created dynamic Icon component as proposed in official docs (v. 0.447.0) and when Vite detects `import { dynamicIconImports } ...` it preloads all ~1500 icons at the beginnig and then lazy imports separate chunks when actual icons are needed.

image
image

This happens only in dev mode, but still it slows down dev server a lot.
Do you have any idea how to solve this issue?

PR #1425, opened 2023-07-10
Add forward and backward overlap icons

james-yeoman:
I'm not an artist in the slightest, but by using `copy` as a base, I was able to get these icons looking serviceable.

These icons may need to be measured in something more capable, since I used [svgviewer](https://www.svgviewer.dev), which doesn't have pixel measurement capabilities. It's also not the best for dimensional accuracy, as you can see from this screenshot.

<details>
<summary>Collapsible screenshot and svg code from svgviewer</summary>

image

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-bring-to-front">
  <!--  Debug Border, remove before SVGOMG  -->
  <rect width="24" height="24" x="0" y="0" stroke-width="1" stroke="red" />
  
  <rect width="8" height="8" x="8" y="8" rx="2" ry="2"/>
  <path d="M 4 10 c -1 0 -2 -1 -2 -2 v -4 c 0 -1 1 -2 2 -2 h 4 c 1 0 2 1 2 2"/>
  <path d="M 14 20 c 0 1 1 2 2 2 h 4 c 1 0 2 -1 2 -2 v -4 c 0 -1 -1 -2 -2 -2"/>
</svg>

<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-send-to-back">
  <!--  Debug Border, remove before SVGOMG  -->
  <rect width="24" height="24" x="0" y="0" stroke-width="1" stroke="red" />
  
  <rect width="8" height="8" x="2" y="2" rx="2" ry="2" />
  <rect width="8" height="8" x="14" y="14" rx="2" ry="2" />
  <path d="M 7 14 v 1 c 0 1 1 2 2 2 h 1" />
  <path d="M 14 7 h 1 c 1 0 2 1 2 2 v 1" />
</svg>


<!-- Lucide copy icon for reference -->
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-copy">
  <!--  Debug Border  -->
[...]

PR #1444, opened 2023-07-20
Improve [`send`/`bring`]`-to-`[`front`/`back`] metadata

danielbayley:
Following #1425…

### DIFF
--- a/icons/send-to-back.svg
+++ b/icons/send-to-back.svg
@@ -9,8 +9,8 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect width="8" height="8" x="2" y="2" rx="2" />
-  <path d="M14 8a2 2 0 0 1 2 2" />
-  <path d="M10 16a2 2 0 0 1-2-2" />
-  <rect width="8" height="8" x="14" y="14" rx="2" />
+  <rect x="14" y="14" width="8" height="8" rx="2" />
+  <rect x="2" y="2" width="8" height="8" rx="2" />
+  <path d="M7 14v1a2 2 0 0 0 2 2h1" />
+  <path d="M14 7h1a2 2 0 0 1 2 2v1" />
 </svg>
--- a/icons/send-to-back.svg
+++ b/icons/send-to-back.svg
@@ -9,8 +9,8 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect x="14" y="14" width="8" height="8" rx="2" />
-  <rect x="2" y="2" width="8" height="8" rx="2" />
-  <path d="M7 14v1a2 2 0 0 0 2 2h1" />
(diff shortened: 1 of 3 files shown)


icons/send-to-back.svg
```diff
@@ -9,8 +9,8 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect x="14" y="14" width="8" height="8" rx="2" />
```
jguddas, line 12: I liked it better before.

danielbayley: > I liked it better before.

@jguddas 😅 Valid, but it isn’t consistent, or accurate… Maybe we ultimately decide that doesn’t matter? cc: @karsa-mistmere @ericfennis

ericfennis: I'm also with @jguddas other looked better. To me 100% it looks like two squares with connecting lines. The corners are tough to see in this adjustment.

karsa-mistmere: I also agree, my train of thought was the same as that of @ericfennis, while it's less consistent with its pair, the original design is a bit more legible.

danielbayley: > The corners are tough to see in this adjustment.

Another option would be to make them slightly sharper to compensate (as in third below)… Here are all the possible options for comparison:

preview

@jguddas @ericfennis @karsa-mistmere Thoughts?

jguddas: My vote is still for the original.

ericfennis: I vote for 1 and 3.

danielbayley: I think 3. So looks like @karsa-mistmere has the deciding vote…

karsa-mistmere: I'm also voting for the original — back when that icon was submitted I'd ruminated over posting a suggestion for no. 3 myself, so I definitely see where you're coming from, but I think all things considered, no. 1 simply has much better legibility.

We could check if alternatively, making `bring-to-front` more consistent with this one would work better (i.e. increasing the central square in that one).

karsa-mistmere: Column 1: current icons
Column 2: `bring-to-front` w/ larger rectangle to match `send-to-back`

image
image

Personally, I'm fine with either.

jguddas: I'm fine with making `bring-to-front` bigger.

danielbayley: > We could check if alternatively, making `bring-to-front` more consistent with this one (i.e. increasing the central square in that one) would work better.

I think that makes that one look much worse 😅 `reverted`, so just the metadata improvements then…

jguddas (approved):

ericfennis (approved):

PR #1459, opened 2023-07-27
Sort out `more` (ellipsis…) icons

danielbayley:


karsa-mistmere:
`more-horizontal-rectangle` is identical to https://lucide.dev/icons/form-input

Circles with a .5 must be avoided, see the issues w/ palette in #1436

danielbayley:
> `more-horizontal-rectangle` is identical to https://lucide.dev/icons/form-input

@karsa-mistmere Oh, it didn’t come up when searching for `more` etc! Missing metadata… Would that not be more appropriately renamed to `more-rectangle` then, given the others?

> Circles with a .5 radius must be avoided, see the issues w/ palette in #1436

Ah, frustrating!

jguddas:
`more` should be `ellipsis` IMO

danielbayley:
> `more` should be `ellipsis` IMO

@jguddas Ideally, but I was going off https://lucide.dev/icons/more-horizontal and https://lucide.dev/icons/more-vertical … Should they not also be renamed to `ellipsis-horizontal` and `ellipsis-vertical`? cc: @karsa-mistmere @ericfennis

ericfennis (approved):

jguddas (changes requested):
circle- and square- should be the group IMO

ericfennis:
@jguddas I don't exactly what you mean. Which icons should be in the group?

jguddas:
> @jguddas I don't exactly what you mean. Which icons should be in the group?

 The name should be flipped, circle should be first IMO.

ericfennis:
@jguddas Hmm not sure about this way of naming. Shouldn't we name it "context" first?

PR #1412, opened 2023-06-29
Add transport icons

danielbayley:
`car-*` alts:

preview

See also #1460.

PR #1460, opened 2023-07-27
Add additional transport icons

danielbayley:
Following on from https://github.com/lucide-icons/lucide/pull/1412#pullrequestreview-1550057373

* `car-sports-front` to represent racing/motorsport, or even a luxury car dealership on a map… Related to #1304.
* `coach-front` to distinguish between coach and existing bus (the biggest difference being short vs long journey times on say a travel booking site…)
* `car-electric-front` alternatives:

preview

Also leaving `bus-school-front` here for reference…

preview

`tram`/`-front` alts:

preview

`station-charge-electric` alternative:

preview


ericfennis:
The zap is too small in these icons. On 100% scale you can barely see them.

Maybe an idea for the electric charging station is too use a plug instead on the end of the hose?
For example the plug from the `cable` icon?

Like this one:
image

ericfennis:
@danielbayley What do you think of this charge station?
image

danielbayley:
> @danielbayley What do you think of this charge station?

@ericfennis Did you miss the draft at the bottom of this https://github.com/lucide-icons/lucide/pull/1460#issue-1824677065? 😅

ericfennis:
@danielbayley I saw it. But it would be nice if we use the same plug icon as the `cable` icon. And the same pump case as `fuel`.

danielbayley:
> The zap is too small in these icons.

@ericfennis It’s no smaller than—and in fact uses exactly the same zap as—the existing [`plug-zap-2`](https://lucide.dev/icons/plug-zap-2) icon. Again going for consistency… But one possibility would be to make [`car`](https://lucide.dev/icons/car-front)/[`taxi`](https://lucide.dev/icons/car-taxi-front)`-front` bigger to free up a bit more space…?

> it would be nice if we use the same plug icon as the `cable` icon. And the same pump case as `fuel`.

Agreed. Looks like there is some wiggle room to the left of [`fuel`](https://lucide.dev/icons/fuel) to play with…

jguddas:
The z is too small IMO.

PR #1453, opened 2023-07-22
Add grooming/bathroom/appliances/amenities icons

danielbayley:
Related to #1451, #1454 and #1461.

`shave-face` (shavers only socket) related to #1250.

PR #1451, opened 2023-07-22
Add furniture/amenities icons

danielbayley:
Related to #1453, #1454 and #1461.

`wardrobe`/`cabinet-filing` alts:

preview

PR #1461, opened 2023-07-27
Add amenities icons

danielbayley:
Related to #1451, #1453 and #1454. `washing-machine` drum matches `yin-yang` symbol in #1394.

`safe`/`vault` alts:

preview


ericfennis (approved):

icons/vault.svg
```diff
@@ -0,0 +1,22 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <rect width="18" height="18" x="3" y="3" rx="2" />
+  <circle cx="7.5" cy="7.5" r=".5" />
```
karsa-mistmere, line 13: Still not a fan of the .5 radius circles, in Figma they look awful:
image

Maybe, like `safe`, this could also do with some legs to make it more legible?

icons/washing-machine.svg
karsa-mistmere, line 1: Super cute!

karsa-mistmere (changes requested):

jguddas:
I would drop safe, I couldn't for the life of me figure it out before looking at the name.

Also use case wise, vault and safe are pretty close.

PR #1462, opened 2023-07-28
Add/reneme emoji faces

danielbayley:
These were lacking personality…

Using info from https://emojipedia.org/smileys and https://emojibase.dev/shortcodes.

Related to #1465.


karsa-mistmere:
I feel like we should probably use one of the standard [shortcodes](https://emojibase.dev/shortcodes/?) as names, not make up new ones, the GitHub shortcode system seems like an obvious candidate.

danielbayley:
> I feel like we should probably use one of the standard [shortcodes](https://emojibase.dev/shortcodes/?) as names, not make up new ones, the GitHub shortcode system seems like an obvious candidate.

@karsa-mistmere Nice idea, so should we use that, but swapping them around so that `face-` is the prefix?

karsa-mistmere:
> > I feel like we should probably use one of the standard [shortcodes](https://emojibase.dev/shortcodes/?) as names, not make up new ones, the GitHub shortcode system seems like an obvious candidate.
> 
> @karsa-mistmere Nice idea, so should we use that, but swapping them around so that `face-` is the prefix?

Most default GitHub shortcodes do not actually have "face" in them. :)

`:angry:` => 😠
`:frowning:` => 😦
`:sligthly_smiling_face:` => 🙂
`:smile:` => 😄 
`:laugh:` => 😆 
`:smiley:` => 😃 
`:neutral_face:` => 😐

danielbayley:
> Most default GitHub shortcodes do not actually have "face" in them. :)

@karsa-mistmere Right but we ideally want them grouped together in the file tree…

> the shortcodes as metadata.

Without underscores?

karsa-mistmere:
> > Most default GitHub shortcodes do not actually have "face" in them. :)
> 
> @karsa-mistmere Right but we ideally want them grouped together in the file tree…

These are just a handful of the icons we have that are also Unicode emojis, so I think not necessarily.

> > the shortcodes as metadata.
> 
> Without underscores?

No, I meant as separate metadata items, other than tags.

danielbayley:
> These are just a handful of the icons we have that are also Unicode emojis, so I think not necessarily.

I have more… 😅

> separate metadata items, other than tags.

@karsa-mistmere What do you mean exactly? A separate array of properties, like `categories`?

karsa-mistmere:
> > These are just a handful of the icons we have that are also Unicode emojis, so I think not necessarily.
> 
> I have more… 😅

Surely, but I meant many of our existing icons are also valid emoticons: heart, broken heart, cat, dog, tree, gem etc.

> > separate metadata items, other than tags.
> 
> @karsa-mistmere What do you mean exactly? A separate array of properties, like `categories`?

Just a single property with a scalar unicode code value.

danielbayley:
> Just a single property with a scalar unicode code value.

This is just an idea at this stage though?

@karsa-mistmere Something like this?
~~~ json
"unicode": "1F602",
~~~

jguddas:
I don't like any of our smilies, so :-1: for adding more like them.

:+1: for the 💩

danielbayley:
👀 and 💩 moved to #1465.

danielbayley:
> These were lacking personality…

> I don't like any of our smilies

@jguddas Me neither, particularly 😅 Hence trying to improve them…

ericfennis:
Sorry I really love to see the effort you put into making all these emojis, but I don't think we should add emojis. Simply because emojis are not icons. Yes we have some basic emojis but they have some good use cases. For example a "Customer satisfaction" form. or a "Like" or reaction button (like GitHub has). But I don't think we will need all these. If you for example want to make a chat application I think it's more obvious to use the standard unicode emojis instead of lucide icons.
I feel a bit ashamed to say I need to close this. Please next time discuss it first in an issue before your put in all the effort and make a PR. I really appreciate your effort and I hope you understand my decision.

danielbayley:
@ericfennis Bit rash… What about the improvements to the existing ones? ([`smile`](https://lucide.dev/icons/smile), [`frown`](https://lucide.dev/icons/frown) etc…)

> These were lacking personality

I understand not adding more emoji for emoji sake, but what about adding emoji as metadata to their equivalent, already existing icons? Or @karsa-mistmere suggestion of adding `unicode`s to those? Some of that work is already done here…

> discuss it first in an issue

I don’t see any advantage of moving that stuff to issues instead… The main reason for preferring to review stuff in PRs is because we have @jguddas’ handy preview bot, which we don’t have in issues, as far as i’m aware? Also better to look at something more concrete, with the aid of file review features…

> before your put in all the effort

Most of them already exist in a massive _Illustrator_ file at this point…

See also my https://github.com/lucide-icons/lucide/pull/1465#issuecomment-1659050476

Issue #1446, opened 2023-07-20
Vercel/Next.js/React/Node 'too many open files' error


leeuwd:
## Prerequisites

* Version: `lucide-react@0.262.0`, `next@13.4.10`
* Are you running from source/main:  Next.js's Webpack
* Are you using a released build: Yes, https://github.com/lucide-icons/lucide/tree/v0.262.0
* Operating system: Node.js 18.x
* Bits:

## Step to reproduce

Installed using `yarn`, following https://github.com/lucide-icons/lucide/tree/main/packages/lucide-react#lucide-react. 

Error happens when all icons are imported as React Component individually, and also when all icons are imported using Next.js [dynamic import](https://github.com/lucide-icons/lucide/tree/main/packages/lucide-react#lucide-react).

## Any message or error

Error message via Vercel logs:

```
[Error: EMFILE: too many open files, open '/var/task/node_modules/lucide-react/dist/esm/icons/share.mjs'] {
  errno: -24,
  code: 'EMFILE',
  syscall: 'open',
  path: '/var/task/node_modules/lucide-react/dist/esm/icons/share.mjs'
}
[Error: EMFILE: too many open files, open '/var/task/node_modules/lucide-react/dist/esm/icons/share.mjs'] {
  errno: -24,
  code: 'EMFILE',
  syscall: 'open',
  path: '/var/task/node_modules/lucide-react/dist/esm/icons/share.mjs',
  page: '/robots.txt'
}
RequestId: ace94785-66ac-47ca-81d7-ff1ab00793bc Error: Runtime exited without providing a reason
Runtime.ExitError
```

Client side, nothing to go on: 

<img width="347" alt="image" src="https://github.com/lucide-icons/lucide/assets/1755190/20cefd17-d4ba-4bfc-bfd7-c589385a8fad">

## Resources

<img width="625" alt="image" src="https://github.com/lucide-icons/lucide/assets/1755190/8acfbc7a-77f4-4216-8fd5-2967b5226122">

Update; perhaps related?

- https://github.com/vercel/next.js/issues/52646

ThimoDEV:
+1

leeuwd:
Solved by adding the following to `next.config.js`:

```js
// Modularize Lucide icon imports; otherwise, Vercel
// throws 'too many open files' error (https://github.com/lucide-icons/lucide/issues/1446)
// @see https://nextjs.org/docs/architecture/nextjs-compiler#using-named-imports
modularizeImports: {
[...]

Issue #1457, opened 2023-07-25
Vercel/Astro/React/Node 'too many open files' error


guysolan:
This issue is similar to Issue [1446](https://github.com/lucide-icons/lucide/issues/1446). However, the issue is with Astro not NextJS, therefore the modularizeImports fix does not work.

Please note: I have never opened an issue before so apologies for any bad practice.

## Prerequisites

* Version: "lucide": "^0.261.0", "lucide-react": "^0.263.1",
* Are you running from source/main: Astro /dist directory
* Are you using a released build: Yes
* Operating system: Node.js 18.x
* Bits:

## Step to reproduce

npm create astro@latest

npx astro add react

npm install lucide-react

Deploy to  vercel using '@astrojs/vercel/serverless'

## Any message or error

Error message via Vercel logs:

```
Time
July 25 11:54:33.53 GMT+01:00
Status Code
500
Host

Request ID
4vpgm-1690282473526-b50ce645ed85
Request User Agent
Vercel Edge Functions
Level
Error

Cache
MISS
Type
Serverless Function

Function
/render
Location
Washington, D.C., USA (iad1)
Runtime
Node.js 18.x

Execution Duration / Limit
5ms / 10s
Memory Used / Limit
204 MB / 1024 MB
10:54:33 AM [ssr] Error: EMFILE: too many open files, open '/var/task/node_modules/lucide-react/dist/esm/icons/shirt.mjs' (x3)
```

## Resources

<img width="951" alt="image" src="https://github.com/lucide-icons/lucide/assets/56299900/fbb1fd74-01cf-4b35-9345-71b243d5e029">

josephkruse:
I'm also getting this error in Vercel as of yesterday. The modularizeImports fix didn't work for me in either.

ericfennis:
@guysolan Thanks for pointing this out.
Hmm, it looks like NodeJS module system with `.mjs` works differently and has a maximum of "open files". 
I will look if we can switch back to `.js` modules.. Will try if this fixes it.

cstrnt:
Downgrading to `"0.261.0"` solved the issue for me without any configuration :)

smolak:
> Downgrading to `"0.261.0"` solved the issue for me without any configuration :)

[...]

PR #1471, opened 2023-08-01
Revert .mjs modules in build for Lucide React

ericfennis:
Closes: #1446 #1457

Reverting switch to .mjs modules. Probably causing issues in NodeJS because it resolves in a different way than classic `.js` modules.

Not sure if this will fix it but we need to test this if this works.


sukalov:
if you don't wanna downgrade next, you can just specify the path to 'lucide-react' module.
smth like this `import { Sun, Moon } from '../../node_modules/lucide-react';`. worked for me

aikins01:
Fixed it for me!

PR #1374, opened 2023-06-18
Add drinks icons

danielbayley:
`goblet-crack` closely related to https://github.com/lucide-icons/lucide/pull/1372#issue-1761954212. Uses the same crack from https://lucide.dev/icons/heart-crack (existing `wine` glass is too small to fit…)

`mug` alts…

preview

Related to #1454.

---

Leaving `coffee-beans` here, for reference…

preview

PR #1472, opened 2023-08-01
Additional kitchen/appliances/amenities/oven icons

danielbayley:
Closely related to #1454. Also related to #1374, #1451, #1453 and #1461.

Kitchen/oven/hob alts:

preview


ericfennis:
I like the icons, but these icons are way too specific for the Lucide library.

PR #1485, opened 2023-08-07
Fix: Copy SVG properties into the child in react-native-svg

### LINKED ISSUE #1478: Icons stop working (black silhouettes) in React Native after over the air (OTA) updates
## Prerequisites

* Version: ^0.263.0 ((React native))
* Are you running from source/main: Both
* Are you using a released build: Yes
* Operating system: ios 16.5.1
* Bits:

## Step to reproduce

Install lucide react native as per the documentations. During development, the builds work fine, and production builds work fine **until an over the air update is served through Codepush. (expo updates could have the same issue too)

### Actual behavior

Icons black out and only shows the silhouettes after an over the air update. It might be a JS bundling issue that is resulting in this.

## Any message or error

No message or error

## Resources

See screenshots below:

Before an OTA update:
image

After an OTA update:
IMG_1325

alwyntan:
Copy SVG properties into its children.

Fixes #1478 , where icons are blacked out after OTA updates. Some related issues in `react-native-svg`, where children properties are not copied by default:
https://github.com/software-mansion/react-native-svg/issues/1284
https://github.com/software-mansion/react-native-svg/issues/1287

We might want to also consider removing the stroke, fill, etc off the defaultAttributes as they get over-ridden anyways, and to have them only apply to the children elements instead of the parent `<Svg>` element. I kept them there since I'm not sure if removing them off the <Svg> element would cause unwanted behavior in some cases.


ericfennis (approved):

ericfennis:
Tnx for the fix!

PR #1486, opened 2023-08-08
fix: generate alias files for nextjs import mapping

### LINKED ISSUE #1482: Import bug in Next.js (v13.4.13)
## Prerequisites

* Version: 0.264.0
* Are you running from source/main: nop
* Are you using a released build: nop
* Operating system: Mac OS
* Bits: -

## Step to reproduce

Import icon in a Next.js v13.4.13 (latest)

### Actual behavior

I see an error when importing the icon.

## Any message or error

```
Attempted import error: 'modularize-import-loader?name=CalendarDays&from=default&as=default&join=./icons/calendar-days!lucide-react' does not contain a default export (imported as 'CalendarDays').
```

--- 

It works well on v0.263.x, I suspect a bug in the new ESM export.

jguddas:
Resolves #1482.

https://github.com/lucide-icons/lucide/issues/1482#issuecomment-1667976412

### DIFF
--- a/.github/workflows/lucide-angular.yml
+++ b/.github/workflows/lucide-angular.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - packages/lucide-angular/**
+      - tools/build-icons/**
       - pnpm-lock.yaml
 
 jobs:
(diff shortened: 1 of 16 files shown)


ericfennis:
This is a good idea, this way we don't have to update those aliases files. I will test this with my NextJS PR

ericfennis:
Hmm rollup is removing the aliases modules during build 😞 , I searching for a solution to preserve those modules

PR #1493, opened 2023-08-09
feat: added fallback background to svg preview for browsers without prefers-color-scheme support

jguddas:
…refers-color-scheme support


ericfennis (approved):
Nice work, let go!

PR #1499, opened 2023-08-13
Add conditional exports property Lucide Svelte `package.json`

### LINKED ISSUE #1284: Lucide-svelte slow import of icon
## Prerequisites

* Version: Lucide-svelte 0.216.0
* Are you running from source/main: import using vite
* Operating system: win10
* Bits: 64

## Step to reproduce
- Install svelte-kit
  - npm create svelte@latest my-app
  - cd my-app
  - npm install
- Npm install lucide-svelte
- Import icon in svelte page
  - import { Edit } from 'lucide-svelte';
- Npm run dev
- In the browser navigate to the page where the icon is imported
- Notice how long it takes to reload the page presumably because vite takes a long time generate the server bundle

### Actual behavior
It shouldnt take this long to refresh the svelte page when an icon is imported in it.
Note this happens for both server and client bundles (when using dynamic import in svelte page)

## Any message or error

No console ouput

## Resources

No resources

### LINKED ISSUE #1403: Sveltekit runs out of memory.
Using sveltekit's "npm build" - all the icons need transforming - even if not used.
node_modules/.pnpm/lucide-svelte@0.252.0_svelte@3.59.2/node_modules/lucide-svelte/dist/svelte/icons/venetian-mask.svelte

Very slow of course, but also cause the builder to run out of memory :(

Steps To Reproduce

npm run build

### LINKED ISSUE #1475: lucide-svelte bundling all 1221 icons
I wanted to bring up something that has been on my mind for a little while, when I import an icon from lucide-svelte and build my sveltekit project, all 1221 icons are directly bundled with it. This results in build times of about 30 seconds (on my machine).

To be honest, I am not sure if this can be considered a "bug".

The workaround I found is to import the file directly from it's "source".

## Step to reproduce
+page.svelte
```svelte
<script lang="ts">
import {UserIcon} from "lucide-svelte"
</script>

<UserIcon/>
```
inside your terminal
```bash
$ npm run build
```

## Workaround
+page.svelte
```svelte
<script lang="ts">
  import UserIcon from "lucide-svelte/dist/svelte/icons/user.svelte"
</script>

<UserIcon/>
```
inside your terminal
```bash
$ npm run build
```

### Without Workaround
image
image
### With Workaround
image
image

ericfennis:
closes #1475 
closes #1403 
closes #1284


JanTrichter:
@ericfennis tried to check this out and run it locally but I cannot run the build command:
`RollupError: Could not resolve "./circle-slashed.svelte" from "src/icons/index.js"`

and this is the error message in the `index.js` file:

```
Unexpected use of file extension "svelte" for "./circle-slashed.svelte"eslint(import/extensions)
Unable to resolve path to module './circle-slashed.svelte'.eslint(import/no-unresolved)
```

ericfennis:
@JanTrichter YEah I didn't test this code yet. So tnx for testing!
Will check what's going on.

JanTrichter:
@ericfennis The error only appears in the `index.js` file, the index.ts is fine.

Here is a list of all the exports that don't work for me:
```js
export { default as CircleSlashed } from './circle-slashed.svelte';
export { default as SidebarClose } from './sidebar-close.svelte';
export { default as SidebarOpen } from './sidebar-open.svelte';
export { default as Sidebar } from './sidebar.svelte';
export { default as Slash } from './slash.svelte';
export { default as SortAsc } from './sort-asc.svelte';
export { default as SortDesc } from './sort-desc.svelte';
```

And here is a list of all the icons that have an alias:

```bash
grep -rnw 'icons' -e 'aliases' | grep -o 'icons/[a-zA-Z0-9-]*'
```
Output:
```
icons/braces
icons/circle-slash-2
icons/text-select
icons/panel-left
icons/panel-left-open
icons/ban
icons/arrow-up-narrow-wide
icons/arrow-down-wide-narrow
icons/edit-3
icons/edit-2
icons/edit
icons/sparkles
icons/panel-left-close
```

Looks like the alias could be a problem here.

JanTrichter:
@ericfennis I tried to make it work and after installing `@sveltejs/package` and adding the script to the npm scripts: 

```json
    "package": "svelte-package -i ./dist/svelte",
```

It did something after that but...

It looks like the `exports` field is not working for me or in general so I had to pass the `-i` flag. The generated output looks really similar to what you would normally get (without the `esm` folder tho and it's flattened so the `/svelte` folder is missing and it's all in the parent directory) but the `lucide-svelte.d.ts` is missing and I couldn't use the package after linking it.

ericfennis:
@JanTrichter Ok I think I fixed it. Added the correct exports to package.json and made some adjustments to the build config.
I release beta version on `lucide-svelte@0.269.0-beta.0`. I tested it on Vite and Sveltekit.
With this configuration I think I fixed #1403 as well. 

Let me know it this work for you!

```sh
pnpm add lucide-svelte@0.269.0-beta.0
```

JanTrichter:
@rbozan See my earlier comment and suggested fix in this conversation.

tools/rollup-plugins/plugins.js
```diff
@@ -14,7 +14,7 @@ const plugins = (pkg, minify, esbuildOptions = {}) =>
     license({
       banner: `${pkg.name} v${pkg.version} - ${pkg.license}`,
     }),
-    bundleSize(),
+    // bundleSize(),
```
jguddas, line 17: Why remove this?

ericfennis: Tnx, this was temporary. wiil place it back!

JanTrichter:
@ericfennis why was this completed? There are still two pending review comments?

ericfennis:
@JanTrichter which review comments do you mean. I fixed the Warning and the reverted the Bundle size one.

After that I tested everything once again in Sveltekit en Svelte with vite.

What are your doubts about the implementation?

JanTrichter:
@ericfennis my comments about the `default` and `types` in the `package.json` are still pending.

Gerschtli:
@ericfennis This change breaks SvelteKit SSR, see https://github.com/lucide-icons/lucide/issues/1536

Gerschtli:
> my comments about the default and types in the package.json are still pending.

FYI: You need to submit your review comments. Pending means, they are not published yet, so nobody except you can see them ;)

JanTrichter:
> @JanTrichter
> 
> > my comments about the default and types in the package.json are still pending.
> 
> FYI: You need to submit your review comments. Pending means, they are not published yet, so nobody except you can see them ;)

thanks @Gerschtli wasn't aware of GitHub handling it that way ^^

ericfennis:
@JanTrichter Tnx for your comments, these were crucial! I make a fix now!

JanTrichter:
@ericfennis that one's on me. I already commented that almost two weeks ago but left the comments on `pending` because I wasn't aware of how GitHub handles code reviews 💀

PR #1500, opened 2023-08-13
Sort `rows`/`columns` icons

danielbayley:
Related to #1055, #1156, #1391, #1498, and #1508.

`rows-3` and `columns-3` double as #1156 `panels-top-bottom` and `panels-left-right` variants, respectively.

PR #1498, opened 2023-08-13
Sort `table`/project icons

danielbayley:
Since `table` is also Projects right here on GitHub…

Untitled

…so `table-dashed` is [`project-template`](https://primer.style/design/foundations/icons/project-template-16) from the same Octicons.

Related to #1055, #1156, #1270, #1391, #1500 and #1508.

PR #1508, opened 2023-08-15
Add `between-*` icons

danielbayley:
Related to #1498 and #1500.


jguddas:
Not a fan of the 1px radius.

danielbayley:
> Not a fan of the 1px radius.

@jguddas The 1px radius is based on the existing https://lucide.dev/icons/?search=layout- icons, plus I think we’ve already gone through this routine… 😅  See https://github.com/lucide-icons/lucide/pull/1184#issuecomment-1529123208

ericfennis:
@danielbayley Can we compare it with 2px? Just curious how it will look.

karsa-mistmere:
I'm fine with the 1px, but I think we could name these `between-horizonal-start` and `between-vertical-start` to leave space for mirrored variants.

PR #1509, opened 2023-08-15
Add `touch-select` icons

danielbayley:
Similar/inspired by ChatGPT mobile…

IMG_5010

---

alts:

preview


ericfennis:
I'm not in favor of this one.
I think [`text-select`](https://lucide.dev/icons/text-select) is way better for this usecase. 

I close this one.

PR #1511, opened 2023-08-15
Optimise `pin` icons/add `pinned` variant/improve metadata

danielbayley:
An alternative way of showing the toggled state of `pin` vs `pinned`. The thinking here is that it might be more appropriate to appear as being `pinned` to a vertical surface…

pin

If we didn’t necessarily want to add a simply rotated `pin-horizontal` variant, I suppose we could just add `pinned` like this instead:

preview


ericfennis:
I chose this one. The design is odd to me. This is a bit weird way of showing something is "pinned". I think from UX point of view it's better to use color or a "fill" to indicate something is "pinned".

PR #1517, opened 2023-08-17
replaced SvelteComponentTyped with SvelteComponent

### LINKED ISSUE #1507: Migrate generated Svelte icons from `SvelteComponentTyped` to `SvelteComponent`
## Prerequisites

* Version: 0.268.0
* Are you running from source/main: no
* Are you using a released build: yes
* Operating system: macOS 12.6.5
* Bits: 64-bit

## Step to reproduce

1. Try to copy an existing icon's base type definition:

   ```ts
   import type { SvelteComponentTyped } from 'svelte';
   import { type IconProps, type IconEvents, Link2Icon } from 'lucide-svelte';

   const data: { example: string, icon: SvelteComponentTyped<IconProps, IconEvents, {}>, /* ... */ }[] = [
     { example: 'abc', icon: Link2Icon }, /* ... */
   ]
   ```
2. Note the deprecation from `SvelteComponentTyped`, with the following message:

   > `@deprecated` Use SvelteComponent instead. See PR for more information: https://github.com/sveltejs/svelte/pull/8512
3. Attempt to use `SvelteComponent` instead:

   ```diff
   -import type { SvelteComponentTyped } from 'svelte';
   +import type { SvelteComponent } from 'svelte';
   -{ example: string, icon: SvelteComponentTyped<IconProps, IconEvents, {}>, /* ... */ }
   +{ example: string, icon: SvelteComponent<IconProps, IconEvents, {}>, /* ... */ }
   ```

### Actual behavior

A TypeScript error is outputted as a result of performing step 3.

## Any message or error

> Type 'typeof Link2Icon' is missing the following properties from type 'SvelteComponent<IconProps, IconEvents, {}>': $$prop_def, $$events_def, $$slot_def, $capture_state, and 6 more.ts(2740)

## Resources

* [Migration guide](https://github.com/sveltejs/svelte/pull/8512)

EDIT: I see that this was noted here as well: https://github.com/lucide-icons/lucide/pull/1388#issuecomment-1605775698

---

On an unrelated note, could a TypeScript type that all icons can extend from be added as well? Perhaps something like the following:

```ts
export type LucideIconComponent = SvelteComponent<IconProps, IconEvents, {}>;
```

JanTrichter:
Closes #1507 

Build it and linked it to a small demo repo and it seems to work as before. 
Maybe we would need to change some documentation for some edge cases? @ericfennis 

On the same note: `SvgProps` is also [deprecated](https://github.com/sveltejs/language-tools/blob/master/docs/preprocessors/typescript.md#im-getting-deprecation-warnings-for-sveltejsx--i-want-to-migrate-to-the-new-typings):
image

Just taking a quick look but couldn't find a replacement in `svelte/elements` so that's something we could add to this PR as well once we find it.

### DIFF
--- a/packages/lucide-svelte/scripts/buildTypes.mjs
+++ b/packages/lucide-svelte/scripts/buildTypes.mjs
@@ -18,7 +18,7 @@ const writeDeclarationFile = (typesFile, directory, content) => {
 };
 
 const getComponentImport = (componentName) =>
-  `export declare class ${componentName} extends SvelteComponentTyped<IconProps, IconEvents, {}> {}\n`;
+  `export declare class ${componentName} extends SvelteComponent<IconProps, IconEvents, {}> {}\n`;
 
 const ICONS_DIR = path.resolve(currentDir, '../../../icons');
 const TYPES_FILE = 'lucide-svelte.d.ts';
@@ -27,7 +27,7 @@ const TYPES_FILE = 'lucide-svelte.d.ts';
 let declarationFileContent = `\
 /// <reference types="svelte" />
 /// <reference types="svelte-check/dist/src/svelte-jsx" />
-import { SvelteComponentTyped } from "svelte";
+import { SvelteComponent } from "svelte";
 
 interface IconProps extends Partial<svelte.JSX.SVGProps<SVGSVGElement>> {
   color?: string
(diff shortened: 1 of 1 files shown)


ericfennis (commented):
@JanTrichter Thanks for fixing this!
Do you know from which version of Svelte this is changed?
Maybe we need to adjust the `peerDependencies` in `package.json`.

JanTrichter:
@ericfennis it has been deprecated with the [Svelte 4.0.0 release](https://github.com/sveltejs/svelte/releases/tag/svelte%404.0.0) so we should be fine

ericfennis (approved):
Thanks for reply, Good to go!

PR #1551, opened 2023-09-02
Adjust name from "data-lucide-name" to "data-lucide" in documentation

### LINKED ISSUE #1523: HTML icon example seems to be wrong
Example: https://lucide.dev/icons/settings

```html
<i data-lucide-name="settings"></i>
```

Does not show anything, but if I remove `-name`, as it's described here https://lucide.dev/guide/packages/lucide, then it works

```html
<i data-lucide="settings"></i>
```

JanTrichter:
Closes #1523 

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other:

### Description
Update the HTML examples in the documentation to use the `data-lucide` instead of the deprecated `data-lucide-name` attribute. (updated see #1169)

## Before Submitting

### DIFF
--- a/docs/.vitepress/lib/createCodeExamples.ts
+++ b/docs/.vitepress/lib/createCodeExamples.ts
@@ -24,7 +24,7 @@ const getIconCodes = (): CodeExampleType => {
       codes: [
         {
           language: 'html',
-          code: `<i data-lucide-name="Name"></i>
+          code: `<i data-lucide="Name"></i>
 `,
         },
       ],


JanTrichter:
I also added `srcExclude` to the vitepress config so that a `.md` file can be used for the README. More infos for that can be found [here](https://vitepress.dev/reference/site-config#srcexclude).

ericfennis (approved):
Awesome work, thanks!

Issue #1576, opened 2023-09-26
Dynamic Import makes NextJS development server very slow


morrisseybr:
### Package

- [X] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [X] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

0.279.0

### Browser

- [ ] Chrome/Chromium
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [X] Other/not relevant

### Operating system

- [X] Windows
- [ ] Linux
- [ ] macOS
- [ ] Other/not relevant

### Description

I'm attempting to utilize the dynamicIconImports object to create a unified component for rendering icons based on an icon name prop in NextJS 13 with the app router. I followed the example provided in the documentation, but it significantly slows down my development server. The time it takes for the initial compilation has increased from ~8 seconds to ~22 seconds, and the time for hot reloading even minor changes (such as a simple text modification) has gone from ~800ms to ~12 seconds.

![image](https://github.com/lucide-icons/lucide/assets/48535379/6aa17f16-8dcf-4a7d-b8de-a11ab0240243)

I've tried changing the component from server-side to client-side rendering. I've also attempted to add the library to the transpilePackages list in the next.config.js file. I even experimented with using the dynamicIconImports object with a fixed name (ignoring the name prop), but none of these approaches seemed to help.

My solution to the problem was to rewrite the use of the next/dynamic function for importing the 'lucide-react' module and utilize the promise result to obtain the necessary icon. This restored the development server to its normal performance.

![image](https://github.com/lucide-icons/lucide/assets/48535379/de49a85b-444f-4c8e-bf97-e95742287541)

I believe that the way the dynamicIconImports feature is implemented causes the NextJS development server to read and process all the imports when it shouldn't be doing so during compilation.

I am planning to submit a pull request to enhance the consumption of Lucide.icons by introducing a simple IconsName type. Additionally, I will revise the example of dynamic usage within the NextJS environment to eliminate the use of the dynamicIconImports object.

Would anyone like to contribute to this issue I'm addressing? Your contributions would be greatly appreciated. Thank you!

### Steps to reproduce

[...]

PR #1580, opened 2023-09-27
Changed the exemple of dynamic icons using NextJS

morrisseybr:
This pr resolves the issue #1576.


dani-z:
Tried this locally, but unfortunately, it crashed my app. Are you sure this works with Next 13+ App router? Here's my error 
image

morrisseybr:
> Tried this locally, but unfortunately, it crashed my app. Are you sure this works with Next 13+ App router? Here's my error image

Can you show your implementation and the use of the example that you are doing?

jguddas (approved):
👍 from me.

IMO we should remove the `dynamicIconImports` file, it's an antipattern and only leads to performance bottlenecks and flickering.

And hey, you can do this other ways such as described by your change here.

KajSzy:
> > Tried this locally, but unfortunately, it crashed my app. Are you sure this works with Next 13+ App router? Here's my error image
> 
> Can you show your implementation and the use of the example that you are doing?

I think his issue is that `dynamicIconImports` uses kebab-case icon names, while your apporach requires PascalCase
I've stumbled at exactly the same issue in my project that I very recently moved from non-optimized generic component to the one using `dynamicIconImports` ;)
But as reported in issue the slowness of dev server brought me here

bernatfortet:
@KajSzy converting the name to pascal case solved my issue. Dev server is super fast now. Thanks!

ben519:
@morrisseybr this only works client-side. Any idea how to make it work within a Server Component?

docs/guide/packages/lucide-react.md
```diff
@@ -125,43 +126,31 @@ const Icon = ({ name, ...props }: IconProps) => {
...
     </Suspense>
   );
-}
+};
 
-export default Icon
+export default Icon;
 ```
 
 ##### NextJS Example
```
ericfennis, line 135: We should make separate sections for the App router and pages router.

Pages router should use the docs like it was before.

And create a new section specially targeting the App Router.
With this code example:
```tsx
import { LucideProps, icons } from 'lucide-react';

interface IconProps extends LucideProps {
  name: keyof typeof icons;
}

const Icon = ({ name, ...props }: IconProps) => {

  return <LucideIcon {...props} />;
};

export default Icon;
```

ericfennis (changes requested):

PR #1587, opened 2023-10-06
feat(icons): added globe-lock icon

karsa-mistmere:
- [x] New Icon

### Description
Added a globe-lock icon

### Icon use case
* VPN
* Secure connection
* Internet security
* Password/digital identity protection
* Privacy mode
* Data encryption

### Alternative icon designs
No viable alternates were made.

### Concept

### Author, credits & license
- [x] I've based them on the following Lucide icons: `globe-2` and `file-lock`

### Naming

### Design

## Before Submitting


jguddas:
Why not have the lock in the bottom right corner?

karsa-mistmere:
> Why not have the lock in the bottom right corner?

Because I tried that and found that it looked pretty damn awful compared to this. :D 
(Hence no _viable_ alternatives).

image

jguddas:
> > Why not have the lock in the bottom right corner?
> 
> Because I tried that and found that it looked pretty damn awful compared to this. :D (Hence the "no _viable_ alternatives").
> 
> image

True, I would be fine with changes to the `globe` icon if that helps.

ericfennis:
@karsa-mistmere Really nice icon!
The lock in the right bottom looks a bit better to me.

karsa-mistmere:
image

karsa-mistmere:
image

Not sure what you mean?

ericfennis:
@jguddas I think you can better post the changed SVG.

jguddas:
icon (19)

https://lucide-site-git-feat-added-edit-page-jguddas.vercel.app/edit?PHN2ZwogIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIKICB3aWR0aD0iMjQiCiAgaGVpZ2h0PSIyNCIKICB2aWV3Qm94PSIwIDAgMjQgMjQiCiAgZmlsbD0ibm9uZSIKICBzdHJva2U9ImN1cnJlbnRDb2xvciIKICBzdHJva2Utd2lkdGg9IjIiCiAgc3Ryb2tlLWxpbmVjYXA9InJvdW5kIgogIHN0cm9rZS1saW5lam9pbj0icm91bmQiCj4KICA8Y2lyY2xlIGN4PSIxMiIgY3k9IjEyIiByPSIxMCIgLz4KICA8cGF0aCBkPSJNMjEuNSAxNUgxN2EyIDIgMCAwIDAtMiAydjQuNSIgLz4KICA8cGF0aCBkPSJNMi4yIDEwSDRhMiAyIDAgMCAxIDIgMnYxYTIgMiAwIDAgMCAyIDIgMiAyIDAgMCAxIDIgMnY0LjgiIC8+CiAgPHBhdGggZD0iTTUuNCA0LjVBMyAzIDAgMCAwIDggN2EyIDIgMCAwIDEgMiAydi41YTIgMiAwIDEgMCA0IDBWOWEyIDIgMCAwIDEgMi0yaDQuNyIgLz4KPC9zdmc+

karsa-mistmere:
Here's a 100% comparison of all three versions:
image

ericfennis:
The second row looks the best to me. 

I don't know but it feels a bit nicer with the lock on the bottom. Maybe because the flat bottom side?

What do you think?

karsa-mistmere:
That's exactly why the first row looks better to me, it has slightly better centre of gravity. :)

icons/globe-lock.svg
```diff
@@ -0,0 +1,18 @@
...
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M7 3.34V5a3 3 0 0 0 3 3" />
+  <path d="M11 21.95V18a2 2 0 0 0-2-2 2 2 0 0 1-2-2v-1a2 2 0 0 0-2-2H2.05" />
+  <path d="M21.54 15H17a2 2 0 0 0-2 2v4.54" />
+  <path d="M12 2a10 10 0 1 0 9.54 13" />
+  <path d="M20 6V4a2 2 0 1 0-4 0v2" />
+  <rect width="8" height="5" x="14" y="6" rx="1" />
```
jguddas, line 17: icon - 2023-10-13T213723 981

```suggestion
  <path d="M14 22v-4a1 1 0 0 1 1-1h6a1 1 0 0 1 1 1v3a1 1 0 0 1-1 1h-9" />
  <path d="M15.7 9.5A15.3 15.3 0 0 0 12 2a14.5 14.5 0 0 0 0 20 10 10 0 1 1 9.862-11.658" />
  <path d="M20 17v-2a2 2 0 1 0-4 0v2" />
  <path d="M2 12h10.9" />
```

ericfennis:
@karsa-mistmere I'm ok with that!
@jguddas Do you have a final opinion on this?

jguddas:
> @karsa-mistmere I'm ok with that!
> @jguddas Do you have a final opinion on this?

What about my proposal, based on globe?

I'm not a super big fan of any of the proposals including mine, quite a lot of compromises.

I definitely think the name has to change if we want to base it on globe-2 rather than globe.

ericfennis:
I think the first and second row looks a bit more balanced. Some gaps between some continents looks a bit too big.

I agree on renaming this icon and `globe-2` to `earth` as we discussed in discord.

jguddas:
+1 for row 2

karsa-mistmere:
> > What about my proposal, based on globe?
>
> I think the first and second row looks a bit more balanced. Some gaps between some continents looks a bit too big.

I think @jguddas means this one: 
image

I made two very similar ones, but I'm not a fan of these, the paths are just too broken up:
image

jguddas:
I don't feel 100% about either option, what about ‘lab‘ing it?

ericfennis:
@jguddas Hmm I think the usecase is a valid usecase. I think we just need to agree on the styling.
@karsa-mistmere Can you make a decision, or do you want me to make the decision?

karsa-mistmere:
> @jguddas Hmm I think the usecase is a valid usecase. I think we just need to agree on the styling. @karsa-mistmere Can you make a decision, or do you want me to make the decision?

I still prefer my original design, since it means we don't have to also update `globe` to match, and I very much prefer our current globe icon to any of the proposed alternates.

jguddas:
I just think it's cramped and not all that pretty.

ericfennis (changes requested):
Ok I want to move on with this one.
I'm ok with the current design, and I think we should move forward.
I think we should rename `globe-2` to `earth` and call this one `earth-lock`.
I also think we should add the other `globe-lock` design. With the `globe` icon in the background.

jguddas:
I think it should be `planet-earth` so we keep our selves open for the material type earth.

karsa-mistmere:
> I think it should be `planet-earth` so we keep our selves open for the material type earth.

Do you mean this one?
image

jguddas:
> > I think it should be `planet-earth` so we keep our selves open for the material type earth.
> 
> Do you mean this one? image

I mean just instead of calling it `earth` we should call it `planet-earth` and `planet-earth-lock`

karsa-mistmere:
> > > I think it should be `planet-earth` so we keep our selves open for the material type earth.
> > 
> > 
> > Do you mean this one? image
> 
> I mean just instead of calling it `earth` we should call it `planet-earth` and `planet-earth-lock`

Yeah, but what exactly do you mean keep ourselves open for "the material type earth"?

jguddas:
> > > > I think it should be `planet-earth` so we keep our selves open for the material type earth.
> > > 
> > > 
> > > Do you mean this one? image
> > 
> > 
> > I mean just instead of calling it `earth` we should call it `planet-earth` and `planet-earth-lock`
> 
> Yeah, but what exactly do you mean keep ourselves open for "the material type earth"?

https://www.thesaurus.com/browse/earth

karsa-mistmere:
> > > > > I think it should be `planet-earth` so we keep our selves open for the material type earth.
> > > > 
> > > > 
> > > > Do you mean this one? image
> > > 
> > > 
> > > I mean just instead of calling it `earth` we should call it `planet-earth` and `planet-earth-lock`
> > 
> > 
> > Yeah, but what exactly do you mean keep ourselves open for "the material type earth"?
> 
> https://www.thesaurus.com/browse/earth

Haha, I totally thought you meant Material as in https://m3.material.io/styles/icons/overview

If and when we add this as an icon, `soil` or `ground` would be much more obvious choices for a name, so I think we're in the clear.

jguddas (approved):

ericfennis (approved):

PR #1589, opened 2023-10-07
fix/docs: release tags don't work for newly-created/updated icons

triphora:

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description
As per 3391b5b717a23312fe3ebc45ec2c40e66aa185a4, the format of new GitHub release tags is just the version number without the `v` prefix. As we can see on https://github.com/lucide-icons/lucide/releases?page=2, versions older than 0.266.0 use a `v` prefix. The current docs assume that all versions contain the `v` prefix; this PR fixes it so that the format of versions both from before and after this change are accounted for.

https://github.com/lucide-icons/lucide/assets/60205699/e414b050-543b-4c2a-81c9-5ecdf0bfe373

The "home hero" is also fixed so that it shows the `v` part to make it clear this is a version number.

## Before Submitting


jguddas (approved):

ericfennis (approved):
Awesome work tnx!

PR #1595, opened 2023-10-10
fix: gh-icon improvements

jguddas:
- Icon with backdrop is now no longer broken in dark mode
- If you add an icon, let's say `folder-example`, it will show `folder` as the backdrop.
- If you open the `/gh-icon` route without a backdrop name, it will still work.


ericfennis (approved):
Nice work!

PR #1603, opened 2023-10-12
Fixed import of toKebabCase helper function

SamstyleGhost:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other:

### Description

There were comments in multiple files asking to import the 'toKebabCase' helper function instead of redefining the function everywhere. So, fixed the import statements in all such instances and removed the existing comments asking for this to be done. Overall, not a feature or a bug fix, just some code cleanup.

## Before Submitting 


jguddas (changes requested):
Don't import things outside package scope.

ericfennis (changes requested):
I agree with @jguddas We should not import outside the package scope.
But the cleanup is nice, and we should do this in another way.

We could create a separate package in the packages directory called "utils". Which exports this function. This is also useful for other code we want to share between packages.

Would you like to create this package? In the `package.json` file the name should be "@lucide/utils" and the `private` property should be set on `true`. 
All dependent package get a new "dependency" field: `"@lucide/utils": "workspace:*",`

If you need help let me know.

SamstyleGhost:
@ericfennis Ok got it! I'll work on it! 
Thanks @jguddas

PR #1620, opened 2023-10-19
New icons: user-search and user-round-search

jmsv:
## What is the purpose of this pull request?

- [x] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

New icons:
- `user-search`
- `user-search-2`

### Icon use case 

- Searching for users
- Recruitment

### Concept 

- [x] I have provided valid use cases for each icon.
- [x] I have not added any a brand or logo icon.
- [x] I have not used any hate symbols.
- [x] I have not included any religious or political imagery.

### Author, credits & license

- [ ] The icons are solely my own creation.
- [ ] The icons were originally created in #<issueNumber> by @<githubUser>
- [x] I've based them on the following Lucide icons: `user-check`, `user-plus`, `user-x` etc. + corresponding `-2` variations, plus `search`
- [ ] I've based them on the following design: 

### Naming 

- [x] I've read and followed the [naming conventions](https://lucide.dev/guide/design/icon-design-guide#naming-conventions)
- [x] I've named icons by what they are rather than their use case.
- [x] I've provided meta JSON files in `icons/[iconName].json`.

### Design 

- [x] I've read and followed the [icon design guidelines](https://lucide.dev/guide/design/icon-design-guide)
- [x] I've made sure that the icons look sharp on low DPI displays.
- [x] I've made sure that the icons look consistent with the icon set in size, optical volume and density.
- [x] I've made sure that the icons are visually centered.
- [x] I've correctly optimized all icons to two points of precision.

## Before Submitting 

- [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [x] I've checked if there was an existing PR that solves the same issue.

PR #1630, opened 2023-10-22
feat: added renameIcon script

jguddas:
Examples:
https://github.com/lucide-icons/lucide/pull/1629
https://github.com/lucide-icons/lucide/pull/1628


jguddas:
`sh` or `mjs`, what is better?

jmsv:
> `sh` or `mjs`, what is better?

I'd go for `mjs` because it's OS-agnostic, consistent with other lucide scripts, and (imo at least) it's easier to work with, especially for more complex stuff

For example it'd be cool if this script could handle regex replace. Would be helpful for renaming lots of things with similar names, e.g. all of the `user*-2` icons as-discussed in #1620

ericfennis (commented):
Nice script!
I think we can also add this to the package.json scripts.

We should also add a documentation section in the contribution guidelines about this line.

PR #1642, opened 2023-10-27
Linting SVG files

ericfennis:
- [ ] New Icon
- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update

### Description
This adds a feature to lint SVG files. I've implemented the `@html-eslint` with some rules that match our way of writing the SVGs.

Because it is part of ESLint, lint-staged will check the format of SVGs.
I also added a GitHub action to check this when a PR is opened. GitHub will automatically add a comment if the formatting is incorrect.


.eslintrc.js
```diff
@@ -1,10 +1,27 @@
...
+  fill: 'none',
+  stroke: 'currentColor',
+  'stroke-width': 2,
+  'stroke-linecap': 'round',
+  'stroke-linejoin': 'round',
+}).map(([attr, value]) => ({
+  tag: 'svg',
+  attr,
+  value: String(value),
+}));
```
jguddas, line 15: importing from https://github.com/lucide-icons/lucide/blob/main/scripts/render/default-attrs.json would be better.
```suggestion
import DEFAULT_ATTRS from './scripts/render/default-attrs.json' assert { type: 'json' };
const requiredSVGAttrs = Object.entries(DEFAULT_ATTRS)
  .map(([attr, value]) => ({ tag: 'svg', attr, value: String(value) }));
```

.eslintrc.js
```diff
@@ -37,4 +55,35 @@ module.exports = {
...
+  overrides: [
+    {
+      files: ['./icons/*.svg'],
+      parser: '@html-eslint/parser',
+      rules: {
+        'prettier/prettier': 'off',
+        '@html-eslint/require-doctype': 'off',
+        '@html-eslint/no-duplicate-attrs': 'error',
+        '@html-eslint/no-inline-styles': 'error',
+        '@html-eslint/require-attrs': ['error', ...requiredSVGAttrs],
```
jguddas, line 67: Or just inline it…
```suggestion
        '@html-eslint/require-attrs': [
           'error',
             ...Object.entries(DEFAULT_ATTRS)
               .map(([attr, value]) => ({ tag: 'svg', attr, value: String(value) }))
        ],
```

.github/workflows/pull-request.yml
```diff
@@ -41,6 +41,22 @@ jobs:
           # output: ::$ANNOTATION_SEVERITY file=icons/accessibility.json,line=2,endLine=3,title=$ANNOTATION_TITLE::$ANNOTATION_DESCRIPTION%0A%0A+  "contributors": ["hi"],%0A@@ -13 +14 @@%0A+}%0A
       - name: Fail if contributors have changed
         run: git diff --exit-code -- icons/*.json
+
+  lint-icons:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
jguddas, line 48: You are checking out `main` due to this being run on `pull_request_target`.

.github/workflows/pull-request.yml
```diff
@@ -41,6 +41,22 @@ jobs:
...
+      - uses: pnpm/action-setup@v2
+        with:
+          version: 8
+      - uses: actions/setup-node@v3.8.1
+        with:
+          node-version: 18
+          cache: 'pnpm'
+
+      - name: Run ESLint
+        run: pnpm lint
```
jguddas, line 58: I would really try to avoid running `pnpm install` on `pull_request_target`, it's slow and not really secure.

Alternatively, we can call `https://lucide.dev/api/format`, write the results to the file do a diff like we do for the contributors.

The `/api/format` route would use [`optimizeSvg`](https://github.com/lucide-icons/lucide/blob/main/scripts/render/processSvg.mjs).

jguddas (changes requested):
Maybe split the PR into:
- Adding linting of svgs
- Run linting of svgs in PRs

.github/workflows/pull-request.yml
ericfennis: I've created a separate workflow with just the `pull_request` trigger. I will now lint all files.

ericfennis:
@jguddas Thanks for your feedback!

package.json
```diff
@@ -29,21 +29,22 @@
     "generate:contributors": "node ./scripts/updateContributors.mjs icons/*.svg",
     "generate:nextJSAliases": "node ./scripts/generateNextJSAliases.mjs",
     "postinstall": "husky install",
-    "lint:js": "eslint --ext .ts,.js,.mjs ./{packages/lucide,scripts}",
```
jguddas, line 32: Now the :js name does not make sense anymore.

tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
```
jguddas, line 1: Why do we have this file now?

ericfennis: When running Eslint running from the root, it reads also TS files, which requires a tsconfig in the root.

package.json
ericfennis: true, i've renamed it to `lint:es` 😛

PR #1656, opened 2023-10-31
fix export IconNode type

### LINKED ISSUE #1654: Export the `IconNode` type
### Package

- [X] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [ ] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] all JS packages
- [ ] site

### Description

Hi! Thanks a lot for this great project.

As a user of Lucide in a component library, I want to add the possibility for our users to pass Lucide icon objects to some of the widgets `icon` property. Unfortunately, the `IconNode` seems to only be used internally. Would it possible to export it for TS projects?

(I also have [another question](https://github.com/lucide-icons/lucide/discussions/1594) if you feel like it 😁)

### Use cases

Consuming Lucide objects from projects.

### Checklist

- [X] I have searched the existing issues to make sure this bug has not already been reported.

Sukomal07:
fixes #1654

### DIFF
--- a/packages/lucide/rollup.config.mjs
+++ b/packages/lucide/rollup.config.mjs
@@ -5,7 +5,7 @@ import pkg from './package.json' assert { type: 'json' };
 
 const outputFileName = pkg.name;
 const outputDir = 'dist';
-const inputs = ['src/lucide.ts'];
+const inputs = ['src/lucide.ts', 'src/types.ts'];
 const bundles = [
   {
     format: 'umd',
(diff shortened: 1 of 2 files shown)


packages/lucide/rollup.config.mjs
```diff
@@ -5,7 +5,7 @@ import pkg from './package.json' assert { type: 'json' };
 
 const outputFileName = pkg.name;
 const outputDir = 'dist';
-const inputs = ['src/lucide.ts'];
+const inputs = ['src/lucide.ts', 'src/types.ts'];
```
ericfennis, line 8: Please revert this.

ericfennis (changes requested):
I've just tested this. Unfortunately this is not going to work, since these inputs are used to bundle. In the bottom of this file we bundle the types, and that can be only one file.


So what we should do is export the types in: `packages/lucide/src/lucide.ts`.
Put this line at the bottom and this is solved.
```ts
/*
 Types exports.
*/
export * from './types';
```

Sukomal07:
@ericfennis done

ericfennis (approved):
Awesome work!

Sukomal07:
> Awesome work!

Thanks for help

olaf-k:
that was fast, thanks a lot guys!

PR #1663, opened 2023-11-06
`lucide-vue` Deprecation warning

ericfennis:
### Description
Added deprecation warning for. `lucide-vue`.
Vue v2 will be EOF at the end of 2024. We won't continue maintaining this package.


packages/lucide-vue/src/createVueComponent.ts
```diff
@@ -47,6 +49,13 @@ export default (iconName: string, iconNode: IconNode): Component => ({
       children = [],
     },
   ) {
+    if (showDeprecationWarning) {
+      console.warn(
+        '[Lucide Vue] This package will be deprecated end of 2023. Please use upgrade to Vue 3 and use the latest lucide package for Vue.',
```
mittalyashu, line 54: `Please upgrade to Vue 3 and use the latest lucide package for Vue.'`

ericfennis: Thanks!

PR #1664, opened 2023-11-06
Fixed Illustrator template link + fixed typo

zaaakher:
## What is the purpose of this pull request?

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other:

### Description

Fixed the links to the Adobe Illustrator template file in `illustrator-guide.md` and `CONTRIBUTE.md` 

Although I couldn't find a way to access the illustrator template from the website domain (i.e. https://lucide.dev/guide/illustrator-template.ai) so I just used the github link.

I also fixed a small typo in `illustrator-guide.md`

## Before Submitting 

- [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [x] I've checked if there was an existing PR that solves the same issue.

PR #1689, opened 2023-11-18
Improve Adobe Illustrator template

danielbayley:

- [ ] New Icon
- [ ] Bug fix
- [x] New Feature
- [x] Documentation update
- [x] Other: Improve Adobe Illustrator template

### Description

- Remove all unnecessary Swatches, Strokes, Symbols and Graphic Styles
- Add useful Graphic Styles to easily toggle between shape/guide stroke and black stroke with the correct properties, which can be useful when designing more complex icons
- Add base shapes for reference, with correct rounding
- Changed example shape on Draw layer to existing arrow icon
- Fixed builtin grid by changing gridline every 72px with 1 subdivision (useless), to gridline every 1px. It might be preferable for some users to use the builtin grid rather than an actual Grid layer with guides, particularly as this can be toggled seperately with a hotkey (I _think_ certain preferences are saved with the document)
- All colours now match Lucide red (#F56565), also added as Swatch
- Convert guide layers to template layers

## Before Submitting 

---

screenshot

---

Related to #1664.


ericfennis (approved):
Nice work @danielbayley !

PR #1756, opened 2023-12-27
Cleanup tools

nix6839:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other: Clean up about tools

### Description
I haven't applied formatting yet to make it easier to view the diff. If you agree with this PR, I can also include formatting in the commit.
- Bump pnpm version
- Use prettier instead of eslint-plugin-prettier for formatting (According to the [official documentation of Prettier](https://prettier.io/docs/en/integrating-with-linters.html#notes), it is generally not recommended to use eslint-plugin-prettier. Additionally, by using Prettier directly, we can easily format not only JavaScript files, but also other types of files such as JSON, Markdown, and more.)
- Upgrade root prettier version to 3
- Fix some errors about tools

## Before Submitting 


lint-staged.config.mjs
```diff
@@ -0,0 +1,23 @@
+/**
+ * @param {string[]} filenames
+ * @returns {string}
+ */
+const filenamesToAjvOption = (filenames) => filenames.map((filename) => `-d ${filename}`).join(' ');
+
+/** @satisfies {import('lint-staged').Config} */
```
jguddas, line 7: We should run eslint and typecheck here as well IMO.

nix6839: After this PR is merged, I will submit a separate PR related to ESLint. How about we handle it together at that time?

packages/lucide-angular/.eslintrc.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
         'plugin:@typescript-eslint/recommended',
         'plugin:@angular-eslint/recommended',
         'plugin:@angular-eslint/template/process-inline-templates',
+        'prettier',
```
jguddas, line 11: That feels wrong, why is that here?

nix6839: > Turns off all rules that are unnecessary or might conflict with [Prettier](https://github.com/prettier/prettier).
> -- https://github.com/prettier/eslint-config-prettier

What did you feel was wrong?

ericfennis (approved):

PR #1792, opened 2024-01-10
Add `grip-down-right` icon

ocavue:

- [x] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

This PR adds a new `grip-down-right` icon, shown below.

image

### Icon use case 

This kind of icon is commonly used as a resize handle, at the bottom right corner of a resizable window. 

- Example 1: https://stackoverflow.com/questions/55166626/how-are-dots-in-the-corner-called-and-how-to-disable-them)
- Example 2: https://superuser.com/questions/614020/where-does-the-resize-handle-in-browser-textboxes-come-from

### Alternative icon designs 

For the use case of the resize handle, browsers show multiple slashes at the bottom right corner. It seems that Lucide doesn't have this either so I will submit a [separate PR](https://github.com/lucide-icons/lucide/pull/1793) for this icon. 

image

image

### Concept 

### Author, credits & license

- [x] I've based them on the following Lucide icons: `grip`

### Naming 

### Design 

## Before Submitting 


jguddas (changes requested):
- Icon is not visually centered
- Name describes usecase rather than content

karsa-mistmere:
> * Icon is not visually centered
> * Name describes usecase rather than content

I think in this instance the icon not being centered is intentional. 🤔

ocavue:
> Icon is not visually centered

I don't see a way to make it visually centered. It's just an icon that uses a corner of the space, like existing [rss](https://lucide.dev/icons/rss), [move-3d](https://lucide.dev/icons/move-3d) and [triangle-right](https://lucide.dev/icons/triangle-right).

> Name describes usecase rather than content

There are already three [`grip` icons](https://lucide.dev/icons/?focus=&search=grip). It seems that my `grip-down-right` just follows the existing pattern.

ericfennis (approved):
I don't see why we shouldn't add this

PR #1793, opened 2024-01-10
Add `handle-down-right` icon

ocavue:

- [x] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

This PR adds a new `handle-down-right` icon, shown below.

image

### Icon use case 

This is to represent a resize handle at the bottom right corner of a resizable object (e.g. `<textarea>`). (For this reason, this icon is not centered). 

Browsers would display two (or more, depending on the browser) slashes for CSS [`resize`](https://developer.mozilla.org/en-US/docs/Web/CSS/resize) rule.

### Alternative icon designs 

 

image

### Concept 

### Author, credits & license

- [x] I've based them on the following Lucide icons: `slash`
- [x] I've based them on the following design: 
  - [SmartIcons Glyph / angle-2](https://yesicon.app/si-glyph/angle-2) Apache 2.0 
  - [Vaadin Icons / handle-corner](https://yesicon.app/vaadin/handle-corner) CC BY SA 4.0 
    

### Naming 

### Design 

## Before Submitting 


jguddas (changes requested):
- Icon is not visually centered
- Name describes usecase rather than content

damien-schneider:
"In my opinion, 'handle-down-right' is more intuitive. In the Phosphor library, it is named 'notche,' but I find this less intuitive.

Regarding the positioning, I don't have a strong opinion. As for the number of slashes, I believe that 2 is much better and more in line with the spirit of the library."

ericfennis:
@jguddas I think looking at the usecase of this icon, I think positioning it centered makes it hard to use this icon. 

For example, if you want to position it with CSS, it is intuitive to use it like this: 
```css
.icon-handle-down-right {
  position: absolute;
  bottom: 0;
  right: 0;
}
```

If we center this, you need to pixel-position it. Which is annoying from DX point of view. We had a related discussion before with all the `signal` icons. We decided to prefer DX over the guidelines.

For naming, we like to use generic names for what it represents.
But not sure what generic names we should use for this. Since it is hard to for two little lines. I can imagine we want to have other variants as well for other corners. 
Since it represents "grip lines" I think `grip-lines-right-down` will work.

jguddas:
I'm against adding this.

damien-schneider:
> I'm against adding this.

 You know that I'm using lucid for a full complex SaaS, and the only icon I still need to have an other library (making it non-consistent) is this one, so this icon is needed, really....

karsa-mistmere:
I agree with @ericfennis and I'm okay with adding this as either `grip-lines-right-down`, `grip-lines-bottom-right` or `grip-lines-sw`.

ericfennis:
Not sure why @jguddas is against it. It has a valid use case. And seems like a nice addition next to the current `grip` icons.

damien-schneider:
I really don't understand why it has been close, this icon has a real use case and is missing to the lib, very sad that it hasn't been implemented 😕

ericfennis (approved):

karsa-mistmere:
As discussed with @ericfennis, we should brainstorm more about the name of this icon.

We were in agreement, that it should be `*-bottom-right` as opposed to `*-down-right`, since "down" implies a direction or movement, rather than placement.

Some alternate suggestions I have in order of preference:
- `corner-slashes-bottom-right`
- `corner-grip-bottom-right`
- `slashes-bottom-right`
- `grip-bottom-right`
- `handle-bottom-right`

PR #1811, opened 2024-01-17
fix: Add the license banner inside of the script tag instead of an HTML comment

AdrianGonz97:
Fixes https://github.com/lucide-icons/lucide/pull/1707#issuecomment-1894976168

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description
It turns out that HTML comments in Svelte files turn into whitespace during SSR, but *not* in CSR. This leads to an extra whitespace appearing during initial renders, followed by the whitespace disappearing during hydration. 

Here's a small demo:
img

This is certainly more of a bug on Svelte's end, but a fix for now is to move the banner inside of the script tag instead. 

---

I rearranged the build scripts to inject the banner during icon creation instead (just for svelte files, other file types will have the banner appended at the end of the build) to keep it simple and avoid having to do any fancy string manipulations.

## Before Submitting 


packages/lucide-svelte/scripts/license.mjs
```diff
@@ -0,0 +1,17 @@
...
+export function getJSBanner() {
+  const prefixedBanner = BANNER.split('\n')
+    .map((line) => ` * ${line}`)
+    .join('\n');
+
+  return `/**
+${prefixedBanner}
+ */
+`;
+}
```
jguddas, line 17:
```suggestion
export function getJSBanner() {
  return `/**
 * @license ${pkg.name} v${pkg.version} - ${pkg.license}
 *
 * This source code is licensed under the ${pkg.license} license.
 * See the LICENSE file in the root directory of this source tree.
 */
`;
}
```

jguddas (changes requested):

jguddas (approved):

packages/lucide-svelte/scripts/appendLicense.mjs
```diff
@@ -32,18 +27,13 @@ for (const file of files) {
...
   }
 
   // Places icon block comment at the top of the Svelte component class
-  if(/icons\/(.*?)\.svelte\.d\.ts/.test(filepath)) {
-    const svelteFilepath = filepath.replace('.d.ts', '')
+  if (/icons\/(.*?)\.svelte\.d\.ts/.test(filepath)) {
+    const svelteFilepath = filepath.replace('.d.ts', '');
     let svelteFileContents = await readFile(svelteFilepath, { encoding: 'utf-8' });
 
     const blockCommentRegex = /\/\*\*[\s\S]*?\*\//;
```
ericfennis, line 39: We need to change this regex, this now also matches with the license block comment which results in twice adding the license block comment. 

Here is the correct new regex:
``` suggestion
    const blockCommentRegex = /\/\*\*\n\s\*\s(@component\s@name)[\s\S]*?\*\//;
```

packages/lucide-svelte/scripts/appendLicense.mjs
```diff
@@ -1,13 +1,8 @@
-import { getCurrentDirPath } from '../../../scripts/helpers.mjs';
```
ericfennis, line 1: I think we can better rename this to `appendBlockComments.mjs`. I've also added other block comments for the icon components.

ericfennis (changes requested):

ericfennis (approved):
Awesome!

isker:
It's fixed for me. Thank you!

Issue #1832, opened 2024-01-29
Play icon doesn't seem centered


bernaferrari:
### Package

- [X] lucide

### Version

0.289.1

### Browser

- [X] Other/not relevant

### Operating system

- [X] Other/not relevant

### Description

The bounds make play button uncentered. Phospor gets it right

Lucide:
![image](https://github.com/lucide-icons/lucide/assets/351125/147e17a4-59c0-4c0f-adaf-4269e99d9d9e)

<img width="70" alt="image" src="https://github.com/lucide-icons/lucide/assets/351125/1d5eaf2c-881b-4ee4-b15e-9ea8f2b5da27">

Phosphor:
<img width="71" alt="image" src="https://github.com/lucide-icons/lucide/assets/351125/5c02dd7f-04b3-446b-8a9c-6df9e89282c8">
<img width="68" alt="image" src="https://github.com/lucide-icons/lucide/assets/351125/3d6111bb-4cc1-4e9c-b9f7-2d6f1b84f430">

### Checklist

- [X] I have searched if someone has submitted a similar issue before and there weren't any. (Please make sure to also search closed issues, as this issue might already have been resolved.)

PR #1834, opened 2024-01-29
fix(icons): `play` icon uncentered

bernaferrari:
Fix https://github.com/lucide-icons/lucide/issues/1832

Lucide (black) vs Phosphor (red) now:
image

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### DIFF
--- a/icons/play.svg
+++ b/icons/play.svg
@@ -9,5 +9,5 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <polygon points="5 3 19 12 5 21 5 3" />
+  <polygon points="7 3 21 12 7 21 7 3" />
 </svg>
--- a/icons/play.svg
+++ b/icons/play.svg
@@ -9,5 +9,5 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <polygon points="7 3 21 12 7 21 7 3" />
+  <polygon points="6 3 20 12 6 21 6 3" />
 </svg>


icons/play.svg
```diff
@@ -9,5 +9,5 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <polygon points="5 3 19 12 5 21 5 3" />
+  <polygon points="7 3 21 12 7 21 7 3" />
```
karsa-mistmere, line 12:
```suggestion
  <path d="m6 3 15 9-15 9z" />
```

I think moving it 2px to the right would make it appear even more off centre in most other standard use cases (e.g. on square buttons), but maybe 1px to the right and a 1px increase in width to make the triangle more equilateral would do the trick without drastically messing up vertical alignment to `square`:
image
(current vs. 1px to the right vs. 2px to the right vs. with proposed changes)


image
image
(current vs. @bernaferrari vs. @karsa-mistmere)

karsa-mistmere: > Middle one looks best to me, pure gut feel tho.

Which middle one? 😄

jguddas: image
The middle one of those.

ericfennis: The middle one looks the best to me as well.
@karsa-mistmere Is the middle one the suggestion you proposed?

karsa-mistmere: Nah, the middle one is simply 1px to the right, my proposal is the rightmost one:

image
image

ericfennis: The second one and the most right one look both good to me 😅

karsa-mistmere: Agreed, either of these, I'd also prefer an equilateral(ish) play icon, but it's not a must.

ericfennis: I vote for the second one.

ericfennis: @karsa-mistmere Can provide the code suggestion for second one?

karsa-mistmere: I've updated the PR.

karsa-mistmere (approved):

ericfennis (approved):

PR #1638, opened 2023-10-26
Renamed user-*-2 to user-round-*

jmsv:
## What is the purpose of this pull request?

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other:

Rename icons

### Description

Renamed all `user-*-2` icons to `user-round-*` as-discussed in #1620, using script from #1630:

```sh
node ./scripts/renameIcon.mjs user-2 user-round
node ./scripts/renameIcon.mjs user-check-2 user-round-check
node ./scripts/renameIcon.mjs user-circle-2 user-round-circle
node ./scripts/renameIcon.mjs user-cog-2 user-round-cog
node ./scripts/renameIcon.mjs user-minus-2 user-round-minus
node ./scripts/renameIcon.mjs user-plus-2 user-round-plus
node ./scripts/renameIcon.mjs user-search-2 user-round-search
node ./scripts/renameIcon.mjs user-square-2 user-round-square
node ./scripts/renameIcon.mjs user-x-2 user-round-x
node ./scripts/renameIcon.mjs users-2 users-round
```

- [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [x] I've checked if there was an existing PR that solves the same issue.

PR #1847, opened 2024-02-03
Add `radical` icon (square root)

smnandre:
- [x] New Icon 
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

Add a "square root" icon, in two formats like the other math operators

### Icon use case 

* Buttons for calculator
* Section/tags for sorting math lessons

### Concept 

### Author, credits & license
- [x] I've based them on the following Lucide icons: activity, percent-square

### Naming 

### Design 

## Before Submitting 


jguddas:
The square-root feels a bit big.

smnandre:
> The square-root feels a bit big.

On both icons ? Do you prefer like this ? 

square-root-smaller

gruckion:
I like it

smnandre:
Capture d’écran 2024-02-03 à 17 50 03
Capture d’écran 2024-02-03 à 17 46 39

icons/square-root-square.svg
```diff
@@ -0,0 +1,14 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M19 3H5a2 2 0 0 0-2 2v14c0 1.1.9 2 2 2h14a2 2 0 0 0 2-2V5a2 2 0 0 0-2-2Z"/>
+  <path d="M7 12h2l2 5 2-10h4"/>
```
jguddas, line 13:
```suggestion
  <path d="M7 12h2l2 5 2-10h4" />
  <rect x="3" y="3" width="18" height="18" rx="2" />
```

jguddas (changes requested):

jguddas:
And the name should be `radical` instead of `square-root`.

https://en.wikipedia.org/wiki/Radical_symbol

Besides the IMO better fitting name, it also doesn't mess with the group name `square`.

icons/square-root-square.svg
smnandre: Thank you @jguddas

jguddas (approved):

karsa-mistmere:
Nice work!

I think one final thing is in order: `radical-square` should be `square-radical` instead.
(Something that in the near future will be added to our naming guidelines).

smnandre:
> Nice work!
Thank you (and all the feedback!)
 
> I think one final thing is in order: `radical-square` should be `square-radical` instead. (Something that in the near future will be added to our naming guidelines).

Ok! (i looked other icons, but said to myself it would be better if inversed, so... perfect :) )

karsa-mistmere (approved):

smnandre:
Not sure to understand the linter message

> Contributors have changed!
> 
> Don't add people who have only performed automatic optimizations.
> 
> +}

Is there something i should do there ?

jguddas (approved):

ericfennis (approved):

danielbayley:
@smnandre Nice, I have almost identical in my drafts… Should plain `radical` not also be horizontally centred though?

> I think one final thing is in order: `radical-square` should be `square-radical` instead. (Something that in the near future will be added to our naming guidelines).

@karsa-mistmere My line of thinking with previous naming was the other way around… The reasoning being that in a sorted list, having to search through _many_ `square-*` icons would be much less user friendly than having them more grouped by `name-square`… What do you think? cc: @ericfennis @jguddas

smnandre:
> @smnandre Nice, I have almost identical in my drafts… Should plain `radical` not also be horizontally centred though?

It was an attempt to improve a bit the readability of the shape... all my respect for spotting it :) 

Capture d’écran 2024-02-10 à 03 39 39


Before(svg) before 
After(svg) after 

So you prefer the centered one ? @danielbayley 

For the name tell me, and i'll change the names :)

karsa-mistmere:
I've been thinking that possibly we should also consider consistency with `activity` and keep the lower end 1px below the base line, this would also improve vertical centering. 🤔
image
image

smnandre:
> I've been thinking that possibly we should also consider consistency with `activity` and keep the lower end 1px below the base line, this would also improve vertical centering. 🤔

I tried to be as aligned as possible with the other mathematical symbols, the ones that probably would be near this one: cf https://github.com/lucide-icons/lucide/pull/1847#issuecomment-1925394067

danielbayley:
> For the name tell me, and i'll change the names :)

The order needs to be agreed, and codified into the guidelines, as @karsa-mistmere mentioned…

karsa-mistmere:
> The order needs to be agreed, and codified into the guidelines, as @karsa-mistmere mentioned…

I'm pretty sure, @ericfennis, @jguddas and I have already agreed upon `square-*` vs. `*-square`, we just haven't had time to put it in writing yet. So `square-radical` is definitely the way to go.

smnandre:
Ok so i don't have to re-rename them! Tell me if you need anything else :)

jguddas (approved):

icons/radical.svg
```diff
@@ -0,0 +1,13 @@
...
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M3 12h4l3 8 3-16h7" />
```
karsa-mistmere, line 12: I think we should add an extra pixel of height to the bottom both to better match `activity` and to adjust the centre of gravity vertically:
```suggestion
  <path d="M3 12h4l3 9 4-17h7" />
```

danielbayley:
> I'm pretty sure, @ericfennis, @jguddas and I have already agreed upon `square-*` vs. `*-square`

@karsa-mistmere Ah ok, where can I catch up on the reasoning, was there an issue?

jguddas:
> > I'm pretty sure, @ericfennis, @jguddas and I have already agreed upon `square-*` vs. `*-square`
> 
> @karsa-mistmere Ah ok, where can I catch up on the reasoning, was there an issue?

#1874

karsa-mistmere:
> @karsa-mistmere Ah ok, where can I catch up on the reasoning, was there an issue?

Originally, there was some discussion in https://github.com/lucide-icons/lucide/pull/1638 regarding this matter, most of the reasoning for why we went with what we did is there, on Discord we just formalized this decision into [some kind of rule](https://github.com/lucide-icons/lucide/pull/1874) that would be easy to follow by our contributors and hopefully result in reasonably consistent names.

smnandre:
Is there something more i can do on this one ? :)

karsa-mistmere (approved):

PR #967, opened 2023-02-28
Adds `brain` (`ai`) icons

karsa-mistmere:
closes #612

PR #1902, opened 2024-02-20
feat(icons): redesign brain and related icons

karsa-mistmere:

- [x] Icon redesign

### Description
The current brain icon is a bit too detailed in some places but not detailed enough in others and kind of doesn't really look like a brain at all. This PR updates it to look a bit more stylised and at the same time somewhat more organic.

### Alternative icon designs 
image

### Author, credits & license

- [x] I've based them on the following Lucide icons: the same ones I originally created in #967

### Design 

## Before Submitting 


jguddas:
I never thought I would say that, but I think it feels to fun and cute, maybe we could decutify it?

icons/brain.svg
ericfennis, line 1: On 100% scale, the center curved line looks a bit like a triangle. Can we try one with a longer line?

karsa-mistmere: I've updated the PR, let me know what you think.

jguddas: Feels better to me with the longer line.

ericfennis (approved):

PR #1907, opened 2024-02-22
Upgrade lucide-angular to angular 17

ekowbaah:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other:

### Description
Upgrade lucide-angular to angular version 17.
This PR: includes upgrading angular to version 17 and removing unused packages (protractor &  puppeteer) which are not needed anymore.
Unfortunately, this upgrade is not backward compatible.

## Before Submitting 


karsa-mistmere:
But why is it necessary to upgrade the package in the first place? 🤔

I'm currently using `lucide-angular` with multiple Angular 17 projects without any problems, have you maybe run into any issues? If so, what are they?

ekowbaah:
@karsa-mistmere I was thinking it was necessary since we are 4 major versions behind or you would like to wait until it's no longer compatible?

karsa-mistmere:
> @karsa-mistmere I was thinking it was necessary since we are 4 major versions behind or you would like to wait until it's no longer compatible?

Pretty much, yeah. There's no point dropping support for earlier versions if the package built with v13 works just fine with all the later versions. If and when we turn out to be incompatible with v18 (or v23, or v65) we can figure out which is the earliest version of Angular we can build the package with that will also fix that for the time being theoretical incompatibility issue.

The thing is, as of today, any other supported version of Angular is used in more projects than 17 is: https://www.npmjs.com/package/@angular/cli?activeTab=versions

Upgrading our package to 17 would mean that most (if not all) of these people would be unable to use the latest version of Lucide in their Angular projects.

PR #1915, opened 2024-02-24
fix: readded gap between x-ray preview comment icons.

jguddas:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other: Preview comment improvement

### Description

There used to be a gap between icons in the x-ray section.

But it is gone, and it doesn't look great to have them bud against each other.

Example:
image

## Before Submitting 

### DIFF
--- a/scripts/generateChangedIconsCommentMarkup.mjs
+++ b/scripts/generateChangedIconsCommentMarkup.mjs
@@ -26,7 +26,7 @@ const getImageTagsByFiles = (files, getBaseUrl, width) =>
 
       return `<img title="${file}" alt="${file}" ${widthAttr} src="${url}/${base64}.svg"/>`;
     })
-    .join('');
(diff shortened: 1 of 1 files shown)


ericfennis (approved):

scripts/generateChangedIconsCommentMarkup.mjs
```diff
@@ -26,7 +26,7 @@ const getImageTagsByFiles = (files, getBaseUrl, width) =>
 
       return `<img title="${file}" alt="${file}" ${widthAttr} src="${url}/${base64}.svg"/>`;
     })
-    .join('');
+    .join(width.includes('%') ? ' ' : '');
```
karsa-mistmere, line 29: This will squash previews on mobile devices, why not keep them at 400px, but still add the space?

PR #1906, opened 2024-02-22
chore(icons): update icon names to match naming guidelines no. 7 to 9

karsa-mistmere:
## What is the purpose of this pull request?
- [x] Bug fix

### Description
Our naming guidelines were updated in #1874, but a lot of icons don't yet match the new official naming scheme, this PR updates most (if not all) of these icons, while also adding backward compatible aliases.

A `renamePattern` helper script was also created that helps streamline these batched renames, e.g. most of the hard work was done via running:
```bash
pnpm renamePattern "^([a-z-]+)-(square|circle|diamond|octagon|hexagon|triangle)$" "\2-\1"
```

## Before Submitting 

- [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [x] I've checked if there was an existing PR that solves the same issue.

PR #1916, opened 2024-02-24
Update `power-circle` icon

jguddas:
- [x] Other: Icon update

### Description
Hand optimized icon

## Before Submitting


icons/power-circle.svg
```diff
@@ -9,7 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <circle cx="12" cy="12" r="10" />
   <path d="M12 12V6" />
-  <path d="M8 7.5A6.1 6.1 0 0 0 12 18a6 6 0 0 0 4-10.5" />
+  <path d="M8 7.5a6 6 0 1 0 8 0" />
+  <circle cx="12" cy="12" r="10" />
```
karsa-mistmere, line 14: I've also updated this in #1906, I'm pretty sure this should match `square-power`:

image


```svg
  <circle cx="12" cy="12" r="10" />
  <path d="M12 12V7" />
  <path d="M16 9a5 5 0 1 1-8 0" />
```

jguddas: Yeah, that looks a lot better with the smaller one, let's get the rename PR merged and then do the change for you suggest here.

PR #1918, opened 2024-02-24
fix(icons): update `pentagon` icon

jguddas:
- [x] Other: Icon update

### Description
Hand optimised corners to have 2px radius.

## Before Submitting


karsa-mistmere:
I'm not super happy about the optical volume and centre of gravity of this icon, I feel like maybe it should be scaled up towards the top as such:
image

What do you think?

icons/pentagon.json
```diff
@@ -1,7 +1,8 @@
 {
   "$schema": "../icon.schema.json",
   "contributors": [
-    "danielbayley"
+    "danielbayley",
+    "jguddas"
```
ericfennis, line 5: Name shouldn't be here, It's not a design change

jguddas: If you do manual changes/work that should count as a contribution IMO.

ericfennis:
@karsa-mistmere To me it is not really centered if you place the icon in a single form.

ericfennis (dismissed):

karsa-mistmere:
> @karsa-mistmere To me it is not really centered if you place the icon in a single form.

I find it to be a bit too bottom-heavy. 🤔

image

ericfennis:
@karsa-mistmere ah, seeing this. I think I'm going with your proposal

ericfennis (approved):

PR #1919, opened 2024-02-24
fix(icons): update `drill` icon

jguddas:
- [x] Other: Icon update

### Description
Hand optimized icon

## Before Submitting


karsa-mistmere:
It would be nice to also fix the gap issue somehow, some ideas:
image

icons/drill.svg
```diff
@@ -9,11 +9,11 @@
...
 >
-  <path d="M14 9c0 .6-.4 1-1 1H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9c.6 0 1 .4 1 1Z" />
-  <path d="M18 6h4" />
+  <path d="M12 10v3a1 1 0 0 1-1 1H8" />
   <path d="M14 4h3a1 1 0 0 1 1 1v2a1 1 0 0 1-1 1h-3" />
+  <path d="M14 9a1 1 0 0 1-1 1H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9a1 1 0 0 1 1 1z" />
+  <path d="M18 6h4" />
   <path d="m5 10-2 8" />
-  <path d="M12 10v3c0 .6-.4 1-1 1H8" />
+  <path d="M5 22a3 3 0 0 1-3-3 1 1 0 0 1 1-1h7a1 1 0 0 1 1 1v2a1 1 0 0 1-1 1z" />
```
ericfennis, line 17: hmm it's adding more code, why should we change this?

ericfennis: Can you explain what the benefit is, since it has the same visual result?

jguddas: We should not use curves to approximate arcs, we just use the arc for arcs.

It's the right tool for the job and not doing so can result in inconsistencies, and icons not looking the same as others.

The reason this for example had to be hand optimized is that it's not the 3 points of precision approximation that we normally require.

ericfennis (approved):

ericfennis (approved):

karsa-mistmere:
> It would be nice to also fix the gap issue somehow, some ideas: image

Pinging this.

karsa-mistmere:
> > It would be nice to also fix the gap issue somehow, some ideas: image
> 
> Pinging this.

> @[ericfennis](https://github.com/ericfennis) ericfennis requested a review from [karsa-mistmere](https://github.com/karsa-mistmere) [1 minute ago](https://github.com/lucide-icons/lucide/pull/1919#event-13739777202)

@jguddas, @ericfennis: Pinging this again.

jguddas:
No clue what I need to do here.

karsa-mistmere:
> No clue what I need to do here.

The trigger doesn't have the necessary 2px gap (and it's also a bit too off-grid).

PR #1924, opened 2024-02-25
fix(icons): update `goal` icon

jguddas:
- [x] Other: Icon update

### Description
Moved arcs so they have their center at (12|12)

## Before Submitting


karsa-mistmere (approved):

ericfennis:
@jguddas @karsa-mistmere 
Off-topic: Shouldn't we use `perf(icons):` instead of `fix(icons):` for this? 
If we implement the new build flow this will trigger a patch release, I'm not sure if that's really for this kind of improvements.

ericfennis (approved):

PR #1938, opened 2024-03-01
fix(icon): reorder the elements in circle-slash.svg

Ffada94:
- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description
Adding a fill value to the CircleSlash React component caused the line element to be painted over.

This was due to the ordering of the elements within the SVG. Elements are painted in the order they are listed.

This commit swaps the ordering of the elements.

## Before Submitting

### DIFF
--- a/icons/circle-slash.svg
+++ b/icons/circle-slash.svg
@@ -9,6 +9,6 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <line x1="9" x2="15" y1="15" y2="9" />
   <circle cx="12" cy="12" r="10" />
+  <line x1="9" x2="15" y1="15" y2="9" />
 </svg>


jguddas (changes requested):
Oh, that's not something I have thought about.

Currently, the rule is all path elements first and then sort alphabetically.
But maybe there is a better heuristic for sorting?

If we can agree on changing to a better heuristic, we can merge this.

karsa-mistmere:
> Oh, that's not something I have thought about.
> 
> Currently, the rule is all path elements first and then sort alphabetically. But maybe there is a better heuristic for sorting?
> 
> If we can agree on changing to a better heuristic, we can merge this.

Honestly, I wish your optimization script didn't reorder elements at all. 😶

jguddas:
> > Oh, that's not something I have thought about.
> > Currently, the rule is all path elements first and then sort alphabetically. But maybe there is a better heuristic for sorting?
> > If we can agree on changing to a better heuristic, we can merge this.
> 
> Honestly, I wish your optimization script didn't reorder elements at all. 😶 (It usually also makes it hard to see in changes to see what actually changed.)

We need some deterministic way to sort since there is no way to remember where an element originally was placed. So if we don't have sorting it will never stabilize and every time you edit something you will have the problem of things moving around and the diff being hard to read.

What we could do is sort by bounding-box volume, biggest to smallest.

Issue #1957, opened 2024-03-08
idea: @deprecated for certain aliases


jguddas:
### Package

- [ ] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [ ] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [X] all JS packages
- [ ] site

### Description

I would love to see aliases that are for example typos get an @deprecated.

Currently, we can only deprecate a whole icon but not an alias.

### Use cases

#1956 thank you @devjayprakash for bringing this to our attention.

### Checklist

- [X] I have searched the existing issues to make sure this bug has not already been reported.

jguddas:
A solution could be to change the type of the `depracated` field from boolean to array or to add a new field called `depracatedAliases`.

karsa-mistmere:
> A solution could be to change the type of the `depracated` field from boolean to array or to add a new field called `depracatedAliases`.

As I've stated on Discord, this doesn't really suffice, a description should also be provided why this alias was deprecated in the first place and what steps should be taken.

jguddas:
> > A solution could be to change the type of the `deprecated` field from boolean to array or to add a new field called `deprecatedAliases`.
> 
> As I've stated on Discord, this doesn't really suffice, a description should also be provided why this icon/alias was deprecated in the first place and what steps should be taken.

What about this

```json5
{
  name: 'turtle',
  depracated: true,
  depracationReason: "I don't like turtles.",
  aliases: [
    {
      name: 'turdle',
      depracated: true,
      depracationReason: "Spelling mistake.",
    }
  ]
}
```

karsa-mistmere:
Apart from the typos, yeah. We could even keep support for the current version:

```json5
{
  name: 'turtle',
  deprecated: true,
  deprecationReason: "I don't like turtles.",
  aliases: [
    'tortoise',
    {
      name: 'turdle',
      deprecated: true,
      deprecationReason: "Spelling mistake.",
    }
  ]
}
```

PR #1960, opened 2024-03-08
feat(icons): add deprecationReason and alias deprecation feature

karsa-mistmere:
closes https://github.com/lucide-icons/lucide/issues/1957

- [x] New Feature

### Description
When deprecating icons (looking at you, brand icons) or aliases, it is important to clearly mark the reason of deprecation as well as any steps the users should take, this PR adds the possibility to add the reason for deprecation to icons as well as the possibility to mark aliases as deprecated as per #1957 

## Before Submitting 


icon.schema.json
```diff
@@ -21,7 +21,26 @@
...
+              },
+              "deprecated": {
+                "type": "boolean",
+                "default": false
+              },
+              "deprecationReason": {
+                "type": "string"
+              }
+            }
+          }
```
jguddas, line 42: I would say we should enforce that you need to provide a deprecation reason.
And also enforce that you only can set a depracation reason when you set deprecated to true.
```suggestion
          {
            "type": "object",
            "required": [
              "name",
              "deprecate",
              "deprecationReason"
            ],
            "properties": {
              "name": {
                "type": "string"
              },
              "deprecated": {
                "const": true
              },
              "deprecationReason": {
                "type": "string"
              }
            }
          }
```

icon.schema.json
```diff
@@ -51,6 +70,12 @@
     "deprecated": {
       "type": "boolean",
       "default": false
+    },
+    "deprecationReason": {
+      "type": "string"
+    },
```
jguddas, line 76: We can also enforce the relation ship here.

icon.schema.json
karsa-mistmere: > I would say we should enforce that you need to provide a deprecation reason. And also enforce that you only can set a depracation reason when you set deprecated to true.

Good idea, although I think a pre commit hook if more feasible than enforcing it via JSON schema.

icons/between-horizontal-end.json
```diff
@@ -33,6 +33,10 @@
     "account"
   ],
   "aliases": [
-    "between-horizonal-end"
+    {
+      "name": "between-horizonal-end",
+      "deprecated": true,
+      "deprecationReason": "Renamed because of typo. This export will be removed in v1.0"
```
jguddas, line 39: I think adding this here feels wrong, I would prefer adding this in the generator.
I would vote to keep them after v1 in general, too much hassle for the user for not that much upside.
```suggestion
      "deprecationReason": "Renamed because of typo."
```

icon.schema.json
jguddas: We can just run ajv on commit tho, we already do the lint run in the pr workflow.

icons/between-horizontal-end.json
karsa-mistmere: Ideally, after v1.0 deprecated aliases could be kept and automatically removed in the next major release following the one they were added in. If and when we get there I agree that the generator should add this information. For now I think this "dirty hack" should do, as the aforementioned functionality is just too much of a hassle.

icon.schema.json
karsa-mistmere: I've added a `dependentRequired` clause on both instances.

icon.schema.json
karsa-mistmere: Neato, I didn't realise we did, I've added a `dependentRequired` clause to both.
```json
  "dependentRequired": {
    "deprecated": ["deprecationReason"]
  }
```

jguddas:
My vote is to `deprecationReason` an enum instead of free text, so we can ensure that there are no typos and control what depracation types there are.

karsa-mistmere:
> My vote is to `deprecationReason` an enum instead of free text, so we can ensure that there are no typos and control what depracation types there are.

Any ideas as to what these options should be?

jguddas:
> > My vote is to `deprecationReason` an enum instead of free text, so we can ensure that there are no typos and control what depracation types there are.
> 
> Any ideas as to what these options should be?

```jsonc
// depracation-reasons.en.json
{
  "alias.typo": "Renamed because of typo, use {name} instead.",
  "icon.brand": "Brand icons have been deprecated and are due to be removed, we recommend using simpleicons.org instead, refer to https://github.com/lucide-icons/lucide/issues/670"
}
```
https://formatjs.io/docs/intl-messageformat

karsa-mistmere:
I think i18n is a bit of an overkill, since all of our packages are in English, but I've implemented a rudimentary template function and converted the deprecation reason to an enumeration.

(I meant options apart from these two, so that we are somewhat futureproof, but we can think about that later)

icon.schema.json
```diff
@@ -21,7 +21,32 @@
...
+              "name": {
+                "type": "string"
+              },
+              "deprecated": {
+                "const": true
+              },
+              "deprecationReason": {
+                "$ref": "#/$defs/deprecationReasons"
+              },
+              "gracePeriod": {
```
jguddas, line 40: I asked copilot for a better name since I had no clue from just looking at the variable name what it is gonna do.
```suggestion
              "toBeRemovedInVersion": {
```

jguddas: Another option could be a boolean `toBeRemovedInNextMajorVersion` 🤷

karsa-mistmere: Fine by me, renamed it.

icon.schema.json
```diff
@@ -49,8 +74,27 @@
...
+      "description": "The version this icon will be removed in."
+    }
+  },
+  "dependentRequired": {
+    "deprecated": ["deprecationReason", "toBeRemovedInVersion"]
+  },
+  "$defs": {
+    "deprecationReasons": {
+      "type": "string",
+      "enum": ["icon.brand", "alias.typo", "alias.name"]
```
jguddas, line 93:
```suggestion
      "enum": ["icon.guideline-violation.brand", "alias.typo", "alias.guideline-violation.number"]
```

jguddas: And you should not be able to do `alias[0].depracationReason: "icon.bla"`, only "alias.bla" should be allowed, they should be seperated imo.

jguddas (changes requested):

tools/build-icons/utils/deprecationReasonTemplate.mjs
```diff
@@ -0,0 +1,13 @@
+export function deprecationReasonTemplate(
+  deprecationReason,
+  { componentName, iconName, toBeRemovedInVersion },
+) {
```
jguddas, line 4:
```suggestion
) {
const removalNotice = toBeRemovedInVersion ? `This alias will be removed in ${toBeRemovedInVersion}` : undefined
```

jguddas (changes requested):

ericfennis (approved):
Nice lovely work!

icon.schema.json
karsa-mistmere: Done.

karsa-mistmere:
> What about my unanswered pr comments?

They've now been addressed.

tools/build-icons/utils/deprecationReasonTemplate.mjs
```diff
@@ -0,0 +1,17 @@
...
+  const removalNotice = toBeRemovedInVersion
+    ? ` This ${deprecationReason.startsWith('icon') ? 'icon' : 'alias'} will be removed in ${toBeRemovedInVersion}`
+    : '';
+
+  switch (deprecationReason) {
+    case 'alias.typo':
+      return `Renamed because of typo, use {@link ${componentName}} instead.${removalNotice}`;
+    case 'alias.naming':
+      return `This name is no longer compliant with our guidelines, use {@link ${componentName}} instead.${removalNotice}`;
```
jguddas, line 13:
```suggestion
      return `This icon has been renamed because it is no longer compliant with our guidelines, use {@link ${componentName}} instead.${removalNotice}`;
```

karsa-mistmere: What about?

```suggestion
      return `The name of this icon had to be changed because it didn't meet our guidelines anymore, use {@link ${componentName}} instead.${removalNotice}`;
```

jguddas: Maybe instead of `had to be changed` do `was changed`.

jguddas (approved):

PR #1970, opened 2024-03-09
fix(lucide-solid): use jsx for exports too

nix6839:

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

In #1964, the build output was specified as .jsx, but it was not modified in package.json.

## Before Submitting 


karsa-mistmere (approved):
Ugh, right, my bad, I think I accidentally reverted this change before committing the changes.

PR #1974, opened 2024-03-11
fix(icons): optimized `chef-hat` icon

jguddas:
- [x] Other: Icon update

### Description
Optimized paths by moving things around

## Before Submitting


karsa-mistmere:
I think we should add a 1px rounding to the base. 🤔

PR #1987, opened 2024-03-16
fix(icons): added rounding to `star` icon

jguddas:
- [x] Other: Icon update

### Description
Added rounding

## Before Submitting


karsa-mistmere:
could benefit from scaling it up for better visual weight to make up for the increased corner rounding.

ericfennis (commented):
@jguddas Don't forget `star-off` and `star-half`, they should match the `star` icon 😊

ericfennis (changes requested):
`star-half` should be updated as well for consistency.

ericfennis (approved):
Thanks, nice work!

PR #1992, opened 2024-03-16
fix(icons): increased `file-question` icon question mark size to match other icons

jguddas:
- [x] Other: Icon update

### Description
Changed question mark size to be consistent with `circle-questionmark`, `shield-questionmark`, etc.

## Before Submitting


ericfennis (approved):
Yess better

PR #1999, opened 2024-03-18
fix(icons): redesigned `person-standing` icon

jguddas:
- [x] Other: Icon update

### Description
Redesigned icon, fixing gap violations and increasing size.

## Before Submitting


ericfennis:
Hmm, I don't like it. It does not fit the style IMO

jguddas:
> Hmm, I don't like it. It does not fit the style IMO

What do you not like about it?

For me, it feels to professional and sterile.

jguddas:
image

ericfennis:
@jguddas The body is not one shape, which made it feel a bit too abstract.

PR #2002, opened 2024-03-18
fix(icons): added round corners to `signpost` icon

jguddas:
- [x] Other: Icon update

### Description
Added rounding

## Before Submitting


ericfennis:
Hmm not sure if this is worth the change looking at how much code we add and what the difference is. The changes are hard to spot on 100% scale.

karsa-mistmere:
> Hmm not sure if this is worth the change looking at how much code we add and what the difference is. The changes are hard to spot on 100% scale.

Yes, but when scaling them, it's very evident:
image

ericfennis:
@karsa-mistmere True!

ericfennis (approved):

PR #2018, opened 2024-03-25
feat(figma): add data API endpoint for new plugin

karsa-mistmere:

- [x] New Feature

### Description
This PR adds a new data API endpoint the new Figma plugin can use to fetch the icon nodes, categories, aliases et cetera all at once.

## Before Submitting 


jguddas (commented):
You can move the computation of the return value out of the event handler since it is static.

karsa-mistmere:
> You can move the computation of the return value out of the event handler since it is static.

Yeah, even though Vercel will cache it anyway, it sure doesn't hurt.

PR #2030, opened 2024-04-02
Adds nextjs specific dynamic import map

zomars:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

As [specified by the Next.js team](https://github.com/vercel/next.js/discussions/11291#discussioncomment-4303) we should not use ´dynamic´ inside a React component. This uses the recommended approach with a map with dynamic imports.

### Concept 

- [ ] I have provided valid use cases for each icon.
- [ ] I have not added any a brand or logo icon.
- [ ] I have not used any hate symbols.
- [ ] I have not included any religious or political imagery.

### Author, credits & license

### Naming 

- [ ] I've read and followed the [naming conventions](https://lucide.dev/guide/design/icon-design-guide#naming-conventions)
- [ ] I've named icons by what they are rather than their use case.
- [ ] I've provided meta JSON files in `icons/[iconName].json`.

### Design 

- [ ] I've read and followed the [icon design guidelines](https://lucide.dev/guide/design/icon-design-guide)
- [ ] I've made sure that the icons look sharp on low DPI displays.
- [ ] I've made sure that the icons look consistent with the icon set in size, optical volume and density.
- [ ] I've made sure that the icons are visually centered.
- [ ] I've correctly optimized all icons to three points of precision.

## Before Submitting 

### DIFF
--- a/docs/guide/packages/lucide-react.md
+++ b/docs/guide/packages/lucide-react.md
@@ -150,16 +150,15 @@ module.exports = nextConfig
 You can then start using it:
 
 ```tsx
-import dynamic from 'next/dynamic'
 import { LucideProps } from 'lucide-react';
-import dynamicIconImports from 'lucide-react/dynamicIconImports';
+import nextjsDynamicIconImports from 'lucide-react/nextjsDynamicIconImports';
 
 interface IconProps extends LucideProps {
-  name: keyof typeof dynamicIconImports;
+  name: keyof typeof nextjsDynamicIconImports;
 }
 
 const Icon = ({ name, ...props }: IconProps) => {
-  const LucideIcon = dynamic(dynamicIconImports[name])
+  const LucideIcon = nextjsDynamicIconImports[name];
 
   return <LucideIcon {...props} />;
 };
--- a/packages/lucide-react/README.md
+++ b/packages/lucide-react/README.md
@@ -124,16 +124,15 @@ module.exports = nextConfig
 You can then start using it:
 
 ```tsx
-import dynamic from 'next/dynamic'
 import { LucideProps } from 'lucide-react';
-import dynamicIconImports from 'lucide-react/dynamicIconImports';
+import nextjsDynamicIconImports from 'lucide-react/nextjsDynamicIconImports';
 
 interface IconProps extends LucideProps {
-  name: keyof typeof dynamicIconImports;
+  name: keyof typeof nextjsDynamicIconImports;
 }
 
 const Icon = ({ name, ...props }: IconProps) => {
-  const LucideIcon = dynamic(dynamicIconImports[name])
+  const LucideIcon = nextjsDynamicIconImports[name];
 
   return <LucideIcon {...props} />;
 };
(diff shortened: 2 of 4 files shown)


zomars:
On further inspection, would it make sense to publish this map as a separate `lucide-next` package?

ericfennis:
Hmm, yess, we shouldn't have `next` as a dependency in `lucide-react`. 
We should move this to a separate package, and maybe the current dynamicImport file should be in there as well.
We could maybe generate all "dynamicImports" when installing the package, It would be nice if we could remove the `dynamicImports` from the `lucide-react` itself,  currently, it is often used wrong. 
We should scope this a bit more for specific use-cases.

zomars:
> Hmm, yess, we shouldn't have `next` as a dependency in `lucide-react`. We should move this to a separate package, and maybe the current dynamicImport file should be in there as well. We could maybe generate all "dynamicImports" when installing the package, It would be nice if we could remove the `dynamicImports` from the `lucide-react` itself, currently, it is often used wrong. We should scope this a bit more for specific use-cases.

Agreed. Please let me know how I can help in here. RN we have it [hard coded in our open source repo](https://github.com/calcom/cal.com/blob/main/packages/ui/components/icon/dynamicIconImports.tsx) but it make sense as you said to have it generated on install.

vsnthdev:
Not stale, we need this PR because the issue hasn't been solved yet.

michaelschufi:
Hi @zomars 
Thank you for this PR and the link to the hardcoded file! 
This is working well so far. But I wonder how you are handling this at the moment. I've seen that in your main branch you have stopped using this dynamic import map. If I may ask - What is your strategy now and what's been your experience regarding performance with lucide-react and Next.js?

dsbrianwebster:
@michaelschufi been banging my head against the wall wondering why my development server was so slow because of this (https://github.com/lucide-icons/lucide/issues/1576).

Sharing my current workaround for now below...

```tsx
import { type LucideProps, icons } from 'lucide-react';

type IconComponentName = keyof typeof icons;

interface IconProps extends LucideProps {
  name: string; // because this is coming from the CMS
}

// 👮‍♀️ guard
function isValidIconComponent(componentName: string): componentName is IconComponentName {
  return componentName in icons;
}

export function DynamicIcon({ name, ...props }: IconProps) {
  // we need to convert kebab-case to PascalCase because we formerly relied on
  // lucide-react/dynamicIconImports and the icon names are what are stored in the CMS.
  const kebabToPascal = (str: string) =>
    str
      .split('-')
      .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
      .join('');

  const componentName = kebabToPascal(name);

  // ensure what is in the CMS is a valid icon component
  if (!isValidIconComponent(componentName)) {
    return null;
  }

  // lucide-react/dynamicIconImports makes makes NextJS development server very slow
  // https://github.com/lucide-icons/lucide/issues/1576
  const Icon = icons[componentName];

  return <Icon {...props} />;
}

```

PR #2033, opened 2024-04-03
fix(icons): Add axis tag to scale-3d

gurtt:

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

Add `axis` tag to `scale-3d`, bringing it in line with similar `axis-3d` and `move-3d`.

## Before Submitting 

### DIFF
--- a/icons/scale-3d.json
+++ b/icons/scale-3d.json
@@ -3,8 +3,7 @@
   "contributors": [
     "lscheibel",
     "ericfennis",
-    "jguddas",
-    "gurtt"
+    "jguddas"
   ],
   "tags": [
     "gizmo",
--- a/icons/scale-3d.json
+++ b/icons/scale-3d.json
@@ -3,12 +3,14 @@
   "contributors": [
     "lscheibel",
     "ericfennis",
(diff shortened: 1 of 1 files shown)


icons/scale-3d.json
```diff
@@ -3,12 +3,14 @@
   "contributors": [
     "lscheibel",
     "ericfennis",
-    "jguddas"
+    "jguddas",
+    "gurtt"
```
karsa-mistmere, line 7: Normally, we only give contribution for substantial design changes or optimizations, not metadata changes, but @ericfennis and @jguddas may veto me on this one (should they both agree otherwise).

gurtt: Thanks for clarifying 👍 I couldn't find any good documentation for these fields

ericfennis: Yess only for design changes

jguddas: I usually add my name when I manually change something inside the `icon.svg`.

PR #2057, opened 2024-04-12
feat(icon): add bookmarks, add tall bookmarks, add new series of badges (plus, minus, check, slash, magnifying glass, x, cog) to bookmark regular and bookmark tall 

kdsaft:

- [x] New Icon
- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

Created a bookmarks icon to represent a set of bookmarks. Created an alternative bookmark that is taller to work better with the badges. For individual bookmarks with badges, I increased the prominence/ visibility of the badges for fast scanning. I added some additional badges. Created all badges for the current-sized bookmark and the new taller one. Also added a bookmark with a slash.

### Icon use case 

**Bookmarks** — PURPOSE: When you have a collection of bookmarks. USE CASES: 1. If you have a series of bookmarks in your app/web page and you need to create navigation to get to that page (labeled 'Bookmarks' or 'Library'). 2. In a settings menu to adjust your bookmark attributes. 3. In a share menu to share a collection of bookmarks.

**Bookmarks Tall Family (bookmarks-tall)** – PURPOSE: To update the visual style of the current bookmark to look more similar to a physical bookmark and to give more vertical space to the badges. USE CASES: 1. The current squarer bookmark works well visually when used with other icons centered shaped, this style helps the icon stand more apart. 2. When the user is scanning to find the bookmark button, a shape that is closer to a physical bookmark will be seen faster.

**Bookmark badges** — PURPOSE: Reposition the bases so that they are more prominent. I added some additional badges to have a more complete family. USE CASES: 1. When scanning to add or remove a bookmark it is easer to see when the outline changes rather then just the center. 2. SEARCH: when the user wants to search within their bookmarks 3. COG: when the user wants to change the settings of the bookmark.

**Bookmark with slash** — PURPOSE: To unbookmark something and to match the style of other icons in the set. USE CASE: 1.  When the user is looking at article that is bookmarked, they would click this icon to unbookmark it. In this use case, the minus or X icon feels more appropriate for a list of bookmark; here they feel like you are delete the article. 2. When using with other icons that have an OFF state.

### Concept 

### Author, credits & license

- [x] I've based them on the following Lucide icons: bookmark, slash, plus, minus, check, x, cog, search

### Naming 

### Design 

## Before Submitting 


jguddas:
Try https://lucide-studio.vercel.app to optimize the icon.

kdsaft:
@jguddas Thanks for the link to optimize the icons

karsa-mistmere:
Hey @kdsaft,

Sorry for taking so long to reply. To be honest, this PR is highly unlikely to ever go through as is; adding alternates that look so extremely similar just isn't something we do – on the other hand I'm very much open to the idea of redesigning our current bookmark icons, as you're right: they could definitely do with more optical volume.

Your work is definitely on the right track, however, as of late we've been "Lucidising" a lot of our core icons, so my proposal would be to follow suit here and add some additional rounding to corners where applicable, i.e.:

image

<a title="Open lucide studio" target="_blank" href="https://lucide-studio.vercel.app/edit?value=%3Cpath+d=%22M18.21+21.907A.5.5+0+0+0+19+21.5V4a2+2+0+0+0-2-2H7a2+2+0+0+0-2+2v17.5a.5.5+0+0+0+.79.407l5.048-3.535a2+2+0+0+1+2.324+0z%22+/%3E&name=image">icons<br/>Open lucide studio</a>

Should @ericfennis & @jguddas be okay with this, we should simply update existing `bookmark-*` icons with this new base shape and there won't be any need for the `-tall` variants.

ericfennis:
@karsa-mistmere I agree taller variants are similar to the current bookmark icons. 
But I think I like the current bookmark icon shape more, it looks a bit more balanced.

karsa-mistmere:
> @karsa-mistmere I agree taller variants are similar to the current bookmark icons. But I think I like the current bookmark icon shape more, it looks a bit more balanced.

Tbh to me they feel somewhat less balanced, a bit too top heavy (and definitely a bit too small), elongating them moves the centre of gravity a bit more to the actual centre point:

image

jguddas:
+1 for the original size

kdsaft:
Thanks for the feedback. My thinking was that bookmarks tend to be long and thin—and I thought it was important to echo that ratio in the design. But I understand. I will close off the PR.

PR #2061, opened 2024-04-12
fix(icons): redesigned `fast-forward` icon

jguddas:
- [x] Other: Icon update

### Description
Redesign.

## Before Submitting


jguddas:
Also needs to be changed for `rewind`.

karsa-mistmere:
> Also needs to be changed for `rewind`.

That, and we should also take into consideration the discussions in https://github.com/lucide-icons/lucide/pull/1043#issuecomment-1532579985

PR #2067, opened 2024-04-13
fix(icons): arcified `key` icon

jguddas:
- [x] Other: Icon update

### Description
Arcified icon.

## Before Submitting


icons/key.svg
```diff
@@ -9,7 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <circle cx="7.5" cy="15.5" r="5.5" />
+  <path d="m15.5 7.5 2.293 2.293a1 1 0 0 0 1.414 0l2.086-2.086a1 1 0 0 0 0-1.414L19 4" />
```
ericfennis, line 12: The path should be optimized a bit. This doesn't affect the look
```suggestion
  <path d="m15.5 7.5 2.3 2.3a1 1 0 0 0 1.4 0l2.1-2.1a1 1 0 0 0 0-1.4L19 4"/>
```

ruru-m07: Agree +1

PR #2083, opened 2024-04-18
Potential compatibility with shadow DOM

unit-404:
When using custom elements, needs to create icons again when inside of shadow DOM. Shadow DOM invisible from document selectors, so can't be replaced directly. **There is no new any icon.**

- [ ] New Icon
- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [x] Other:

### Description

Adds `rootElement` field for `createIcons` (default, is `document`).

## Before Submitting 


ericfennis:
@unit-404 Cool idea!
Did you already get it working?

Maybe we can also test this with libraries like Lit.

unit-404:
About get working, I did it in separate code (re-wrap). I.e. re-created that function.

PR #2122, opened 2024-05-03
feat(docs): added accessibility guide

karsa-mistmere:
closes #2121 

- [x] Documentation update

## Before Submitting 


docs/images/a11y/alttext-buttons.svg
jguddas, line 1: Would be nice to have a better text here other than just `Button`.

docs/guide/advanced/accessibility.md
```diff
@@ -1,4 +1,114 @@
...
+screen readers, leading to nonsensical text.
+
+![](../../images/a11y/alttext-buttons.svg?raw=true)
+
+### On icon buttons
+
+Icon buttons are buttons that do not contain any visible text apart from the icon itself (think of
+the close button of a dialog for example).
+
+![](../../images/a11y/alttext-iconbuttons.svg?raw=true)
```
jguddas, line 114: Why?

jguddas (changes requested):
- The content of the images is not accessible, the alt text is different from the content of the image. I would generally prefer if we moved the text out of the images and made them just normal text.
- Would be nice to have a section on flashing images and use reduce motion as well
- Also `aria-label` is not perfect, using something like RadixUIs `AccessibleIcon` is something that we should point people to use instead.
- A section on who this for and why you should do this would also be nice.

jguddas:
We can also point people to resources that they should reference, be it just the WCAG or even component libraries or guides like https://inclusive-components.design/toggle-button/

karsa-mistmere:
> * The content of the images is not accessible, the alt text is different from the content of the image. I would generally prefer if we moved the text out of the images and made them just normal text.

For the most part this is entirely intentional, these are illustrations, I've added accessible text where there was information otherwise missing, but most of these illustrations contain redundant information that is not useful to be read out as is. There's probably room for improvement, on that I can agree.

> Would be nice to have a section on flashing images and use reduce motion as well

We don't really have ootb support for such animations, so I'm not entirely sure if it's relevant within the context of Lucide.

> Also aria-label is not perfect, using something like RadixUIs AccessibleIcon is something that we should point people to use instead.

Well, the current guide does recommend not using it in places where it's problematic, but you're right, there probably space for some more nuance about this.

> A section on who this for and why you should do this would also be nice.

Is it not for everyone using Lucide? 🧌

jguddas:
> > A section on who this for and why you should do this would also be nice.
> 
> Is it not for everyone using Lucide? 🧌

Example: Your should put enough contrast so people with visual impairments, be it physical or just them being outside in the glaring sun can still get the amazing benefit of looking at those cute icons.

jguddas:
> > Would be nice to have a section on flashing images and use reduce motion as well
> 
> We don't really have ootb support for such animations, so I'm not entirely sure if it's relevant within the context of Lucide.

Tailwind based example: You can do `<DogIcon className="animate-ping" />` but what you should consider using `<DogIcon className="motion-safe:animate-ping" />` instead.

karsa-mistmere:
> > > Would be nice to have a section on flashing images and use reduce motion as well
> > 
> > 
> > We don't really have ootb support for such animations, so I'm not entirely sure if it's relevant within the context of Lucide.
> 
> Tailwind based example: You can do `<DogIcon className="animate-ping" />` but what you should consider using `<DogIcon className="motion-safe:animate-ping" />` instead.

Yeah, but this is entirely dependant on Tailwind, I think it should be their responsibility to educate people about using their animations responsibly. :)

(To be frank, this should be their default, it shouldn't be something to opt into.)

docs/guide/advanced/accessibility.md
```diff
@@ -117,6 +116,36 @@ contained icon.
 
 ![](../../images/a11y/alttext-iconbuttons.svg?raw=true)
 
+## A note on `aria-label`
+
+Although you could provide accessible labels to your elements via the `aria-label` attribute, we
+generally suggest using your chosen CSS framework's "visually hidden" utility instead whenever
+possible, since some assistive technology fails to announce `aria-label` attributes in case the
+element is not of an appropriate semantic role.
```
jguddas, line 124: The main reason I don't like `aria-label` is that it gets exposed by all major browsers as AXDescription while normal content is exposed as AXTitle, and according to the spec, AXDescription should be read out in the system language, meaning your `lang=…` attribute gets ignored and your screen reader might sound funky.

karsa-mistmere: Neato, I've updated this description and removed all traces of `aria-label` from the illustrations.

docs/images/a11y/alttext-buttons.svg
jguddas, line 1: We could even set `aria-hidden="true"` on decorative icons like that.

karsa-mistmere: I'm a bit confused, does this mean that simple SVGs with no accessible text are currently being read out? 🤔

jguddas: The read-out text for voiceover for example is "You are currently on an image. To begin interacting with the contents of this image, press Control-Option-Shift-Down Arrow.", that's so super not helpful for decorative images.


image

karsa-mistmere: Oh jeez louise, that's awful. 🤮

+1 for adding aria-hidden then. Best course of action would actually be to automatically add that no matter what.

jguddas: Hiding the icon by default sounds super dangerous.

As I see it there are 2 okayish options:

1. Add a default label that is a descriptive representation of the icon. #2121
2. Make setting a `label` / `aria-hidden` attribute mandatory and well documented for when to use what.

karsa-mistmere: > 1. Add a default label that is a descriptive representation of the icon. https://github.com/lucide-icons/lucide/discussions/2121

Terrible-terrible idea, as I've already explained in #2121:
- Accessible labels on functional icons should describe the use case, not the graphical representation (and even that isn't necessarily unambiguous, especially considering abstract icons).
- Decorative icons on the other hand should have no accessible label whatsoever.

> 2. Make setting a label / aria-hidden attribute mandatory and well documented for when to use what.

Supporting any kind of accessible label would lead to breaking changes, since we cannot use either `<title>` nor `aria-label="..."` as already discussed in previous threads. That means we would have to add at least one extra element to the DOM and inject CSS where there previously were none (or use inline CSS for our `visually-hidden` util).

> Hiding the icon by default sounds super dangerous.

I don't really see why, it is practically _never_ advisable to set an accessible label on an icon itself, but instead on other interactive elements, or using some kind of `visually-hidden`/`sr-only` util class, but that by nature will be on a different element as well.

docs/guide/advanced/accessibility.md
```diff
@@ -1,4 +1,158 @@
...
+As previously stated, you should provide your accessible label on the icon button itself, not the
+contained icon.
+
+![](../../images/a11y/alttext-iconbuttons.svg?raw=true)
+
+## A note on `aria-label`
+
+Although you could provide accessible labels to your elements via the `aria-label` attribute, we
+generally recommend against this for and instead suggest that
+you [use your chosen CSS framework's "visually hidden" utility whenever possible](https://gomakethings.com/revisting-aria-label-versus-a-visually-hidden-class/).
```
ericfennis, line 123: Can you explain why we should be "against" this? It would be nice if we explained this more in the text, about why using "visually hidden" is better than using `aria-label`, we also could provide some sources for blog posts or documentation to prove our "recommendation".

Also maybe this reads a bit nicer
``` suggestion
generally recommend visually hiding textual descriptions. For example by [using utility CSS framework's "visually hidden"](https://gomakethings.com/revisting-aria-label-versus-a-visually-hidden-class/).
```

ericfennis (commented):
Sorry for the late review.
This is such an awesome guide, thank you @karsa-mistmere super useful!
We can improve this even more if we add some small code examples for some things to make it visible in the code. 

I've provided some suggestions and one comment

docs/guide/advanced/accessibility.md
karsa-mistmere: That means something different though. Also, the link itself is the resource that explains why `.visually-hidden` > `[aria-label]`, I've rephrased the link text so that it's clearer.

```suggestion
generally recommend against this and instead suggest that you use your chosen CSS framework's "visually hidden" utility whenever possible. You can [read more about why `aria-label` might not be the best solution](https://gomakethings.com/revisting-aria-label-versus-a-visually-hidden-class/).
```

docs/guide/advanced/accessibility.md
```diff
@@ -1,4 +1,158 @@
...
+In any such case, it is preferred that the accessible name be provided for these interactive
+elements (badges, buttons, nav items etc.) only, _not_ the icons themselves.
+
+### On buttons
+
+Do not provide an accessible label to icons when used on a button, as this label will be read out by
+screen readers, leading to nonsensical text.
+
+![](../../images/a11y/alttext-buttons.svg?raw=true)
+
```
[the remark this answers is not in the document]
karsa-mistmere: @ericfennis: These could be useful, but they break up the guide quite a bit, I've hidden them within details elements.

docs/guide/advanced/accessibility.md
ericfennis: I'm still not happy with the phraise "we generally recommend against this". That suggest that we are against the use of `aria-label`. And from what I found there nothing wrong with it.
I think you should approach it more positively. "If you are using a CSS framework like Tailwind, we recommend using visually hidden utilities whenever possible"

docs/guide/advanced/accessibility.md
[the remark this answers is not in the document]
ericfennis: Nice looks great!

docs/guide/advanced/accessibility.md
jguddas: The options from best to worst:

1. visible label
2. visually hidden label
3. aria-label

karsa-mistmere: > And from what I found there nothing wrong with it.

Well, there kind of is, that's what the article linked is about: https://gomakethings.com/revisting-aria-label-versus-a-visually-hidden-class/#the-aria-label-attribute-is-not-always-translated

But sure, we can rephrase it to be more lenient.

ericfennis: They use the word "avoid" instead of "against", which is I think more right fit. Since using `aria-label` is not wrong. I  think we also should link this article to this paragraph, it gives a bit more context to this topic.

docs/guide/advanced/accessibility.md
```diff
@@ -1,4 +1,204 @@
...
+  <span class="visually-hidden">Go to home</span>
+</button>
+```
+
+:::
+
+## A note on `aria-label`
+
+Although you could provide accessible labels to your elements via the `aria-label` attribute, we
+generally recommend against this and instead suggest that you use your chosen CSS framework's "
```
karsa-mistmere, line 167:
```suggestion
generally recommend avoiding this and instead suggest that you use your chosen CSS framework's "
```

docs/guide/advanced/accessibility.md
karsa-mistmere: No, in certain scenarios it might actually be wrong, since `aria-label` will generally be ignored on non-interactive elements. I don't believe linking to a certain paragraph is necessary, since the entirety of article is about why `aria-label` should be avoided if possible, there isn't a single reason, there are multiple.

Since this thread is on an outdated commit, I've made a suggestion here: https://github.com/lucide-icons/lucide/pull/2122/files#r1636386974

ericfennis: Ah ok true, tnx for suggestion think it reads a bit nicer as well.

ericfennis (approved):

docs/guide/advanced/accessibility.md
```diff
@@ -1,4 +1,204 @@
...
+:::
+
+## A note on `aria-label`
+
+Although you could provide accessible labels to your elements via the `aria-label` attribute, we
+generally recommend avoiding this and instead suggest that you use your chosen CSS framework's "
+visually hidden" utility whenever possible. You can
+[read more about why `aria-label` might not be the best solution](https://gomakethings.com/revisting-aria-label-versus-a-visually-hidden-class/).
+
+E.g. in case of Bootstrap:
```
jguddas, line 171:
```suggestion
### Example - Bootstrap
```

docs/guide/advanced/accessibility.md
```diff
@@ -1,4 +1,204 @@
...
+
+```html
+
+<div>
+  <i data-lucide="phone" aria-hidden="true"></i>
+  <span class="visually-hidden">Phone number</span>
+</div>
+```
+
+Or with tailwindcss:
```
jguddas, line 181:
```suggestion
### Example - Tailwind CSS
```

docs/guide/advanced/accessibility.md
```diff
@@ -1,4 +1,204 @@
...
+
+:::
+
+## A note on `aria-label`
+
+Although you could provide accessible labels to your elements via the `aria-label` attribute, we
+generally recommend avoiding this and instead suggest that you use your chosen CSS framework's "
+visually hidden" utility whenever possible. You can
+[read more about why `aria-label` might not be the best solution](https://gomakethings.com/revisting-aria-label-versus-a-visually-hidden-class/).
+
```
jguddas, line 170:
```suggestion

### Example - Radix UI

https://www.radix-ui.com/primitives/docs/utilities/accessible-icon

***tsx
import { ArrowRightIcon } from 'lucide-react';
import { AccessibleIcon } from '@radix-ui/react-accessible-icon';

<AccessibleIcon label="Next item">
  <ArrowRightIcon />
</AccessibleIcon>
***
```

karsa-mistmere: https://www.radix-ui.com/primitives/docs/utilities/accessible-icon

Ought to be a link with descriptive text.
````suggestion
### Example - Radix UI

Use [Radix UI's built-in accessible icon utility component](https://www.radix-ui.com/primitives/docs/utilities/accessible-icon).

```tsx
import { ArrowRightIcon } from 'lucide-react';
import { AccessibleIcon } from '@radix-ui/react-accessible-icon';

<AccessibleIcon label="Next item">
  <ArrowRightIcon />
</AccessibleIcon>
```
````

PR #2126, opened 2024-05-03
feat: narrowed type for categories in icon.schema.json

jguddas:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other:

### Description

Before the categories field in `icons/*.json` could be any string, now it can only be a valid category name.

## Before Submitting 


karsa-mistmere (approved):

icon.schema.json
```diff
@@ -28,7 +28,55 @@
...
+          "emoji",
+          "files",
+          "food-beverage",
+          "furniture",
+          "gaming",
+          "home",
+          "layout",
+          "mail",
+          "maps",
+          "maths",
```
ericfennis, line 56: I think we should rename this to `math`

jguddas: Not in this PR.

ericfennis: Fine, I will make a separate one!

ericfennis (approved):

PR #2139, opened 2024-05-07
fix(icons): changed `volume-x` icon

jguddas:
- [x] Other: Icon update

### Description
Arcified icon.

## Before Submitting


ericfennis (commented):
Can you also update the other `volume` icons in this PR?

ericfennis (changes requested):
All `volume` icons should be updated. Otherwise icons are not consistent

ericfennis:
Fixed by #2378

PR #2157, opened 2024-05-18
feat(icons): add wifi strength icons

VirtCode:

- [x] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

Adds a `wifi-zero`, `wifi-low` and `wifi-high` icon, which are all a variant of the `wifi` icon in lucide with a few bars removed. They are similar to the wifi strength indicators used in almost all operating systems, e.g. Android, iOS, Windows, etc.

### Icon use case 

These icons can be used to indicate specifically wifi signal strength in an operating systems UI / desktop environment. The `signal-*` icons often cannot be used for that, as they are usually associated with cellular signal and not wifi.

### Alternative icon designs 

There are no alternative designs as the icons should be "parts" of the original and complete `wifi` icon.

### Concept 

### Author, credits & license

- [x] I've based them on the following Lucide icons: `wifi`, inspired by the `signal-*` icons

### Naming 

I tried to keep the naming similar to the `signal-*` icons, so I used the `zero`, `low` and `high` suffixes.

### Design 

I just copied the `wifi` icon and removed the paths that were not needed for the respective icons. Visual center is not relevant in this case.

## Before Submitting 

### DIFF
--- a/icons/wifi-high.svg
+++ b/icons/wifi-high.svg
@@ -1,0 +1,15 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <path d="M12 20h.01" />
+  <path d="M5 12.859a10 10 0 0 1 14 0" />
+  <path d="M8.5 16.429a5 5 0 0 1 7 0" />
+</svg>
--- a/icons/wifi-high.json
+++ b/icons/wifi-high.json
@@ -1,0 +1,18 @@
+{
+  "$schema": "../icon.schema.json",
+  "contributors": [
+    "colebemis",
+    "ericfennis",
+    "jguddas",
+    "VirtCode"
+  ],
+  "tags": [
+    "connection",
+    "signal",
+    "wireless"
+  ],
+  "categories": [
+    "connectivity",
+    "devices"
+  ]
+}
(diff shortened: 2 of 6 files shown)


ericfennis (approved):
LGTM!

icons/wifi-high.svg
karsa-mistmere, line 1: Should this not be `wifi-medium` with `wifi` geting the `wifi-strong` tag?

karsa-mistmere (changes requested):

icons/wifi-high.svg
VirtCode: I was trying to keep the naming similar to the signal icons, where there is [`signal`](https://lucide.dev/icons/signal) and [`signal-high`](https://lucide.dev/icons/signal-high). But then they also have one variant more (5, not 4).

But I don't mind changing `wifi` to `wifi-high` etc. let me know whether I should do that.

karsa-mistmere: No, let's definitely keep `wifi` as `wifi`, high should only ever be a tag. But good points, let's first investigate how signal icons are named and tagged.

VirtCode: If we want to keep the naming consistent between these two, I would also use the `-zero` tag for the lowest icon (the dot), as is with `signal-zero`. As discussed, we also keep `wifi` as `wifi`. 

So we have three options on how we want to name the two intermediate icons (we have three tags left over to work with from the `signal-*` icons):
- `wifi-low` and `wifi-high` (as it currently is)
- `wifi-low` and `wifi-medium`
- `wifi-medium` and `wifi-high`

Let me know which option you think is best in this case.

VirtCode: What do you think, should we leave it as is or rename to something like described above?

karsa-mistmere: Let's keep it as is! I've just merged it, since @ericfennis has already approved it.

PR #2158, opened 2024-05-18
fix(react): added aria-hidden fallback for decorative icons

jguddas:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other: Fixing an accessibility nuisance

### Description

When you use an icon without a label `<Square/>` it will be read out by a lot of screen readers as `Image`.
After asking some people including screen reader users it looks like this is just a nuisance and having icons that don't have a label be hidden by default is the way to go.

## Before Submitting 

### DIFF
--- a/packages/lucide-react/src/Icon.ts
+++ b/packages/lucide-react/src/Icon.ts
@@ -1,7 +1,7 @@
 import { createElement, forwardRef } from 'react';
 import defaultAttributes from './defaultAttributes';
 import { IconNode, LucideProps } from './types';
-import { mergeClasses } from '@lucide/shared';
+import { mergeClasses, hasA11yProp } from '@lucide/shared';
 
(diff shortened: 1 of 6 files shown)


karsa-mistmere (changes requested):
Nice job on the implementation, however, I do have three concerns to raise:

1. There are some linting issues that need to be fixed.

2. While, as we've privately discussed, I'm in support of adding this functionality exactly as implemented, I think we should add it to _every_ package where possible, not just `lucide-react` (most of them should be pretty straightforward with how you've implemented this).

3. This is definitely something that we should explicitly document in our guide. Possibly the Accessibility guide in #2122, but probably also relevant package documentations.

ericfennis (approved):
Awesome work @jguddas!
We should also add this for the other packages.

PR #2183, opened 2024-06-05
fix(icons): [circle-slash] Fix missing line when using `fill` as a background

jrozbicki:
Having <line> after <circle> allows to set `fill` as a background and line would still be visible

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description
Currently, when you set `fill` property on svg, `<line>` won't be visible, as `<circle>` will cover it. Switching order allows to add background to the icon without losing `<line>`

Before:
image

After:
image

## Before Submitting 

### DIFF
--- a/icons/circle-slash.svg
+++ b/icons/circle-slash.svg
@@ -9,6 +9,6 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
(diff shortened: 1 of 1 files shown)


ericfennis (approved):
Nice work!

PR #2215, opened 2024-06-11
chore(ci): consolidate pull request workflows

jguddas:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other:

### Description

Instead of having a bunch of workflows, we can use a pnpm filter to only run the scripts of packages that are affected.

## Before Submitting 


ericfennis (commented):
Hmm, I don't know if this is completely bulletproof.
For example, we should run all the builds if something is changed in the build scripts or `package-lock` because there is a chance it could break it. Currently, this is located in paths that are defined in the top of the script files.

Also with separate workflows, it is more clear what is wrong in which package. Especially if you are a new contributor.

jguddas:
> Hmm, I don't know if this is completely bulletproof. For example, we should run all the builds if something is changed in the build scripts or `package-lock` because there is a chance it could break it. Currently, this is located in paths that are defined in the top of the script files.
> 
> Also with separate workflows, it is more clear what is wrong in which package. Especially if you are a new contributor.

This will run the build for lucide-react for example when you change @lucide/build-icons.

https://pnpm.io/filtering#--filter-package_name-3

PR #2223, opened 2024-06-16
feat(ci): check uniqueness of aliases

jguddas:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other:

### Description

Check that there are no overlaps between icon names and aliases, aliases and names need to be unique.

## Before Submitting 


ericfennis (approved):
I hope this will work, but will we see when we test it!

PR #2236, opened 2024-06-20
fix(icons): changed `message-square-plus` icon plus symbol location

jguddas:
- [x] Other: Icon update

### Description
Changed location of the plus symbol.

## Before Submitting


karsa-mistmere:
I'm not super convinced about what I'm about to say, but I think this maybe changes the original intent too much. 🤔

ericfennis:
IMO, the original is better

PR #2294, opened 2024-07-07
fix(icons): fixed corner in `map-pinned` icon

jguddas:
- [x] Other: Icon update

### Description
Fixed corner.

## Before Submitting


jguddas:
Still not perfect, but better than before

ericfennis (approved):

karsa-mistmere:
I think #2301 supersedes this by also arcifying the pin shape itself.

PR #2298, opened 2024-07-08
fix(site): Fix svelte code examples

bcldvd:

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

Fixes Svelte code examples that were showing `$PascalCase` instead of the Svelte component's tag

image

Issue stemmed from the fact that the `$` and `PascalCase` were not contiguous, and separated by HTML:
image

> Proposed fix is to add a a non-capturing group that matches zero or more occurrences of any sequence starting with <, followed by any characters that are not >, and ending with >; basically skipping over any HTML tags that might occur right after the `$` and before the `Pascal`.

### Behavior after fix
image

## Before Submitting 

### DIFF
--- a/docs/icons/[name].md
+++ b/docs/icons/[name].md
@@ -35,7 +35,7 @@ const codeExample = computed(() => data.codeExamples?.map(
       const camelCaseName = camelCase(params.value.name)
 
       return codeExample.code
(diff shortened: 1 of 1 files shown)


ericfennis (approved):
Thanks @bcldvd !

PR #2326, opened 2024-07-25
fix(icons): Fix rounding to hand-platter icon

karsa-mistmere:
- [x] Bug fix

## Before Submitting 


jguddas:
Feels better but still not very lucidey.

ericfennis (approved):

jguddas (approved):

Issue #2216, opened 2024-06-12
Install error in react 19 (next 15)


chakeawx:
### Package

- [ ] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [X] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

0.394.0

### Can you reproduce this in the latest version?

- [X] Yes
- [ ] No

### Browser

- [ ] Chrome/Chromium
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [X] Other/not relevant

### Operating system

- [X] Windows
- [ ] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [ ] Other/not relevant

### Description

```bash
$ npm i lucide-react
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: my-app@0.1.0
npm ERR! Found: react@19.0.0-rc.0
npm ERR! node_modules/react
npm ERR!   react@"19.0.0-rc.0" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.5.1 || ^17.0.0 || ^18.0.0" from lucide-react@0.394.0
npm ERR! node_modules/lucide-react
npm ERR!   lucide-react@"^0.394.0" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR!
npm ERR!
npm ERR! For a full report see:
npm ERR! C:\Users\chakeawx\AppData\Local\npm-cache\_logs\2024-06-12T09_23_37_917Z-eresolve-report.txt

npm ERR! A complete log of this run can be found in: C:\Users\chakeawx\AppData\Local\npm-cache\_logs\2024-06-12T09_23_37_917Z-debug-0.log
```

### Steps to reproduce

1. init nextjs (version 15)
2. install lucide-react

### Checklist

[...]

PR #2336, opened 2024-07-29
fix(lucide-react): support react 19 recommended prerelease

pantoninho:
#2216

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

I don't think #2216 is fixed because react `19` is installed with a prerelease name. It seems that `rc` is the recommended prerelease to install because:
* [official react 19 upgrade guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#installing) guides the user to install the `rc` prerelease in order to use `react 19`
* `NextJS 15` installs `react@19.0.0-rc`

Also, searching for `react@^19.0.0` on [npm semver calculator](https://semver.npmjs.com) shows no matching versions, which means that adding `^19.0.0` to the peer dependency list has no effect has of now.

image

## Before Submitting 


packages/lucide-react/package.json
```diff
@@ -65,6 +65,6 @@
     "vitest": "^1.1.1"
   },
   "peerDependencies": {
-    "react": "^16.5.1 || ^17.0.0 || ^18.0.0 || ^19.0.0"
+    "react": "^16.5.1 || ^17.0.0 || ^18.0.0 || ^19.0.0-rc"
```
VIKTORVAV99, line 68: Should it not support both the pre-release and normal release?

Otherwise people with be forced to update the package again even if they don't really need to.

ericfennis: @VIKTORVAV99 We made a mistake the specific version `19.0.0` does not exists yet. Only `react@19.0.0-rc`. So the change of @pantoninho is right

ericfennis (approved):

packages/lucide-react/package.json
pantoninho: Sorry for the late reply.
I did not include `19.0.0` because I can't confidently say that `lucide-react` will be compatible with `react@19.0.0`. I thought it would be worse if it was indeed incompatible and broke installations that upgraded to `react@19.0.0`.

That was my line of thought, but I agree with you that the API is most likely stable. If you want I can open a PR that includes both versions.

PR #2346, opened 2024-07-31
fix(icons): lucide-ize send icons

karsa-mistmere:

- [x] Bug fix

### Description
Adds rounding to send icons.

## Before Submitting 


jguddas:
This feels consistent, I feel like we should arcify everything, including things that are arrowy.

jguddas (approved):

Issue #560, opened 2022-04-11
Gender icons


karsa-mistmere:
## Icon Request

* Icon name: gender icons
* Use case: representing genders
* _Screenshots_ of similar icons: 
![image](https://user-images.githubusercontent.com/17746067/162781091-f3d6124b-7c46-40d1-ab28-035cdfc15333.png)

This is a continuation of https://github.com/feathericons/feather/issues/27 as per #119 

Here are a few rough drafts of all 14 gender icons currently implemented by Fontawesome.
(link removed)

All of these more or less comply with our design guidelines, however are probably nowhere near final.
One of the things up for discussion is probably whether we should aim to use the same base element sizes for all of them, but I'm not quite sure that's feasible.

clementoriol:
Hi ! I'm interested in those, but can't seem to find them in the library.

Maybe this issue should be reopened (currently closed as completed)

btnalexandre:
Hello, these icons would indeed be interesting to integrate?

btnalexandre:
poke @danielbayley

tiagobbraga:
up

jguddas:
https://en.wikipedia.org/wiki/Planet_symbols

[...]

PR #2363, opened 2024-08-08
Feat/gender icons

Trukl:

- [X] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

Added new basic gender icons (male, female, both) for now

### Icon use case 

- form with gender selection
- display gender on user profile

### Concept 

- [X] I have provided valid use cases for each icon.
- [X] I have not added any a brand or logo icon.
- [X] I have not used any hate symbols.
- [X] I have not included any religious or political imagery.

### Author, credits & license

- [X] The icons are solely my own creation.

### Naming 

- [X] I've read and followed the [naming conventions](https://lucide.dev/guide/design/icon-design-guide#naming-conventions)
- [X] I've named icons by what they are rather than their use case.
- [X] I've provided meta JSON files in `icons/[iconName].json`.

### Design 

- [X] I've read and followed the [icon design guidelines](https://lucide.dev/guide/design/icon-design-guide)
- [X] I've made sure that the icons look sharp on low DPI displays.
- [X] I've made sure that the icons look consistent with the icon set in size, optical volume and density.
- [X] I've made sure that the icons are visually centered.
- [X] I've correctly optimized all icons to three points of precision.

## Before Submitting 


jguddas (changes requested):
The lack of inclusive language is a no-go for me.

Do we really want to do this, we have a "no politics" rule for a reason?
My vote is for not adding gender icons.

Trukl:
I don't see how it's political, but whatever.
I admit gender-both may not be the best name
let's add the other genders then, if that's the problem

icons/gender-trans.svg
```diff
@@ -0,0 +1,20 @@
...
+  <path d="M1 6L1 2C1 1.44772 1.44772 1 2 1L6 1" />
+  <path d="M15 7L20.5 1.5" />
+  <path d="M1.5 1.5L7 7" />
+  <circle cx="11" cy="11" r="5" />
+  <path d="M11 16V21" />
+  <path d="M8 19H14" />
+  <path d="M3 7L7 3" />
+</svg>
+
+
```
karsa-mistmere, line 20: This does not follow our design guidelines (1px padding, 2px gap between elements).

karsa-mistmere (changes requested):

icons/gender-trans.svg
karsa-mistmere: Also, I think that since all our current arrows have sharp heads, so should these.

jamiemlaw:
@Trukl you can find some examples of gender icons I designed last week in [this thread](https://github.com/lucide-icons/lucide/issues/560) that abide by the two-pixel-gap rule, if you need inspiration.

I also gave them generic names like `venus` rather than `gender-female`, which is in line with what Font Awesome does for theirs.

I have since edited it to leave a note that I wasn't entirely happy with my design for the `transgender` symbol; and that, any politics aside, getting a workable version of that icon in particular would be one of the prerequisites to adding gender icons as a set to Lucide. So might be worth revisiting that one.

PR #2374, opened 2024-08-16
fix(icons): changed `message-square-dashed` icon

jguddas:
- [x] Other: Icon update

### Description
Use same dashes as `square-dashed`.

## Before Submitting


karsa-mistmere (approved):
Nice work, this is much more cohesive with `box-select` (which we should rename to `square-dashed` 😂 )

ericfennis (approved):

Issue #2371, opened 2024-08-15
Terrible Performance on lucide-solid library


FelipeEmos:
### Package

- [ ] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [ ] lucide-react
- [ ] lucide-react-native
- [X] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

0.427.0

### Can you reproduce this in the latest version?

- [X] Yes
- [ ] No

### Browser

- [X] Chrome/Chromium
- [X] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [ ] Other/not relevant

### Operating system

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [ ] Other/not relevant

### Description

The `lucide-solid` lib is with a terrible performance because it is loading all the icons instead of only the ones the project is using...

I created 2 repos to compare both `lucide-solid` and `lucide-react` and the performance difference is really big.

These repos were set up the same way, I used `bun create vite@latest`, deleted everything in the `App.tsx` and just added one icon.

### Results

|Benchmark Repo|[Lucide React](https://github.com/FelipeEmos/lucide-react-test)|[Lucide Solid](https://github.com/FelipeEmos/lucide-solid-test)|
|---|---|---|
|Time| 250 ms | 3500 ms |
|Amount of Icons being loaded|1|all of them|
|Screenshots|![REACT](https://raw.githubusercontent.com/FelipeEmos/lucide-solid-test/main/docs/react.png)|![SOLID](https://raw.githubusercontent.com/FelipeEmos/lucide-solid-test/main/docs/solid.png)|

### Steps to reproduce

Just import any icon in a Vite SolidJS project from the `lucide-solid` library and see the Lighthouse Network tab

### Checklist

- [X] I have searched if someone has submitted a similar issue before and there weren't any. (Please make sure to also search closed issues, as this issue might already have been resolved.)

FelipeEmos:
Oh, I've just realized after continuing my search about the topic that this is probably a known issue... there's a section in the docs where it says 

[...]

Issue #1675, opened 2023-11-10
Vite+Solid and lucide-solid, Loading failed for the module


krishna-santosh:
### Package

- [ ] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [ ] lucide-react
- [ ] lucide-react-native
- [X] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

0.292.0

### Browser

- [X] Chrome/Chromium
- [X] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [ ] Other/not relevant

### Operating system

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] Other/not relevant

### Description

Using Vite+Solid(TS) 

dev server starts but the webpage doesn't load, Its just blank,  no errors in the terminal.

Got this in firefox console when running the dev server.

```
Loading failed for the module with source “http://localhost:5173/node_modules/.pnpm/lucide-solid@0.292.0_solid-js@1.8.5/node_modules/lucide-solid/dist/source/icons/sun.jsx?v=4ec9cb1e”. localhost:5173:14:47
```
When I build it with `pnpm build`  and `pnpm preview` It works fine. Page loads without any issue.

### Steps to reproduce

1) `pnpm create vite@latest` (and choose solid with TS)
2) `pnpm install lucide-solid`
3) Add any Lucide icon in the App.tsx
Ex:
```ts
import { Sun } from "lucide-solid";

<Sun />
```
4) run `pnpm dev`
5) open `http://localhost:5173` in the browser

### Checklist

- [X] I have searched if someone has submitted a similar issue before and there weren't any. (Please make sure to also search closed issues, as this issue might already have been resolved.)

karsa-mistmere:
Hi @krishna-santosh,

So sorry to hear you've run into an issue, however, I've followed the steps and I'm unfortunately unable to reproduce this issue.
Can you please provide a basic reproduction of this error?

krishna-santosh:
Hello @karsa-mistmere 

Here's a Stackblitz Preview of the bug https://stackblitz.com/edit/solidjs-templates-gbewty?file=src%2FApp.tsx

jguddas:
> Hello @karsa-mistmere
> 
[...]

PR #2375, opened 2024-08-16
fix(lucide-solid): compilation issues when starting up Vite Dev server

ericfennis:
A similar issue we had before with the `lucide-svelte`, the dev server gets overloaded with files to compile. This is because Vite is compiling everything when a file is loaded, including the files that aren't used, see [Vite issue](https://github.com/vitejs/vite/issues/8237) . Since Svelte and Solid depend on pre-processors to optimize performance, we can't use pre-compiled files in the dev-server. 
So for now the only fix is by directly importing icons:

```jsx
import Camera from 'lucide-solid/icons/camera';

// Usage
const App = () => {
  return <Camera color="red" size={48} />;
};

export default App;
```

This PR added support for this. I've tested this with the latest versions of Solid Vite and Solid Start.

Closes: #2371 
Closes: #1675


jguddas (approved):

WarningImHack3r:
Just so you know, Eric or everyone else stumbling upon this PR, you may be interested in using my Vite plugin https://github.com/WarningImHack3r/vite-plugin-lucide-preprocess to rewrite your imports for you. It's been made for Svelte but should work with no problem with Solid, let me know if it doesn't!

PR #2379, opened 2024-08-16
fix(icons): fix stethoscope guideline violations

karsa-mistmere:

- [x] Bug fix

## Before Submitting 


jguddas (approved):
Fewer violations, and even better looks 🎉

PR #2389, opened 2024-08-18
fix(icons): changed `*-user-round` icons

jguddas:
- [x] Other: Icon update

### Description
Changed head size to be the same as in circle-user and square-user.

## Before Submitting


karsa-mistmere:
I prefer them as they currently are.

PR #2403, opened 2024-08-25
feat(icons): added `clock-arrow-in` icon

jguddas:
- [x] New Icon

### Description
Added new `clock-arrow-in` icon.

### Icon use case

- Time range start
- Assign timing information
- Go to time settings

### Concept

### Author, credits & license

- [x] The icons were originally created in #<issueNumber> by @colebemis
- [x] I've based them on the following Lucide icons: `clock`

### Naming

### Design

## Before Submitting


ericfennis (approved):

karsa-mistmere:
While reviewing various clock icon variants already in Lucide and some suggested in various PRs but not yet merged, I now wonder if maybe we should go with the bottom row here:

image

image

(green: in Lucide, blue: proposed in PR)

ericfennis:
Can you explain why we should go with the bottom one?
For me, the top one is more obvious and clear. Or does it create an inconsistency with other icons?

karsa-mistmere:
> Can you explain why we should go with the bottom one? For me, the top one is more obvious and clear. Or does it create an inconsistency with other icons?

These icons use 6by6 arrows:

image


These use 8by8 ones:

image


I think `clock` and `calendar` definitely fit more with the first group than the second.

karsa-mistmere:
@ericfennis: Here's a thorough comparison of all icons using 6by6 and 8by8 arrows and what their counterparts could look like:

image

https://www.figma.com/design/8RY2y2crjJlhT07AMuhU4z/lucide-arrows-6v8?node-id=0-1&t=wEuJIvfHulXcLCCB-1

karsa-mistmere:
> I prefer the bigger version, it's more legible, would be quite the big bang change tho, since I think we should try to be consistent.

I think this legibility comes with a pretty steep trade-off, yes, the arrows are 33% more legible, but in case of `clock-` icons, the clock itself is much less legible:

image

ericfennis:
Hmm this one is hard. I agree with @jguddas, I like the bigger one. On 100% scale is way better to read.
I get the consistency argument, but some icons from the top on the right part of the designs, don't look nice IMO. Arrows are way too small. I think the visual weight argument is in play here. In the end, the icon should look nice and balanced on 100%.
image

ericfennis:
@jguddas I think we maybe should rename this to `clock-arrow-left` 🤔

karsa-mistmere (dismissed):
I agree with @ericfennis on this one as well

ericfennis (approved):

PR #2428, opened 2024-09-03
fix(icons): changed `sliders-horizontal` icon

jguddas:
- [x] Other: Icon update

### Description
Lucidified icons.

## Before Submitting


karsa-mistmere:
I'm against these changes, I prefer the current style, I think it's very clean and minimalistic.

These ones also look too much like https://lucide.dev/icons/settings-2 without being distinct enough.

ericfennis:
I agree with @karsa-mistmere. 
But can be added to lab to research the style

PR #2453, opened 2024-09-13
fix(icons): changed `underline` icon

jguddas:
- [x] Other: Icon update

### Description
Increased size to be same as `square`

## Before Submitting


karsa-mistmere:
I think it's crucial here to compare this change with icons of the same group:

image

I'm on the fence about this one. 🤔

jguddas:
I'm not, the bottom one looks better IMO.

jguddas:
> > I'm not, the bottom one looks better IMO.
> 
> Better on its own, but I think not necessarily better in context, as it's now taller than the rest of the icons. I've tried looking at a lot of text editor toolbars, and more often than not, the underline icon is the same height as bold and italic, hence my conflict. 🤔
> 
> This is also a pretty common variant in editor toolbars (a much smaller `U` than `B` and `I` are): image

I feel like it looks better in context as well, I feel like currently the `U` is very small and feels out of place.

jguddas:
What about making the rest of the icons bigger, they are also a bit small.
<a title="Open lucide studio" target="_blank" href="https://lucide-studio.vercel.app/edit?name=bold&base=bold&value=%3Cpath+d=%22M6+12h9a4+4+0+0+1+0+8H7a1+1+0+0+1-1-1V5a1+1+0+0+1+1-1h7a4+4+0+0+1+0+8%22+/%3E%3Crect+x=%223%22+y=%223%22+width=%2218%22+height=%2218%22+rx=%222%22+/%3E">icons<br/>Open lucide studio</a>

karsa-mistmere:
Could work:

image
=>
image

ericfennis:
Not sure about this change. The icons should have the same height as the other text editor icons.
🤔

PR #2468, opened 2024-09-17
fix(icons): redoing the `accessibility` icon

jguddas:
- [x] Other: Icon update

### Description
Redoing the accessibility icon with less sharpness and without basing it on the accessibleicon project icon.

> The Accessible Icon is owned by Triangle, Inc. and was created by Sara Hendren & Brian Glenney. The Accessible Icon is licensed under a Creative Commons Attribution-NonCommercial-ShareAlike 3.0 Unported License.

## Before Submitting

### DIFF
--- a/icons/accessibility.json
+++ b/icons/accessibility.json
@@ -1,7 +1,6 @@
 {
   "$schema": "../icon.schema.json",
   "contributors": [
-    "karsa-mistmere",
     "jguddas"
   ],
   "tags": [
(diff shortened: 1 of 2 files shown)


karsa-mistmere:
Still not a fan of the rag doll approach, it feels contrary to what the aim of Accessible Icon is all about.

I think we should improve our licensing capabilities instead (not that there isn't room for improvement with the icon).

PR #2469, opened 2024-09-17
fix(icons): changed `hourglass` icon

jguddas:
- [x] Other: Icon update

### Description
Decreased height.

## Before Submitting


karsa-mistmere (commented):
I think this has too little visual weight this way, as it's much narrower than square is.

jguddas:
You are right.

Issue #2344, opened 2024-07-30
Update Hand Icons


jamiemlaw:
### Package

- [X] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [ ] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] all JS packages
- [ ] site

### Description

There's something that's always felt a little off to me about the hand icons. I think they look a little wide and stubby. I'm not sure if it has to do with keeping a good optical volume, or because it's the only way to include all five fingers. I've also noticed that on icons like `pointer`, the overlap of the thumb make it look like we're seeing the palm of the hand, but the fingers are curled away, so I guess it's the back of the hand?

<img width="269" alt="image" src="https://github.com/user-attachments/assets/919fa462-3492-40b6-a1d3-d4dcf695a03d">

So, I've had a go at designing some alternatives that only have four fingers, giving the hand a more slender shape.

<img width="911" alt="image" src="https://github.com/user-attachments/assets/419e7dcc-710c-49d2-8e05-d1887fd17f86">

As well as all the current hand-related icons in Lucide, I've included a few others just to show off the repertoire of the new style. I personally feel they're quite expressive. It also includes some ideas requested in issues [#1330](https://github.com/lucide-icons/lucide/issues/1330) and [#1740](https://github.com/lucide-icons/lucide/issues/1740).

CodePen: https://codepen.io/jamiemlaw/pen/xxogQwG/a2c4b71d0628cc9988adaa898e762855

I'm just interested to hear your thoughts. One possible criticism is that certain icons, particularly, `grab` have a small optical volume, just in order to keep them to the same scale as the other icons. I'm also willing to accept that not all of the icons need to be added; some of them are just there for flavour. But it's been fun designing them nonetheless.

### Use cases

n/a

### Checklist

- [X] I have searched the existing issues to make sure this bug has not already been reported.

jguddas:
Those are just lovely, very lucidy IMO.

The thumb feels a bit long in the not curled up icons to me.

karsa-mistmere:
Hi @jamiemlaw, excellent work on these, they're really nice and crisp.

That being said, I share @jguddas's criticism of there being something not quite right about the thumbs, personally, I'd add more rounding to the base or even shorten it as Jakob suggested:
[...]

PR #2480, opened 2024-09-19
fix(icons): changed `hand` and `pointer` icons

jamiemlaw:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other: icon update

### Description

This PR proposes a new series of hand icons, with a more slender and expressive shape, at the cost of a finger.

## Before Submitting 


karsa-mistmere:
As stated in #2344, I'm not really into these changes.

jamiemlaw:
> As stated in #2344, I'm not really into these changes.

Completely fine. I'll leave it between you and @jguddas to battle it out! 😬

karsa-mistmere:
> Completely fine. I'll leave it between you and @jguddas to battle it out! 😬

Let's also not forget that @ericfennis has the deciding vote here.

My case is: there is hardly any precedent of four fingered hand icons, as for hand emojis, [there isn't a single design](https://emojipedia.org/raised-hand#designs) that uses four fingers.

Switching to these will make any future icons that rely on pinky, ring or middle fingers almost completely unfeasible — `hand-metal` is a good example, it's just not recognisable for what it is anymore.

jguddas:
I love them ❤️

ericfennis:
I agree with @karsa-mistmere, we shouldn't do this if we aren't sure about future designs. 

For this reason:
> Switching to these will make any future icons that rely on pinky, ring or middle fingers almost completely unfeasible — hand-metal is a good example, it's just not recognisable for what it is anymore.

But I'm open to researching this design and will suggest that we add this to [lucide-icons/lucide-lab](https://github.com/lucide-icons/lucide-lab) to explore

jamiemlaw:
I feel like this is pretty much a solved problem. Take, for example, Homer Simpson giving the finger, or Fry giving the "Spock" symbol in Futurama—gestures that you would think require all five fingers, but don't. Cartoonists for decades have circumvented the issue of not having all five fingers on one hand.

It also opens up the possibility to include icons that would otherwise be too cramped under the five-finger system, maybe like a hand with splayed fingers, a waving hand, a finger-tap. So, I'd argue on balance it would make more icons possible than it would prohibit.

However, if there's no consensus, I would love to see these at least in the`lab` branch.

jguddas:
Learned something new today:

> Japanese characters usually have all five fingers, and there are different reasons for that: an aversion towards number four due to its similar pronunciation with the word for “death”, the four fingers are considered an offensive reference to the burakumin caste, and the Yakuza tradition of cutting a finger off as punishment.
As a result, many cartoon characters who are typically drawn with four fingers had to be edited to have the fifth finger in order to avoid controversy in Japan.

PR #2491, opened 2024-09-23
fix(icons): Update trash icon path order

dotspencer:

- [X] Bug fix

### Description

Updates path order on `trash` and `trash-2` to preserve outline when using fill. The fill on the open path was overlapping the top line. Fix was the simply move the top line to be the last child in the svg.

**Trash before / after**

**Trash 2 before / after**

## Before Submitting 

### DIFF
--- a/icons/message-square-lock.json
+++ b/icons/message-square-lock.json
@@ -1,0 +1,23 @@
+{
+  "$schema": "../icon.schema.json",
+  "contributors": [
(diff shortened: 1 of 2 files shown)


jguddas:
Sorry, we do not support fills and mighty change the path order at any time.

dotspencer:
@jguddas Unfortunate, ok

PR #2502, opened 2024-09-27
feat: added request-review workflow

jguddas:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other: GitHub workflow

### Description

Adds all the contributor to the icon you are changing as reviewers.

## Before Submitting 


jguddas:
Hmm, I think this might actually not work as intended, I can't add most people, even by hand.

ericfennis:
Hmm not sure if we want this. But I like the idea.

For old icons colebemis is listed as a contributor. Since he is not really involved in this project. Not sure if he should be invited.

jguddas:
> Hmm not sure if we want this. But I like the idea.
> 
> For old icons colebemis is listed as a contributor. Since he is not really involved in this project. Not sure if he should be invited.

We could also mention them in the PR comment.

Another alternative is automatically generated code owners, not sure how that could work tho.

ericfennis:
Hmm no I think we can keep this idea. But we should add an if statement for `colebemis` username.

jguddas:
> Hmm no I think we can keep this idea. But we should add an if statement for `colebemis` username.

I don't think he will get added by this workflow, I think it will ignore him

> Hmm no I think we can keep this idea. But we should add an if statement for `colebemis` username.

It will not add him, I'm not sure what the requirements are, but this is the output I get:
```
gh pr edit 2502 --add-reviewer colebemis
'colebemis' not found
```

ericfennis:
@jguddas Thanks for checking! Maybe @karsa-mistmere wants to leave a review here as well?

ericfennis (approved):

Issue #2498, opened 2024-09-26
NextJs 14 className hydration bug.


nandy1990:
### Package

- [ ] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [X] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

0.446.0

### Can you reproduce this in the latest version?

- [X] Yes
- [ ] No

### Browser

- [X] Chrome/Chromium
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [ ] Other/not relevant

### Operating system

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [ ] Other/not relevant

### Description

When using "use client" className differs by trailing whitespace between server and client.
Inside off createLucideIcon.js className  gets joined together by following line 23: 

className: ["lucide", `lucide-${toKebabCase(iconName)}`, className].join(" ")

so if className is empty you end up  having a trailing space.

Console output in browser:

Warning: Prop `className` did not match. Server: "lucide lucide-euro " Client: "lucide lucide-euro"
    at svg...

### Steps to reproduce

Any file with "use client"  and rendering any icon without className specified leads to this warning/error

### Checklist

- [X] I have searched if someone has submitted a similar issue before and there weren't any. (Please make sure to also search closed issues, as this issue might already have been resolved.)

PR #2503, opened 2024-09-28
tests(shared): added mergeClasses tests

jguddas:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other: Tests

### Description

Checking that we don't have trailing spaces and such.

Related to #2498

## Before Submitting 

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -7,5 +7,12 @@
   "types": "src/index.ts",
   "type": "module",
   "author": "",
-  "license": "ISC"
+  "license": "ISC",
+  "scripts": {
+    "test": "vitest run",
+    "test:watch": "vitest watch"
+  },
+  "devDependencies": {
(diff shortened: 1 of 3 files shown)


ericfennis:
Awesome work @jguddas !

We need also adding a gh workflow for this to check when PRS are created.

jguddas:
> Awesome work @jguddas !
> 
> We need also adding a gh workflow for this to check when PRS are created.

I would prefer a `pnpm run --filter=[$target...HEAD] test` that runs over only changed workspaces, that would reduce the CI cost and I don't think we have ever cought anything with those CI Checks anyway so it's pretty safe.

ericfennis:
Yeah true, we can add it later.

ericfennis (approved):

PR #2506, opened 2024-09-29
fix(icons): rename `box-select` to `square-dashed`

chessurisme:

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

The icon named `box-select` violates the naming convention as outlined in [Naming convention in Lucide's design principle #3](https://lucide.dev/guide/design/icon-design-guide#naming-conventions). Instead of using `box-select`, we will rename it to `square-dashed`, following the convention used in names like `square-dashed-mouse-pointer`.

## Before Submitting 


jguddas (changes requested):
Don't forget to add the alias field for backwards compatibility.

chessurisme:
How do I? I haven't read any ‘alias’ or same thing in the documentation. Would you point out similar icons that renamed like `box-select`? Thanks.

jguddas:
> How do I? I haven't read any ‘alias’ or same thing in the documentation. Would you point out similar icons that renamed like `box-select`? Thanks.

You can search the code it's in the .json, I'm on the phone right now, that's the best I can do.

chessurisme:
I already added an aliases.

jguddas (approved):

PR #2522, opened 2024-10-08
fix(shared): Add trim to `mergeClasses`

ericfennis:
fixes: #2498

Adds trim to `mergeClasses` function. Possibly fixes hydration issues in NextJS.
Also added a test and adjusted GH actions to run this change on each package.


packages/shared/src/utils.ts
```diff
@@ -43,4 +43,5 @@ export const mergeClasses = <ClassType = string | undefined | null>(...classes:
     .filter((className, index, array) => {
       return Boolean(className) && array.indexOf(className) === index;
     })
-    .join(' ');
+    .join(' ')
+    .trim();
```
jguddas, line 47: What about doing `return Boolean(className.trim()) && array.indexOf(className) === index;` instead?

That way you don't get something weird if you do `mergeClasses('lucide', ' ', 'lucide-circle')`.

Not sure about the performance impact tho.

ericfennis: Hmm good idea, I will add that one too! Performance is negligible.

jguddas (approved):

benjick:
@ericfennis this needs `prettier -w`

packages/shared/src/utils.ts
```diff
@@ -41,6 +41,7 @@ export const toPascalCase = <T extends string>(string: T): CamelToPascal<T> => {
 export const mergeClasses = <ClassType = string | undefined | null>(...classes: ClassType[]) =>
   classes
     .filter((className, index, array) => {
-      return Boolean(className) && array.indexOf(className) === index;
+      return Boolean(className) && (className as string).trim() !== '' && array.indexOf(className) === index;
```
benjick, line 44:
```suggestion
      return (
        Boolean(className) &&
        (className as string).trim() !== '' &&
        array.indexOf(className) === index
      );
```

to pass prettier check

ericfennis: Thanks this looks more readable!

PR #2577, opened 2024-11-02
fix(icons): changed `layout-list` icon

jguddas:
- [x] Other: Icon update

### Description
Remove one of the lines to be consistent with other list icons.

## Before Submitting


ericfennis (approved):

karsa-mistmere (changes requested):
For me, this communicates something else entirely than before. My understanding of the current icon vs. the proposal:

image

jguddas:
I agree

PR #2581, opened 2024-11-04
feat(docs): Make input autofocus in the /icons page

danilo-leal:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other:

### Description

This PR simply adds the `autofocus` attribute to the `Input` component on the `/icons` page. This allows users to immediately type their search query, which then avoids an extra click. Here's a quick video with the before & after:

https://github.com/user-attachments/assets/e33d9a9e-4791-4506-8646-7838f5095700

Love Lucide—we use it extensively at Zed! 😃 I basically have the docs tab open every day.

## Before Submitting 


ericfennis:
@danilo-leal Thanks for the PR. Awesome that you use Lucide at Zed!
I will review this behavior. I'm not sure from a UX perspective yet, I will get back to you!

danilo-leal:
Absolutely! I'm constantly refreshing https://lucide.dev/icons/, so having the input automatically focused would be convenient so I don't have to reach for the mouse, but definitely feel free to ditch if you think it doesn't make sense 😊

ericfennis (approved):
Let's try it out!

PR #2596, opened 2024-11-09
fix(icons): combined `layers` and `layers-3` icon

jguddas:
- [x] Other: Icon update

### Description
Arcified corners.

## Before Submitting


tools/build-icons/utils/deprecationReasonTemplate.mjs
```diff
@@ -9,6 +9,8 @@ export function deprecationReasonTemplate(
   switch (deprecationReason) {
     case 'alias.typo':
       return `Renamed because of typo, use {@link ${componentName}} instead.${removalNotice}`;
+    case 'alias.duplicate':
+      return `The icon was combined with another icon that shared the same usecase, use {@link ${componentName}} instead.${removalNotice}`;
```
jguddas, line 13:
```suggestion
      return `The icon was combined with another icon that shares the same use case, use {@link ${componentName}} instead.${removalNotice}`;
```

ericfennis (approved):

karsa-mistmere (approved):

PR #2610, opened 2024-11-13
feat(icons): upgrade snowflake icons to match real life

karsa-mistmere:

- [x] Bug fix

### Description
Real snowflakes have six-fold rotational symmetry, this PR upgrades snowflake icons to look more accurate and have better uniformity between them.

## Before Submitting 


jguddas:
What about arcifying the inner hexagon?

karsa-mistmere:
> What about arcifying the inner hexagon?

Does that even make sense here? 

At 1x, there isn't any perceptible difference, I think:
image

When scaled up, I think no arcs is definitely the way to go:
image

jguddas (approved):

jguddas (approved):

ericfennis (approved):

PR #2619, opened 2024-11-19
docs(showcase): Add reflex to showcase

tgberkeley:

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other:

### Description
This PR adds the company Reflex (https://reflex.dev) to the Showcase page. This Repo has over 20,000 GitHub stars and uses Lucide icons as its only icon type in the library.

## Before Submitting 


tgberkeley:
@jguddas what is the reason for the 👎 above, am I missing something in this PR?

docs/public/company-logos/reflex-dark.svg
karsa-mistmere, line 1: You've mixed up light with dark:
image
image

And maybe "Used in" would be the more appropriate category.

I'll leave the decision up to @ericfennis

tgberkeley: @karsa-mistmere yes I noticed this and fixed it in a follow up commit

karsa-mistmere: > @tgberkeley Can you confirm, lucide is part of the tool or is it only used on the docs page?

I think part of the tool: https://reflex.dev/docs/library/data-display/icon/

jguddas:
> @jguddas what is the reason for the 👎 above, am I missing something in this PR?

They have fewer users than us, so it's more of an advertisement for them than it is for us.

I would rather expect there to be companies such as [adobe](https://github.com/search?q=org%3Aadobe%20lucide&type=code).

But all in all I'm very much against having such a section, it's legally and morally a pretty dark gray area.

docs/public/company-logos/reflex-dark.svg
ericfennis: In that case this should be moved to the "used in" section.

ericfennis (changes requested):
Should be moved to the "Used in" section.

tgberkeley:
@ericfennis just moved it to the used in section

tgberkeley:
@ericfennis is there anything else here that I need to edit to get this PR approved?

ericfennis (approved):
LGTM!

PR #2621, opened 2024-11-19
[DOCS] Svelte 5 specifics for making components with the lucide-svelte package

boyer-nicolas:
Svelte 5 specifics including removing the "ComponentType" typescript type as components are now static.  Also merged some imports into one-liners as we don't need to import each icon in a single line.

- [ ] New Icon
- [ ] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other:

### Description
Added Svelte 5 specifics including how to make components with the now deprecated `ComponentType` typescript type.

## Before Submitting 


ericfennis (approved):
Awesome thank you very much!

PR #2671, opened 2024-12-07
docs: improve badges

realguse:
- [x] Documentation update

### Description
Switches the style of the badges in the readme to `flat-square` or `for-the-badge` to get a more modern look, updates the alt tags for the badges, the badges now use same "color declaration" (small-case, hex code, excluding `%23` which is not necessary etc.)

This is also one of the cleanest readmes I've seen in a long time, so that made it really easy to edit.

## Before Submitting


realguse:
Closing because I'll move it to a bigger pr that generally improves the whole readme (not to drastic changes though)

PR #2674, opened 2024-12-07
docs(readme): improve packages table

realguse:
## What is the purpose of this pull request?
- [x] Documentation update

### Description
Tries to make the packages table in the main README.md more clean and modern.

## Before Submitting
- [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [x] I've checked if there was an existing PR that solves the same issue.

PR #2704, opened 2024-12-21
docs: improve badges

realguse:
Originally in #2674

- [x] Documentation update

### Description
This PR aims to update the badges to a more modern style (`flat-square` and `for-the-badge`).

## Before Submitting


realguse:
@jguddas, based on you contrast "issue", i have not found any solution. There doesn't seem to be a option to change the foreground/textcolor, sorry

jguddas:
I did some testing of the dynamic color switch logic, seams to only trigger when using hex values like by for example adding `&color=febcda` to the end of the url, weird.

Maybe worth doing A bug report for, needs more testing tho, can't right now, I'm only on the phone.

What I suggest we do is find the hex value of the current color and try if adding it to the url helps.

realguse:
Yeah, look likes that's the problem. I'll only add the `color` param to the ones that actually switches color automatically to keep things simple

realguse:
After testing the `color` param on the badges, none of them changed their foreground color. So that means the contrast is sufficient enough according to shields.io

README.md
```diff
@@ -7,10 +7,10 @@
   </a>
 </p>
 <p align="center">
-  <a href="https://github.com/lucide-icons/lucide/blob/main/LICENSE"><img src="https://img.shields.io/npm/l/lucide" alt="license"></a>
-  <a href="https://www.figma.com/community/plugin/939567362549682242/Lucide-Icons"><img src="https://img.shields.io/badge/Figma-F24E1E?logo=figma&logoColor=white" alt="figma installs"></a>
-  <a href="https://github.com/lucide-icons/lucide/actions/workflows/release.yml"><img src="https://github.com/lucide-icons/lucide/actions/workflows/release.yml/badge.svg" alt="build status"></a>
-  <a href="https://discord.gg/EH6nSts"><img src="https://img.shields.io/discord/723074157486800936?label=chat&logo=discord&logoColor=%23ffffff&colorB=%237289DA" alt="discord chat"></a>
+  <a href="https://github.com/lucide-icons/lucide/blob/main/LICENSE"><img src="https://img.shields.io/npm/l/lucide?style=for-the-badge" alt="license"></a>
```
jguddas, line 10:
```suggestion
  <a href="https://github.com/lucide-icons/lucide/blob/main/LICENSE"><img src="https://img.shields.io/npm/l/lucide?style=for-the-badge&color=008c14" alt="license"></a>
```
I like @JoshuaKGoldberg s suggestion of changing the background colors to be a bit darker.
https://github.com/badges/shields/issues/5497

What do you think?

realguse: Yeah, that feels more right. I'll do it now

realguse: I feel like `#008c14` gets too dark, another option is to use `brightgreen` which looks like this:
image

realguse:
wtf happend

realguse:
wtf happend

README.md
jguddas: Minimum contrast to get me to be happy is 4.5.

realguse: Will fix it when i get back my changes which I of course lost

realguse: `#1A871B`

realguse: or should we just use `brightgreen`/`#45CC11` as the color and make the foreground black

realguse: Possible colors:
image

image

image

image

jguddas: We could also make all the badges the lucide brand color, still pretty low tho.

realguse: Yeah, red and white, what badge should have which

realguse: Or should we even care about the contrast, there's a bunch (and I mean a bunch) using the `brightgreen` for example, which isn't contrast sufficient enough

jguddas: If it's not necessary to be readable, it's not necessary and should be removed entirely, I'm pretty strongly opinionated on this one :)

realguse: Found some colors with high enough (very close at least) contrast ratio. I'm not updating the other ones until we've found the final colors

image

realguse:
Im so sorry if i pinged all of you now, I do not know what happened

PR #2817, opened 2025-02-19
feat(script): add generate icon names script

zaaakher:
Related to https://github.com/lucide-icons/lucide/discussions/1778

I just wanted the list of icon names to use to show them in an icon picker for my app users but the `tags.json` and `icon-nodes.json` included tags or svg paths that I didn't need. And deleting the extra parts manually would take forever. So I figured we can add a `generate:icon-names` script which would generate `icon-names.json` that would look like this:

```json
{
  "created_at": "2025-02-19T13:11:15.475Z",
  "icon_names": [
    "a-arrow-down",
    "a-arrow-up",
    "a-large-small",
    "accessibility",
    "activity",
    "air-vent",
    "airplay",
    "alarm-clock",
     // the rest of the icons
```

- [ ] New Icon
- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other:

## Before Submitting 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ packages/**/dynamic.mjs
 packages/**/LICENSE
 categories.json
 tags.json
+icon-names.json
 .vercel
 
 # docs
--- a/.npmignore
+++ b/.npmignore
@@ -13,3 +13,4 @@ netlify.toml
 rollup.config.js
 rollup.plugins.js
 tags.json
+icon-names.json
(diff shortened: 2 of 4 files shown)


jguddas:
One icon can have multiple names, maybe a structure like `[['home', 'house'], ...]` could be a thing?

jguddas:
But thinking about the use case you should really think about maybe using tag Jason so you can show related icons and not only exact word matches for icon search.

zaaakher:
> One icon can have multiple names

Oh I totally missed that part.

> But thinking about the use case you should really think about maybe using tag Jason so you can show related icons and not only exact word matches for icon search.

I was gonna implement the icon picker without a search ability but I think you're right, there are so many icons that a search is almost essential in my use case.


I suppose this PR is redundant, feel free to close it if you concur.

PR #2823, opened 2025-02-20
fix(icons): changed `mic-off` icon

lieonlion:
- [x] Other: Icon update

### Description
Fixed icon `mic-off`. Lower curve cut too far down and the bottom line was visible from above.
image
Current `mic-off` compared to the normal `mic`.
image
image
Fixed `mic-off`.
image

## Before Submitting


ericfennis (approved):
Nice thanks!

PR #2764, opened 2025-01-31
ci(actions): Update & cleanup actions

ericfennis:
- Removed unused actions & workflows
- Updates action versions, we got some warnings that we need to update some versions.
- Removed the `pull_request_target` workflows due to security reasons.

PR #2854, opened 2025-02-28
ci(pr-comment): Fix icon preview comment on PRs

ericfennis:
After closing #2764
Icon PR comment is not working anymore, since it needs write access. The `GITHUB_TOKEN` with `on:  pull_request:` has only read access. With `workflow_run:` we also get write access to PRs and make sure it is secure.

This solution is based on a blog article form GH. https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

### DIFF
--- a/.github/workflows/comment-icon-preview.yml
+++ b/.github/workflows/comment-icon-preview.yml
@@ -2,7 +2,7 @@ name: Icon preview comment
 
 on:
   workflow_run:
-    workflows: ["Pull request icon preview"]
+    workflows: ['Pull request icon preview']
     types:
       - completed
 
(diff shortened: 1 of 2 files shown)


jguddas:
on pull_request does not run on pull requests from forks, this will not fix our core issue.

From the blog post:

Having said that, mixing pull_request_target with an explicit PR checkout is not always vulnerable. The workflow may, for example:

- Reformat and commit the code
- **Checkout both base and head repositories and generate a diff**
- Run grep on the checked out source.

Generally speaking, when the PR contents are treated as passive data, i.e. not in a position of influence over the build/testing process, it is safe.

ericfennis:
@jguddas yes they will still run. See the latest runs from PRs.

But it won't have permission to write a comment to PR, thats why the pr icon preview comment doesn't work anymore.
But I've a fix based on the blogpost see #2854

jguddas:
Okay, I just checked and if it's only blocked for the first time someone contributed, that's kinda okay.

> By default, all first-time contributors require approval to run workflows.

https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks

PR #2859, opened 2025-03-03
fix(shared): consistent kebab case conversion

### LINKED ISSUE #2858: Inconsistent kebab-casing between class attribute and iconNames
### Package

- [x] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [x] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

0.477.0

- [x] Yes
- [ ] No

### Browser

- [ ] Chrome/Chromium
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [x] Other/not relevant

### Operating system

- [ ] Windows
- [ ] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [x] Other/not relevant

### Description

When rendering an icon with `lucide-react`, the kebab-cased class name is inconsistent with the corresponding entry in `iconNames`.

Example with `AArrowDown` icon:
- Class attribute: `class="lucide lucide-aarrow-down"`
- In `iconNames`: `a-arrow-down`

Example with `Loader2` icon:
- Class attribute: `class="lucide lucide-loader2"`
- In `iconNames`: `loader-2`

Root cause appears to be in the `toKebabCase` function.

### Steps to reproduce

N/A

### Checklist

danteissaias:
Closes #2858 

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description
When rendering an icon with `lucide-react`, the kebab-cased class name is inconsistent with the corresponding entry in `iconNames`.

Example with `AArrowDown` icon:
- Class attribute: `class="lucide lucide-aarrow-down"`
- In `iconNames`: `a-arrow-down`

Example with `Loader2` icon:
- Class attribute: `class="lucide lucide-loader2"`
- In `iconNames`: `loader-2`

This PR fixes the ⁠toKebabCase function to properly handle these cases. I've also added relevant test cases.

Note: there are some inherent inconsistencies with icon naming like `ArrowUp10` → `arrow-up-1-0` and `Clock10` → `clock-10`, but there are already aliases for those (`arrow-up-10` → `arrow-up-1-0`).

## Before Submitting 

### DIFF
--- a/packages/lucide-preact/tests/__snapshots__/lucide-preact.spec.tsx.snap
+++ b/packages/lucide-preact/tests/__snapshots__/lucide-preact.spec.tsx.snap
@@ -10,7 +10,7 @@ exports[`Using lucide icon components > should adjust the size, stroke color and
      stroke-width="4"
      stroke-linecap="round"
      stroke-linejoin="round"
-     class="lucide lucide-grid3x3"
+     class="lucide lucide-grid-3x3"
 >
   <rect width="18"
         height="18"
@@ -40,7 +40,7 @@ exports[`Using lucide icon components > should not scale the strokeWidth when ab
      stroke-width="1"
      stroke-linecap="round"
      stroke-linejoin="round"
-     class="lucide lucide-grid3x3"
+     class="lucide lucide-grid-3x3"
 >
   <rect width="18"
         height="18"
@@ -70,7 +70,7 @@ exports[`Using lucide icon components > should render an component 1`] = `
      stroke-width="2"
      stroke-linecap="round"
      stroke-linejoin="round"
-     class="lucide lucide-grid3x3"
+     class="lucide lucide-grid-3x3"
 >
   <rect width="18"
         height="18"
--- a/packages/lucide-react/tests/__snapshots__/lucide-react.spec.tsx.snap
+++ b/packages/lucide-react/tests/__snapshots__/lucide-react.spec.tsx.snap
@@ -10,7 +10,7 @@ exports[`Using lucide icon components > should adjust the size, stroke color and
      stroke-width="4"
      stroke-linecap="round"
      stroke-linejoin="round"
-     class="lucide lucide-grid3x3"
+     class="lucide lucide-grid-3x3"
 >
   <rect width="18"
         height="18"
@@ -40,7 +40,7 @@ exports[`Using lucide icon components > should not scale the strokeWidth when ab
      stroke-width="1"
      stroke-linecap="round"
      stroke-linejoin="round"
-     class="lucide lucide-grid3x3"
+     class="lucide lucide-grid-3x3"
 >
   <rect width="18"
         height="18"
@@ -70,7 +70,7 @@ exports[`Using lucide icon components > should render an component 1`] = `
      stroke-width="2"
      stroke-linecap="round"
      stroke-linejoin="round"
-     class="lucide lucide-grid3x3"
+     class="lucide lucide-grid-3x3"
 >
   <rect width="18"
         height="18"
--- a/packages/lucide-solid/tests/__snapshots__/lucide-solid.spec.tsx.snap
+++ b/packages/lucide-solid/tests/__snapshots__/lucide-solid.spec.tsx.snap
@@ -10,7 +10,7 @@ exports[`Using lucide icon components > should adjust the size, stroke color and
      height="48"
      stroke="red"
      stroke-width="4"
-     class="lucide lucide-icon lucide-grid3x3"
+     class="lucide lucide-icon lucide-grid-3x3"
      data-testid="grid-icon"
 >
   <rect width="18"
@@ -50,7 +50,7 @@ exports[`Using lucide icon components > should not scale the strokeWidth when ab
      height="48"
      stroke="red"
      stroke-width="1"
-     class="lucide lucide-icon lucide-grid3x3"
+     class="lucide lucide-icon lucide-grid-3x3"
      data-testid="grid-icon"
 >
   <rect width="18"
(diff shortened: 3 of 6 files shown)


jguddas:
I have a fun requirement for you, **backwards compatibility**.

We have a deprecation mechanism([the aliases field](https://github.com/lucide-icons/lucide/blob/main/icon.schema.json#L142)), you could add a new `deprecationReason`: `alias.case` or something.

```json5
{
  aliases: [
    { name: "loader2", deprecationReason: "alias.case" }
  ]
}
```

danteissaias:
> I have a fun requirement for you, **backwards compatibility**.
> 
> We have a deprecation mechanism([the aliases field](https://github.com/lucide-icons/lucide/blob/main/icon.schema.json#L142)), you could add a new `deprecationReason`: `alias.case` or something.
> 
> ```js
> {
>   aliases: [
>     { name: "loader2", deprecationReason: "alias.case" }
>   ]
> }
> ```

Correct me if I am wrong, but isn't `toKebabCase` only used to generate the `className`? I'm confused as to the need for aliases here.

jguddas:
> > I have a fun requirement for you, **backwards compatibility**.
> > We have a deprecation mechanism([the aliases field](https://github.com/lucide-icons/lucide/blob/main/icon.schema.json#L142)), you could add a new `deprecationReason`: `alias.case` or something.
> > ```js
> > {
> >   aliases: [
> >     { name: "loader2", deprecationReason: "alias.case" }
> >   ]
> > }
> > ```
> 
> Correct me if I am wrong, but isn't `toKebabCase` only used to generate the `className`? I'm confused as to the need for aliases here.

If we merge your change the following code will stop working.

```jsx
<style>
  .lucide-loader2 {
    animation: var(--animate-spin); /* spin 1s linear infinite */
  }
  @keyframes spin { to { transform: rotate(360deg) } }
</style>
<LucideLoader2/>
```

danteissaias:
> > > I have a fun requirement for you, **backwards compatibility**.
> > > We have a deprecation mechanism([the aliases field](https://github.com/lucide-icons/lucide/blob/main/icon.schema.json#L142)), you could add a new `deprecationReason`: `alias.case` or something.
> > > ```js
> > > {
> > >   aliases: [
> > >     { name: "loader2", deprecationReason: "alias.case" }
> > >   ]
> > > }
> > > ```
> > 
> > 
> > Correct me if I am wrong, but isn't `toKebabCase` only used to generate the `className`? I'm confused as to the need for aliases here.
> 
> If we merge your change the following code will stop working.
> 
> ```js
> <style>
>   .lucide-loader2 {
>     animation: var(--animate-spin); /* spin 1s linear infinite */
>   }
>   @keyframes spin { to { transform: rotate(360deg) } }
> </style>
> <LucideLoader2/>
> ```

I see you're point, however I don't see how aliases would solve this. `loader-2` and `loader2` are both `Loader2` in pascal case so these would be the same React component.

At the moment, this behavior isn't even portable across different lucide packages at the moment (e.g. it would be `lucide-loader-2` in `lucide` and `lucide-loader2` in `lucide-react`).

Other than adding both classes to the generated element (e.g. `lucide lucide-loader-2 lucide-loader2`), I don't see a way to make this backwards compatible. Do you have any other ideas?

danteissaias:
> > > I have a fun requirement for you, **backwards compatibility**.
> > > We have a deprecation mechanism([the aliases field](https://github.com/lucide-icons/lucide/blob/main/icon.schema.json#L142)), you could add a new `deprecationReason`: `alias.case` or something.
> > > ```js
> > > {
> > >   aliases: [
> > >     { name: "loader2", deprecationReason: "alias.case" }
> > >   ]
> > > }
> > > ```
> > 
> > 
> > Correct me if I am wrong, but isn't `toKebabCase` only used to generate the `className`? I'm confused as to the need for aliases here.
> 
> If we merge your change the following code will stop working.
> 
> ```js
> <style>
>   .lucide-loader2 {
>     animation: var(--animate-spin); /* spin 1s linear infinite */
>   }
>   @keyframes spin { to { transform: rotate(360deg) } }
> </style>
> <LucideLoader2/>
> ```

In fact, your example has not worked since https://github.com/lucide-icons/lucide/commit/6fbd5ee06ad3fa72acd5162fbaa62d827f67df30. `loader-2` was renamed to `loader-circle` but in lucide-react the classes don't reflect aliases but rather than original icon name.

danteissaias:
I see you're point, however I don't see how aliases would solve this. `loader-2` and `loader2` are both `Loader2` in pascal case so these would be the same React component.

The example you provided hasn't worked since https://github.com/lucide-icons/lucide/commit/6fbd5ee06ad3fa72acd5162fbaa62d827f67df30 (when `loader-2` was renamed to `loader-circle`) since the classes that are generated always reflect the icon's original name rather than an alias.

This class isn't currently documented and I think it'd be difficult to document in it's current state since classes can vary across different packages. For example it would be `lucide-loader-2` in `lucide` and in `lucide-loader2` in `lucide-react`,

To preserve backwards compatibility, one approach could be to include both classes, such as ⁠`lucide lucide-loader-2 lucide-loader2`. Do you have any other ideas?

danteissaias:
Thanks for raising these issues, closing this PR in favor of another approach #2878.

PR #2861, opened 2025-03-03
chore: icon alias improvements

jguddas:
- enforce that the shape of `aliases` is always an array of objects.
- made the alias object required to have a `name` field, no `{}` allowed.
- made it so you can't have additional properties in the alias object.
- made it so you need to set `depracted: true` if you provide `depracationReason` or `toBeRemovedInVersion`.

And adjusted all the files to comply with the new more strict schema.


ericfennis:
@jguddas Can you update this one? We should get this one merged.

ericfennis (approved):
Super nice! 🔥

PR #2871, opened 2025-03-06
Fix CVE-2024-21538

anupamme:
This PR fixes CVE-2024-21538 by updating package dependencies.


ericfennis (changes requested):
Can you update the lock file?

PR #2877, opened 2025-03-07
ci(node): Use correct node version

ericfennis:
- Make sure we use the defined node version from the package.json in github actions


Edgarruiz8585:
>@Edgarruuz8585
use-correct-node-version

Edgarruiz8585:
>@Edgarruuz8585 * Make sure we use the defined node version from the package.json in github actions

/run_Edgarruiz8585.

PR #2895, opened 2025-03-12
feat(icons): added `share-ios` icon

oosawy:
- [x] New Icon

### Description
Added new `share-ios` icon.

### Icon use case

- The `share` icon but feels more like Apple.
- Instruction indicating the Safari Share icon.
- Enhancing the Platform-Adaptive experience.

### Concept

### Author, credits & license

- [x] The icons were originally created in #<issueNumber> by @colebemis
- [x] I've based them on the following Lucide icons: `share`

### Naming

### Design

## Before Submitting


karsa-mistmere:
Hi @oosawy,

I can't really see us moving forward with this as is for the following reasons:
- the top of the square is half a pixel off grid, and would need better pixel perfection
- the name `share-3` goes against our naming conventions, it would have to be named something that makes it unique, but:
- there just isn't enough difference to set this apart from `share`, at most we could consider giving `share` an update in style, but that would require a little "market research" on what most of out users would prefer

oosawy:
Thanks for the feedback!

First, I fixed the pixel misalignment since I needed it for my own use as well.

I agree that there isn’t enough difference to justify a separate icon from `share`. And I'm also not sure if updating the style of the existing icon makes sense.

I can’t do an actual market research, but [I looked into other icon libraries](https://react-icons.github.io/react-icons/search/#q=share) for reference. For share icons with a square and an upward arrow, most of them tend to have a closed square. Examples include `IoIosShare`, `IoShare`, `MdIosShare`, `GoShare`, `TbShare2`, and `RxShare2`. On the other hand, some like `RiShare2Line`, `FiShare`, and our `LuShare` are open square.

That said, I still feel like it’s a pretty minor difference. Since my use case (PWA install instruction) is quite specific, I’m not sure if changing the `share` style is necessary. I do like that the standard variant stays neutral and isn’t overly tied to Apple’s design.

To make this PR easier to find, I renamed `share-3` to `share-ios`, based on the naming conventions used in some icon libraries (though I’m not sure if a brand name like this would be acceptable here).

karsa-mistmere:
> To make this PR easier to find, I renamed `share-3` to `share-ios`, based on the naming conventions used in some icon libraries (though I’m not sure if a brand name like this would be acceptable here).

iOS is a trademark, we cannot officially use it.

PR #2945, opened 2025-03-25
feat(icons): added `decimals-arrow-left` icon and `decimals-arrow-right`

AnnaSasDev:
- [x] New Icon

### Description
Added new `decimals-arrow-left` icon and `decimals-arrow-right` icons.

### Icon use case

In a mathemical context (excel clone, ecomonic reports, etc...), having a quick buton to change the amount of precision. 

### Concept

### Author, credits & license

### Naming

### Design

## Before Submitting


karsa-mistmere (changes requested):

karsa-mistmere (approved):

airstrike:
I just wanted to start by saying it's amazing that you all have added this PR. I love lucide and having proper icons for spreadsheet applications in the kit is amazing.

One thing to note is that the icons as originally proposed were slightly different. The idea is the top and bottom row denote "before" and "after", respectively.

Apologies for the Discord screenshot:

image

You'll notice the "add decimal" variant has the arrow on the top row going left, meaning clicking on the icon will move the 0 to the left, yielding one more zero

Conversely, the "remove decimal" variant takes us from two zeroes to one. The decimal separators are there to illustrate these actions pertains to decimals and not just zero for padding, for instance. An argument could be made we don't need the decimal separator so it can be more generic.

The currently proposed icons both have two zeroes on the top row, so it's harder to tell what action they perform.

Thanks again for all the work that goes into lucide and for the beautiful design on these two icons!

ericfennis:
@AnnaSasDev Why the zero before the arrow? Can't we make it simpler with just an arrow?
image

airstrike:
> @AnnaSasDev Why the zero before the arrow? Can't we make it simpler with just an arrow? image

This is actually a galaxy brain move. Most spreadsheet apps have very complex icons for this operation and it's always hard to tell which one is which. By making them "one zero" and "two zeroes" it's super clear which one has more zeroes and which one has fewer, which solves the issue quite elegantly given these two icons are ~always presented as a pair.

ericfennis (changes requested):
Apply changes from comment from @karsa-mistmere

ericfennis (approved):

PR #2950, opened 2025-03-26
fix(@lucide/svelte): Add `.js` extensions to imports

### LINKED ISSUE #2756: Can't use Lucide Svelte with moduleResolution to "NodeNext"
### Package

- [ ] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [ ] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [x] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

0.474.0

- [x] Yes
- [ ] No

### Browser

- [ ] Chrome/Chromium
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [x] Other/not relevant

### Operating system

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [ ] Other/not relevant

### Description

Hello,

We are trying to migrate our app to Svelte 5, when creating a new sample application, SvelteKit generates a project with the ModuleResolution sets to "NodeNext" by default. From the documentation it looks like this is the new default and standardized way to do this. However, when I try to use Lucide Svelte, it's no longer possible to import the icons like this (which still works well with "moduleResolution": "node") :
`import { ArrowDownIcon } from 'lucide-svelte';`

I get an error saying that :
`Module '"lucide-svelte"' has no exported member 'ArrowDownIcon'.ts(2305)`

I guess this has something to do with the exports line in the package.json but I'm not sure.

### Steps to reproduce

1. npx sv create my-app
2. select typescript and library mode
3. pnpm add lucide-svelte@latest
4. import { ArrowDownIcon } from 'lucide-svelte';

Image

### Checklist

abdel-17:
closes #2756 

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

If you're using `moduleResolution: NodeNext` in the tsconfig of a Svelte project, TS doesn't recognize the types for the index import, so you get type errors for this:
```svelte
<script lang="ts">
    import { ArrowDownIcon } from "@lucide/svelte";
</script>
```

But not this:
```svelte
<script lang="ts">
    import ArrowDownIcon from "@lucide/svelte/icons/arrow-down";
</script>
```

This is because the package doesn't have file extensions in its imports. This PR fixes that by adding extensions to all imports and using `moduleResolution: NodeNext` to avoid this issue in the future.

## Before Submitting 


abdel-17:
I just realized the original issue is about the `lucide-svelte` package, but my PR fixes the issue for `@lucide/svelte`. Sorry.

ericfennis (approved):

ericfennis:
Sorry for the late review.
Nice, and thanks for the fix.
This still feels weird that we have to do it like this. We use Typescript files, and this syntax doesn't make sense 🤦 . 

Feels like something the Svelte compiler should fix.

PR #2995, opened 2025-04-05
fix(icons): changed `bed` icon

jguddas:
- [x] Other: Icon update

### Description
Moved bedding line and added 1px corner.

## Before Submitting


karsa-mistmere:
I'm fine with moving the line, but thumbs down for the curve, it makes the icon design inconsistent at different scales.

jguddas:
> I'm fine with moving the line, but thumbs down for the curve, it makes the icon design inconsistent at different scales.

Maybe we should even put that into the guidelines, or can you come up with a situation where you would want a corner there.

I remember we had a similar discussion when we did the house icon redesign.

PR #3011, opened 2025-04-09
fix(icons): redesigned `brush` icon

jguddas:
- [x] Other: Icon update

### Description
Redesign.

## Before Submitting


ericfennis:
The current brush icon was designed to match with the `pencil` icon to match nicely when used together.
image

Are we okay with this change to have a different stick design?

jguddas:
> The current brush icon was designed to match with the `pencil` icon to match nicely when used together. image
> 
> Are we okay with this change to have a different stick design?

I think it looks even better in context here.

ericfennis:
@karsa-mistmere You agree?

karsa-mistmere:
> @karsa-mistmere You agree?

I'm kind of on the fence here, the handle does feel a bit too narrow in comparison, how about something like this?

<a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=%3Cpath+d=%22M11+17+21.378+5.624a1+1+0+0+0-3.002-3.002L7+13%22+/%3E%3Cpath+d=%22m11.5+9+3.5+3.5%22+/%3E%3Cpath+d=%22M3+17a4+4+0+1+1+4+4H3a1+1+0+0+1-.708-1.708C3+18.5+3+18+3+17%22+/%3E&name=hand&base=hand">icons<br/>Open lucide studio</a>

image

karsa-mistmere:
@ericfennis: What about an oval shape then? I think the core issue is that brushes are pretty much never perfectly straight like pencils are.

image

jguddas:
> @ericfennis: What about an oval shape then? I think the core issue is that brushes are pretty much never perfectly straight like pencils are.
> 
> image

I think the version in this PR is more lucide-y.

ericfennis:
@jguddas I agree, I think we should go with the one from the pr.

ericfennis (approved):

jguddas:
Press the button @karsa-mistmere 🚀

PR #3016, opened 2025-04-10
feat(ci): added npm package provenance attestation

jguddas:

- [ ] New Icon
- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

> You can generate provenance statements for the packages you publish. This allows you to publicly establish where a package was built and who published a package, which can increase supply-chain security for your packages.

https://docs.npmjs.com/generating-provenance-statements

## Before Submitting 

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,6 +38,8 @@ jobs:
     if: github.repository == 'lucide-icons/lucide'
     runs-on: ubuntu-latest
     needs: pre-release
+    permissions:
+      id-token: write
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -80,11 +82,15 @@ jobs:
 
       - name: Publish
         run: pnpm --filter ${{ matrix.package }} publish --access public --no-git-checks --ignore-scripts
```
ericfennis, line 84: According to the link in the PR description, we also need to add the `--provenance` flag here

.github/workflows/release.yml
```diff
@@ -111,6 +117,8 @@ jobs:
 
       - name: Publish
         run: pnpm --filter lucide-static publish --no-git-checks
```
ericfennis, line 119: Here as well

ericfennis (commented):
Nice addition. I'm curious if we can see this in NPM as well.

.github/workflows/release.yml
jguddas: We use pnpm, the link is only describing how to do it with non, pnpm uses the env instead of the argument.

ericfennis: Ah, that makes sense. I thought pnpm would proxy the flags.

ericfennis (approved):

PR #1982, opened 2024-03-14
feat(icons): adds microchip icon

karsa-mistmere:
## What is the purpose of this pull request?

- [x] New Icon

### Description
We do not currently have an icon that could represent GPUs, this PR adds a `microchip` icon that could stand in for either GPUs or RAM.

### Icon use case
- gpu
- ram
- microchip

### Alternative icon designs 
image

### Concept 

- [x] I have provided valid use cases for each icon.
- [x] I have not added any a brand or logo icon.
- [x] I have not used any hate symbols.
- [x] I have not included any religious or political imagery.

### Author, credits & license

- [x] I've based them on the following Lucide icons: `cpu`

### Naming 
- [x] I've read and followed the [naming conventions](https://lucide.dev/guide/design/icon-design-guide#naming-conventions)
- [x] I've named icons by what they are rather than their use case.
- [x] I've provided meta JSON files in `icons/[iconName].json`.

### Design 

- [x] I've read and followed the [icon design guidelines](https://lucide.dev/guide/design/icon-design-guide)
- [x] I've made sure that the icons look sharp on low DPI displays.
- [x] I've made sure that the icons look consistent with the icon set in size, optical volume and density.
- [x] I've made sure that the icons are visually centered.
- [x] I've correctly optimized all icons to two points of precision.

## Before Submitting 

- [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [x] I've checked if there was an existing PR that solves the same issue.

PR #3018, opened 2025-04-11
fix(icons): changed `microchip` icon

karsa-mistmere:
The design provided was already explored in #1982 , if the core layout is good enough for Material Symbols, it should be good enough for us and it's much cleaner than our current version:

image

- [x] Other: Icon update

### Description
Revamps the `microchip` icon to have better legibility.

## Before Submitting


ericfennis (approved):

jguddas (changes requested):
I prefer the previous version, the lines don't make it feel more like a chip to me.
And the reduction of the pins also feels wrong, a chip usually has pins that are closeish together.

karsa-mistmere:
> I prefer the previous version, the lines don't make it feel more like a chip to me. And the reduction of the pins also feels wrong, a chip usually has pins that are closeish together.

@jguddas: Dunno, I think the design in this PR is much cleaner than what we currently have, and CPUs have ever more pins than a regular microchip, yet our CPU icon has barely a few and is still perfectly legible.

PR #3031, opened 2025-04-13
fix(icons): changed `heading-6` icon

jguddas:
- [x] Other: Icon update

### Description
Made 6 more monospace looking

## Before Submitting


ericfennis:
If we change this for this, should we not make them all mono-spaced?

PR #3032, opened 2025-04-13
fix(icons): changed `section` icon

jguddas:
- [x] Other: Icon update

### Description
Redesigned icon.

## Before Submitting


karsa-mistmere:
What't the reasoning behind the redesign? Personally, I think I prefer our current design.

PR #3105, opened 2025-04-21
fix(icons): rotate `blend` icon

jguddas:
- [x] Other: Icon update

### Description
Rotate icon.

## Before Submitting


jguddas:
Not 100% sure.

jguddas:
I added some more tags, since this icon can also be used to symbolize marriage.
This icon is wrongly named by use case :/

karsa-mistmere:
> I added some more tags, since this icon can also be used to symbolize marriage. This icon is wrongly named by use case :/

Yeah, maybe let's rename this `circles-interlocked`? 🤔

I'm fine with the flipping too.

PR #3106, opened 2025-04-21
fix(icons): rotate `citrus` icon

jguddas:
- [x] Other: Icon update

### Description
Rotate icon.

## Before Submitting


karsa-mistmere:
There doesn't seem to be a recommendation about orientation here: https://emojipedia.org/lime#designs

But truth be told, I prefer the current one. 🤷‍♂️

ericfennis:
Let's keep the current one

PR #3108, opened 2025-04-21
fix(icons): rotate `pizza` icon

jguddas:
- [x] Other: Icon update

### Description
Rotated icon.

## Before Submitting


jguddas:
Feels pretty good when rotated.

karsa-mistmere:
> Feels pretty good when rotated.

IMO this should follow 🍕, most mainstream designs use our current orientation: https://emojipedia.org/pizza#designs

ericfennis:
So flip it?

jguddas:
> > Feels pretty good when rotated.
> 
> IMO this should follow 🍕, most mainstream designs use our current orientation: https://emojipedia.org/pizza#designs

In that list we have:

from top-left: 9
from top: 5
from top-right: 11
full pie: 1
from bottom: 1

Also comparing it wit other icon sets there is no real common pattern:

https://icones.js.org/collection/all?s=pizza

PR #3109, opened 2025-04-21
fix(icons): rotate `popsicle` icon

jguddas:
- [x] Other: Icon update

### Description
Rotate icon.

## Before Submitting


karsa-mistmere:
> Not 100% sold on this.

🍡🍭

So, me neither.

ericfennis:
hmm

jguddas:
billede

https://icones.js.org/collection/all?s=popsicle

ericfennis:
image

We have a lot of tools an aligned to from bottom left to top right. So maybe this should as well.

🤔

ericfennis:
Let's keep it as it is

PR #3111, opened 2025-04-21
fix(icons): rotate `key` icon

jguddas:
- [x] Other: Icon update

### Description
Rotate icon.

## Before Submitting


karsa-mistmere:
We should also explore this orientation in various `-key` icons.

karsa-mistmere (approved):
> We should also explore this orientation in various `-key` icons.

OK, this is now outdated, since we've changed the `-key` modifier.

PR #3115, opened 2025-04-22
fix(icons): changed `guitar` icon

jamiemlaw:
- [x] Other: Icon update

### Description
Converted the body of the guitar to use arcs instead of cubic Béziers.

## Before Submitting


jguddas (approved):

jamiemlaw:
> I liked it better when you accidentally removed the hole.

In that case I'll re-remove it! :-)

But I also agree: if an icon doesn't _need_ a filled circle and can get by with only stroked paths, all the better.

ericfennis (approved):
Niceee

PR #3141, opened 2025-04-26
fix(icons): removed `trash` icon in favour of `trash-2`

jguddas:
- [x] Other: Icon update

### Description
Removed `trash` icon in favour of `trash-2`.

`trash-2` has the same usecase and is around 3 times as popular as `trash`.

## Before Submitting


ericfennis:
I agree with your reasons. But this little change is changing the look of some apps out there. So I'm a little cautious changing this without some input from others. So we need some more opinions on this.

jguddas:
> I agree with your reasons. But this little change is changing the look of some apps out there. So I'm a little cautious changing this without some input from others. So we need some more opinions on this.

Same was the case when we changed brush 🖌️.

I'm quite happy to not be too quick with changes like this and give people ti.e to provide input.

karsa-mistmere (approved):
TBH, I'm fine with these changes. I find that the alternative variant is much more legible at the cost of very little extra complexity, but I refrained from chipping in, since I almost exclusively use the `trash-2` variant, so I'm not super sure that I'm as impartial as I should be. :)

jguddas:
@ericfennis can you press the merge button, pretty please.

ericfennis:
For these changes I want to do a bit of more user research. Maybe do a code search first?

karsa-mistmere:
> For these changes I want to do a bit of more user research. Maybe do a code search first?

@ericfennis: I’ve found the following usage for these icons across public GitHub repositories:

```
lucide-react
  trash: 106 repositories, 102 exclusive (8.1%)
  trash-2: 1,150 repositories, 1,146 exclusive (91.5%)
  combinations:
    trash-2 only: 1,146 (91.5%)
    trash only: 102 (8.1%)
    both: 4 (0.3%)
  total repositories: 1,252
  repositories sampled/searched: 1,903
```

Based on this, I think there would be very little consequence to merging `trash-2` into `trash`, using the `trash-2` design for the resulting `trash` icon, as @jguddas suggested.

The `trash-2` design also seems much closer to what people generally expect from a generic trash icon, based on a Google image search:
image

ericfennis (approved):

PR #3157, opened 2025-04-27
fix(icons): changed `square-stack` related icons

jguddas:
- [x] Other: Icon update

## Before Submitting


icons/bring-to-front.svg
```diff
@@ -9,7 +9,8 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect x="8" y="8" width="8" height="8" rx="2" />
-  <path d="M4 10a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h4a2 2 0 0 1 2 2" />
-  <path d="M14 20a2 2 0 0 0 2 2h4a2 2 0 0 0 2-2v-4a2 2 0 0 0-2-2" />
+  <path d="M16 15a1 1 0 0 1-1 1" />
+  <path d="M8 9a1 1 0 0 1 1-1" />
+  <rect x="13" y="3" width="8" height="8" rx="1" />
+  <rect x="3" y="13" width="8" height="8" rx="1" />
```
jguddas:
```suggestion
  <path d="M11 20a1 1 0 0 1-1 1H4a1 1 0 0 1-1-1v-6a1 1 0 0 1 1-1" />
  <path d="M13 4a1 1 0 0 1 1-1h6a1 1 0 0 1 1 1v6a1 1 0 0 1-1 1" />
  <rect x="8" y="8" width="8" height="8" rx="1" />
```

ericfennis:
Can you explain more why this change is needed?

karsa-mistmere:
@ericfennis: I think it's about perceived design guideline violations (optical volume in particular), I do recall @jguddas opening another PR about decreasing the optical volume of `copy` around this time, in a manner similar to what's seen below, but I can't seem to find this PR:

image

IMO these should definitely be consistent with [`copy`](https://lucide.dev/icons/copy), so I'm voting nay on these changes.

jguddas:
It's so the border radius and gaps are following the 8px rule.

karsa-mistmere:
> It's so the border radius and gaps are following the 8px rule.

Um... but these squares are fine because they're 8px large and the guidelines actually state
- ≥8px => 2px radius
- <8px => 1px radius

😅

PR #3164, opened 2025-04-28
feat(website): propose new color picker component

jpjacobpadilla:

- [x] Other: A nicer looking color picker for the Lucide website

### Description

I typically use Safari, which doesn’t have a very nice color picker popup, so I decided to design one (based on an existing Vue component) that fits better with Lucide. The proposed color picker is built on a modified version of the Chrome one from the `@ckpack/vue-color` package. 

Proposed design on icons page:

Proposed design on home page:

Default in Safari:

Default in Chrome:

## Before Submitting 


jguddas:
What kind of network traffic and performance impact does this have?

jpjacobpadilla:
According to the performance tab in Chrome (at least on my computer), I don't see any significant change in the LCP or any other metrics.

PR #3198, opened 2025-05-06
Add 9 alignment position icons (header/center/footer grid)

christiangoran:
## Description
This PR adds 9 new icons representing alignment positions in a 3x3 grid layout:

- upper-left
- upper-center
- upper-right
- middle-left
- middle-center
- middle-right
- lower-left
- lower-center
- lower-right

These icons are designed to be used in scenarios where precise header or footer watermark positioning is required in a document editor UI.

### Icon use case
- Document watermark alignment controls (e.g. top-left, bottom-center)
- Text or image alignment tools in design, layout, or markdown editors
- Print and layout configuration panels for PDF generation

### Alternative icon designs
No alternative designs — all icons are symmetrical variants in a unified style.

### Concept

### Author, credits & license
- [x] I've based them on the following Lucide icons: `align-left`, `align-center`, `align-right`

### Naming

### Design

## Before Submitting


karsa-mistmere:
These icons do not follow [our naming conventions](https://lucide.dev/guide/design/icon-design-guide#naming-conventions), please propose alternative names for these that do.

christiangoran:
Updated icon names as requested — let me know if you prefer another naming convention.

jguddas:
Use case feels weak, also not really coherent with the current `align-*` icons.

christiangoran:
Thanks for the feedback!

To better showcase the use case and how these icons are used in practice, I recorded a short video demo of them in a real application. Hopefully this makes the purpose and design more clear:

https://github.com/user-attachments/assets/dbe019df-e69b-4c97-abc2-6ea1c4a54da7

Let me know if you’d like any further adjustments to the icons or naming.

Screenshot 2025-05-06 at 21 23 18

jguddas:
Fair enough, that use case is actually not so special, here are some examples in the wild:

billede
billede

billede
billede
billede
billede
billede

christiangoran:
Thanks! There definitely seem to be a use case besides mine then ;)   
I agree the powerpoint ones are slick with only two lines.

ericfennis (commented):
The usecase is still valid. Other alignment icons don't feel right to use in this case.
So I'm in for adding these. But let's name them in this pattern: `text-align-*`. So they match with the other `text-align` icons

karsa-mistmere:
I think the use case itself is valid, but the proposed designs fall short of clearly conveying their intended meaning.

1) In most UIs, vertical and horizontal alignment are treated as separate concerns. Figma is a good example of this, and it’s a pattern that would translate well to Lucide (I could have sworn someone already opened a PR along these lines, but I can’t seem to find it right now):
image


2) If the goal is to represent both vertical and horizontal alignment in a single icon, simply varying line lengths isn’t very self-explanatory. The icon needs additional context to communicate what the content is being aligned to. That likely means introducing a bounding shape or frame, for example:

image

Issue #2154, opened 2024-05-15
Drone icon


bernatfortet:
### Icon name

Drone

### Use cases

When using map layers showing drone footage, or for interfaces that work with drones (pathing, recording, etc.)

### Design ideas

Here's an example https://fontawesome.com/icons/drone-front?f=classic&s=regular

Or this one from https://atlasicons.vectopus.com/
![drone](https://github.com/lucide-icons/lucide/assets/390010/d6722468-07d8-451b-ba60-11b957ede6c9)

### Checklist

- [X] I have searched if someone has submitted a similar issue before and there weren't any. (Please make sure to also search closed issues, as this issue might already have been resolved.)
- [X] I have searched existing icons to make sure it does not already exist and I didn't find any.
- [X] I am not requesting a brand logo and the art is not protected by copyright.
- [X] I am not requesting an icon that includes religious, political imagery or hate symbols.
- [X] I have provided appropriate use cases for the icon(s) requested.

bernatfortet:
I just came up with this one:
![FRAME-NAME svg](https://github.com/lucide-icons/lucide/assets/390010/da442198-1693-4d45-8da4-5dc1613487af)

bernatfortet:
<img width="104" alt="image" src="https://github.com/lucide-icons/lucide/assets/390010/1eb775f5-e771-42ef-b3f5-e0d4d66cac31">
Another idea

ericfennis:
@bernatfortet Feel free to create a PR

PR #3214, opened 2025-05-11
feat(icons): added `drone` icon

shopped:
Opening a PR for the drone icon that [bernatfortet](https://github.com/bernatfortet) made according to [this issue](https://github.com/lucide-icons/lucide/issues/2154).

Does anyone want to help me double check that I got everything right?


karsa-mistmere:
Hi @shopped,

While I do admit this icon is very cute, it is unfortunately nowhere near Lucide-compliant. :( 

I'd like to ask you to double check [our icon design guidelines](https://lucide.dev/guide/design/icon-design-guide), as this design violates the following rules:
- it has higher than optimal visual weight; since it's rectangular in shape, it should aim to be roughly the same size as rectangle.
- it does not have the necessary 2px gap between its elements.
- on the other hand, the arms do not meet the chassis and leave a tiny gap that will become visible should the icon's stroke width or size be customised.
- it's also not very optimised, the propellers are made up of tiny cubic path segments, even though the could very well be a single arc, you can use https://studio.lucide.dev/ to optimise your design before submitting it

shopped:
Hey @karsa-mistmere

Thanks for taking the time to look this over and reply, the feedback is very helpful. When I have the time I will redo the icon based on the style guidelines, and make sure to read all the contribution docs. Thank you so much again, I will credit you when I revisit!

PR #3267, opened 2025-05-29
fix(icons): changed `blocks` icon

jguddas:
- [x] Other: Icon update

### Description
Increased size of `blocks` icon.

## Before Submitting


ericfennis:
I think I like the alternative more. Feels weird that it's not fully a square 🤔

PR #3282, opened 2025-06-06
perf(docs): optimize IconsOverview initial render - 50% faster load times

### LINKED ISSUE #3203: perf: performance issues on icons page
### Browser

- [x] Chrome/Chromium
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [ ] Other/not relevant

### Operating system

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [ ] Other/not relevant

### Description

There is some performance issue when we click navigate to the icons page, it takes 2-3 seconds to load the page.
I’d like to work on this and submit a pull request to fix it.

Would it be okay if I proceed with a PR for this?
Thanks!

### Steps to reproduce

https://github.com/user-attachments/assets/347585f2-fcb9-42c7-a653-96c8d2e864e2

### Checklist

epifaniofrancisco:
# Performance Optimization: IconsOverview Initial Render

- [ ] New Icon
- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description
This pull request fixes the performance issue where IconsOverview was loading 1600 icons on initial render, bypassing virtual scrolling and causing slow page load times. The Icons page now loads 50%+ faster. Fixes #3203

#### Changes Made:

**IconsOverview.vue**
- Added dynamic `initialGridItems` computed property that calculates based on viewport size
- Reduced initial render from 1600 to 120-200 icons (87% reduction)
- Changed `splice()` to `slice()` to avoid array mutation and improve performance
- Fixed NoResults condition to use `searchResults.length` instead of `list.length`

#### Performance Impact:
- **Before**: 7-13 seconds load time, 1600 icons rendered initially
- **After**: 3-6 seconds load time, ~120-200 icons rendered initially  
- **Result**: 50%+ faster loading, 87% reduction in initial DOM nodes

## Before Submitting

### DIFF
--- a/docs/.vitepress/theme/components/icons/IconsOverview.vue
+++ b/docs/.vitepress/theme/components/icons/IconsOverview.vue
@@ -14,7 +14,15 @@ import CarbonAdOverlay from './CarbonAdOverlay.vue';
 
 const ICON_SIZE = 56;
 const ICON_GRID_GAP = 8;
-const DEFAULT_GRID_ITEMS = 10 * 160;
+
+const initialGridItems = computed(() => {
+  if (containerWidth.value === 0) return 120;
+  
+  const itemsPerRow = columnSize.value || 10;
+  const visibleRows = Math.ceil(window.innerHeight / (ICON_SIZE + ICON_GRID_GAP));
+
+  return Math.min(itemsPerRow * (visibleRows + 2), 200); 
+});
 
 const props = defineProps<{
   icons: IconEntity[];
@@ -115,15 +123,14 @@ function handleCloseDrawer() {
       />
     </StickyBar>
     <NoResults
-      v-if="list.length === 0 && searchQuery !== ''"
+      v-if="searchResults.length === 0 && searchQuery !== ''"
       :searchQuery="searchQuery"
       @clear="searchQuery = ''"
     />
     <IconGrid
       v-else-if="list.length === 0"
-      :key="index"
       overlayMode
-      :icons="[...searchResults].splice(0, DEFAULT_GRID_ITEMS)"
+      :icons="searchResults.slice(0, initialGridItems)"
       :activeIcon="activeIconName"
       @setActiveIcon="setActiveIconName"
     />


jguddas:
Do you know what causes it to be slow, is it the javascript, is it the svgs?

epifaniofrancisco:
> Do you know what causes it to be slow, is it the javascript, is it the svgs?

The Icons page slowdown is caused by **bypassing virtual scrolling during initial load**, resulting in massive DOM rendering. So when the page first loads, `list.length === 0` (virtual list hasn't initialized yet), which triggers a fallback that renders **all 1,600 icons at once**:

```vue
<IconGrid
  v-else-if="list.length === 0"  // Triggers on initial load
  :icons="searchResults.slice(0, DEFAULT_GRID_ITEMS)" // Was 1600 icons
/>
```

Virtual scrolling was designed to render only ~64 visible items, but the fallback was defeating this optimization entirely.

jguddas:
> > Do you know what causes it to be slow, is it the javascript, is it the svgs?
> 
> The Icons page slowdown is caused by **bypassing virtual scrolling during initial load**, resulting in massive DOM rendering. So when the page first loads, `list.length === 0` (virtual list hasn't initialized yet), which triggers a fallback that renders **all 1,600 icons at once**:
> 
> ```vue
> <IconGrid
>   v-else-if="list.length === 0"  // Triggers on initial load
>   :icons="searchResults.slice(0, DEFAULT_GRID_ITEMS)" // Was 1600 icons
> />
> ```
> 
> Virtual scrolling was designed to render only ~64 visible items, but the fallback was defeating this optimization entirely.

Yes, but why is it slow to render 1600 icons, is it js, or just dom load?

jguddas:
If you go to https://lucide.dev/icons, press <kbd>ctrl+s</kbd> and safe the HTML to disc, and then open the saved file in your browser, how quick / slow is that?

epifaniofrancisco:
> If you go to https://lucide.dev/icons, press ctrl+s and safe the HTML to disc, and then open the saved file in your browser, how quick / slow is that?

I tested the static HTML performance as suggested. The static HTML loads instantly, which proves the bottleneck is JavaScript execution, not DOM/SVG rendering.

The real performance killer is **1,600 Vue component instances** being created simultaneously:

```javascript
// Each icon = full Vue component with:
// - Reactive data system
// - Event listeners  
// - Virtual DOM reconciliation
// - Lifecycle hooks
// - Memory allocation
```

jguddas:
It has to be possible to SSG https://lucide.dev/icons @ericfennis?

ericfennis:
Yess, this was a known issue. We are doing SSG, but the during SSG the list we initially rendering is quite large and not based on screen size. This fixes that part.
So thank you for this change!

PR #3283, opened 2025-06-06
fix(icons): changed `computer` icon

jguddas:
- [x] Other: Icon update

### Description
Added line to make it look more like a monitor.

## Before Submitting


karsa-mistmere:
Doesn't feel that much better for me personally. I still think the aspect is wrong, and the icon is too large overall, wish it never got merged in the first place. 😅

PR #3285, opened 2025-06-07
Update lucide.md

D-yo-r:
fix(docs): Updated URL of unpkg

## Description

The URL of unpkg requires `www`

## Before Submitting 


jguddas:
`www.` is not needed.

D-yo-r:
Yes, the problem seems to have been fixed.

PR #3300, opened 2025-06-10
chore(metadata): added tag to `bandage` icon

jamiemlaw:
Added `patch` as a tag to `bandage` icon

Seen on an A/B testing app where tests can be applied as patches. Seemed like an appropriate keyword to add.

image


ericfennis:
Thanks @jamiemlaw !

PR #3337, opened 2025-06-18
Rename Vue package name to `@lucide/vue`

ericfennis:
# Description

Creates new Vue package: `@lucide/vue`


jguddas:
Can we somehow still publish to the old names for backwards compatibility?

ericfennis:
@jguddas Yes, we still do! Maybe a bit confusing in this PR. But I've deleted the old `vue 2` package, which was already not publishing since the beginning of this year.
So `lucide-vue-next` is still publishing!

ThomOrlo:
Thank you for your work on this! I've had a look and there are just a few minor tweaks needed on some of the documentation files. I've noted a few examples with links below in case you haven't already spotted them:
- https://github.com/lucide-icons/lucide/blob/main/docs/.vitepress/sidebar.ts#L77
- https://github.com/lucide-icons/lucide/blob/main/docs/.vitepress/theme/components/home/HomePackagesSection.data.ts#L16
- https://github.com/lucide-icons/lucide/blob/main/docs/.vitepress/lib/codeExamples/createCodeExamples.ts#L40

ericfennis:
Thanks for you sharpness @ThomOrlo . Will fix this!

PR #3341, opened 2025-06-20
chore(metadata): Add tags to `x` icon

jamiemlaw:
## Description

Added 'ex' and 'cross' as tags

## Before Submitting 


ericfennis (approved):
Nice tnx!

PR #3344, opened 2025-06-20
fix(icons): redesigned `vibrate` and `vibrate-off` icon

jguddas:
- [x] Other: Icon update

### Description
Redesign based on `smartphone`.

Generally the current `vibrate` icon is just a bit ugly and small.

## Before Submitting

### DIFF
--- a/icons/vibrate-off.json
+++ b/icons/vibrate-off.json
@@ -1,8 +1,11 @@
 {
   "$schema": "../icon.schema.json",
   "contributors": [
+    "colebemis",
+    "mittalyashu",
     "lscheibel",
-    "ericfennis"
+    "ericfennis",
+    "jguddas"
   ],
   "tags": [
     "smartphone",


icons/vibrate.svg
karsa-mistmere, line 1: Feels a bit more like a circuit plan than a vibrating smartphone. 🤔

image

I could maybe see this design working with some negative space:

image

But all in all, I think I prefer the current one.

ericfennis:
I like the shape is more like the smartphone icon.
But on 100% scale the vibrate shapes are not that obvious to me. I'm also not sold to this one.

jguddas:
> I like the shape is more like the smartphone icon. But on 100% scale the vibrate shapes are not that obvious to me. I'm also not sold to this one.

It's as sharp as it gets.

image

karsa-mistmere:
> > I like the shape is more like the smartphone icon. But on 100% scale the vibrate shapes are not that obvious to me. I'm also not sold to this one.
> 
> It's as sharp as it gets.
> 
> image

Yeah, I'm sorry, but this feels awful. 🙈

ericfennis:
An icon should be clear and recognizable to be useful. This is more important than consistency IMO.
So let's keep the original one for now.

PR #3363, opened 2025-06-25
ci(token-permissions): added top-level permissions to gh actions

dsm23:
* When using dependency review on my own repos that use lucide icons, the scorecard result is coming out a little low
* As of the last time I checked the score was 2.9/10 and the low-barrier for these dependency checks tends to be 3/10
* This PR attemps to increase the OpenSSF scorecard result
* Following advice from the OpenSSF documentation on token-permissions
* [https://github.com/ossf/scorecard/blob/ab2f6e92482462fe66246d9e32f642855a691dc1/docs/checks.md#token-permissions](https://github.com/ossf/scorecard/blob/ab2f6e92482462fe66246d9e32f642855a691dc1/docs/checks.md#token-permissions)

## Before Submitting

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,17 @@ on:
     paths:
       - icons/**/*.svg
 
+permissions: read-all
+
 jobs:
   create-release:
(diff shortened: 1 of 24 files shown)


jguddas (changes requested):
This needlessly increases the permission scope.

`permissions: read-all` should not be added.

PR #3381, opened 2025-07-01
fix(icons): changed `brain` icon

jamiemlaw:
- [x] Other: Icon update

### Description
Optimised paths to reduce number of segments, and added some extra tags

## Before Submitting


jguddas (approved):

ericfennis (approved):
Nice!

luke-layerhealth:
@ericfennis is there a strong reason we need this change? I understand the desire for optimization but the previous icon supported fill and the new icon does not.

PR #3397, opened 2025-07-08
docs: add deep import recommendation for lucide-react

youknowom:
## Description

📄 This pull request adds a section to the `README.md` file recommending **deep imports** for `lucide-react`.

✅ Example added:
```js
// ✅ Recommended
import VideoIcon from 'lucide-react/icons/video';

// ❌ Not recommended
import { VideoIcon } from 'lucide-react';

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,3 +86,16 @@ Thank you to all the people who contributed to Lucide!
 
 <a href="https://www.scipress.io?utm_source=lucide"><img src="docs/public/sponsors/scipress.svg" width="180" alt="Scipress sponsor badge" /></a>
 <a href="https://github.com/pdfme/pdfme"><img src="docs/public/sponsors/pdfme.svg" width="180" alt="pdfme sponsor badge" /></a>
+ 
+ ---
+
+## ⚠️ Importing Icons in React
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by youknowom** [comment]
Thanks @jguddas for the feedback! I’ve added the bundler clarification as suggested. Let me know if anything else needs updating

File: README.md [hunk 1]
```diff
@@ -86,3 +86,21 @@ Thank you to all the people who contributed to Lucide!
 
 <a href="https://www.scipress.io?utm_source=lucide"><img src="docs/public/sponsors/scipress.svg" width="180" alt="Scipress sponsor badge" /></a>
 <a href="https://github.com/pdfme/pdfme"><img src="docs/public/sponsors/pdfme.svg" width="180" alt="pdfme sponsor badge" /></a>
+ 
+ ---
+
+## ⚠️ Importing Icons in React
+
+To keep your bundle size optimized, we recommend importing icons using **deep imports**. This ensures only the icons you use are included in your final build.
```
jguddas, line 94: This is also not what I would recommend, I would recommend you use a bundler / plugin to tree shake your barrel imports.

That's why the text "we recommend" doesn't make me particularly happy, direct imports are a hack that I would prefer people would avoid.

PR #3401, opened 2025-07-09
feat(icons): added `check-upright` and `check-check-upright` icon

craigfrancies:
- [x] New Icon

### Description
Added new `check-upright` and `check-check-upright` icons.

These are alternative designs to the existing `check` and `check-check` icons, drawn at a slightly more upright angle.

The main reason for the change is that the second check in `check-check` can appear quite small and hard to see, especially at smaller sizes. I started with the double check to make better use of the available space. From there, it made sense to create a matching single check with the same slant, so they feel consistent as a pair.

_Comparison to `check` + `check-check`_

image

### Icon use case

- Commonly used in messaging apps to show that a message has been delivered and read by the recipient.
- Can be used in notification systems to show that an item has been acknowledged by the user.
- Used in collaborative tools to indicate that multiple steps of a task have been completed or verified.

### Alternative icon designs

_Not explored_

### Concept

### Author, credits & license

- [x] The icons were originally created in #<issueNumber> by @craigfrancies
- [x] I've based them on the following Lucide icons: `check`, `check-check`

### Naming

### Design

## Before Submitting

### DIFF
--- a/icons/check-upright.json
+++ b/icons/check-upright.json
@@ -1,0 +1,17 @@
+{
+  "$schema": "../icon.schema.json",
+  "contributors": [
+    "craigfrancies"
+  ],
+  "tags": [
+    "check",
+    "tick",
+    "confirm",
+    "done",
+    "approved",
+    "yes"
+  ],
+  "categories": [
+    "notifications"
+  ]
+}


jguddas:
> These are alternative designs to the existing check and check-check icons, drawn at a slightly more upright angle.

We do not accept icons without unique use cases, i.e. no icons that are just alternative designs.

craigfrancies:
> > These are alternative designs to the existing check and check-check icons, drawn at a slightly more upright angle.
> 
> We do not accept icons without unique use cases, i.e. no icons that are just alternative designs.

OK, no problem.

Just to explain a bit more - this was mainly about improving the double check for legibility at small sizes, especially in messaging or notification UIs where it's used to show message delivered or read. The current one can feel quite cramped and muddy.

If naming is the issue, we could go with something like `delivered`. That said, it does have a more upright slant, so it might look a bit out of place next to the existing check – which is why a matching version was included.

PR #3419, opened 2025-07-20
fix(icons): changed `watch` icon

jguddas:
- [x] Other: Icon update

### Description
Rated 45° to increase size.

## Before Submitting


karsa-mistmere:
I'm not a fan, it's a very unusual orientation for a watch icon, also correlating with: ⌚

Rotating the face also makes it feel ever so slightly less awkward, but I'm still voting to close this.

image

PR #3420, opened 2025-07-20
chore(icons): moved `wand` to `wand-sparkels` icon

jguddas:
## Description

They share the same use case.

## Before Submitting 


karsa-mistmere:
> They share the same use case.

I really don't think they do. While there's certainly some overlap, each icon has distinct use cases that aren't trivially interchangeable.

The fairy wand is typically used to represent _actual_ magic, as in fairy tales, games or fantasy: a unique use case in itself (like in gaming apps it could represent magic power or mana points). It is often associated with light-touch, one-click actions such as a quick selection or auto-enhance. The magician's wand usually wouldn't be used in this context.

The magician's wand, on the other hand, represents the _appearance_ of magic: it's an illusionist's accessory. It is more commonly associated with automated processes, generative AI, or multi-step wizards – things that simulate magical outcomes through technical means rather than literal magical metaphors.

PR #3248, opened 2025-05-22
feat(icons): Moved triangle right to be centered

gabebohlmann:
## Description
Triangle was not centered properly in square. The mistake came from the fact that the length from the center of an equilateral triangle to a vertex is not the same as the length from the center to the midpoint of an edge. 

## Before Submitting 

- [X] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [X] I've checked if there was an existing PR that solves the same issue.

Issue #2757, opened 2025-01-29
Stop icon to match play and pause icons


FionaHuangQC:
### Icon name

stop

### Use cases

- in any situation involving audio, it's typical to be able to play, pause, and stop. There are currently `play` and `pause` icons, but only `circle-stop` and `monitor-stop`. 
- to interrupt an ongoing process, a stop icon would be ideal

### Design ideas

![Image](https://github.com/user-attachments/assets/39798387-8ae9-4133-99bb-7d6d998399db)

![Image](https://github.com/user-attachments/assets/76b59dc6-86a3-4bd7-9dab-0eb967be4d2b)

### Checklist

- [x] I have searched if someone has submitted a similar issue before and there weren't any. (Please make sure to also search closed issues, as this issue might already have been resolved.)
- [x] I have searched existing icons to make sure it does not already exist and I didn't find any.
- [x] I am not requesting a brand logo and the art is not protected by copyright.
- [x] I am not requesting an icon that includes religious, political imagery or hate symbols.
- [x] I have provided appropriate use cases for the icon(s) requested.

jguddas:
There was a lot of discussions here https://github.com/lucide-icons/lucide/pull/1043

We also need to round the corners of all those icons to be more lucidy.
 <img src="https://github.com/user-attachments/assets/ca81ac4f-486b-485c-8245-6128a16a2916" width="50%" />

edwloef:
I'd love to see this happen as well

PR #3425, opened 2025-07-21
fix(icons): arcified playback icons

karsa-mistmere:
closes #3248 #2757 #3248 #1989

Also see #1043

## Description

It's been a long time coming, but here come the Lucide-compliant playback icons, with no gap violations, and `square` finally matching `play` so that it can be used standing in for `stop`.

image

## Alternative designs

image

## Before Submitting 

Sidenote: `play` is positioned as the best compromise that works for both circular and rectangular buttons:

image


jguddas (approved):

ericfennis (approved):
Nice @karsa-mistmere, Great work, finally fixed this one! 🎉

zoharma:
I have an issue with the square (used as stop) beacuse it looks too much like an empty checkbox, I think it may need to be slightly smaller than an actual square, and potentially with no rounded edges. Stop is a a unique function and could be less stylised. The circle-stop is better visualised as such, but I have a need to use both (e.g Stop a running queue, vs abort a current item). I also think that the othet icons could be slightly smaller

ericfennis:
Can you post an image or video to explain it a bit more?

PR #3428, opened 2025-07-21
fix(icons): changed `croissant` icon

karsa-mistmere:
- [x] Other: Icon update

### Description
redesigned croissant from scratch to be more minimalistic & clean

## Before Submitting


karsa-mistmere:
> I love it, feels more visually centered.
> 
> Maybe the issue isn't the rotation but that the icon wasn't really centered.

That, and the corners were super wonky too. :)

PR #3446, opened 2025-07-25
ci: auto-close issues about old versions

jguddas:
## Description

Closes issues from people who haven't tried the newest version.

## Before Submitting 


jguddas:
**Can you reproduce this in the latest version?**

Having issues finding/importing an icon? Ensure you are using a version that is new enough!
in the icon preview top right corner of lucide.com you can find the required version

jguddas:
Tested in https://github.com/jguddas/lucide/issues/106

ericfennis (approved):
Nice, I like it. I'm curious if this is not causing problems when we bump a lot, and at the same time, people try to create an issue.

But I think it's worth trying because it's annoying to constantly ask for it.

karsa-mistmere:
> Nice, I like it. I'm curious if this is not causing problems when we bump a lot, and at the same time, people try to create an issue.
> 
> But I think it's worth trying because it's annoying to constantly ask for it.

Yeah, I'm also a bit cautious about this. On one hand, issues like this are super annoying. On the other, I do wish the decision algorithm was more sophisticated, I think we should only auto-close issues about "missing" icons, not actual bug reports that weren't tested on the _latest_ version, since yeah, we do bump versions _a lot_.

karsa-mistmere:
I wish the solution was as simple as adding a disclaimer to the issue template that people should check the versions of icons before submitting a bug report about them. 😅

Alas, we know for a fact that people simple do not effing read these disclaimers. 😤

ericfennis:
I agree.
Maybe a safer idea that also defeats the goal: Just adding an automated comment on a newly opened issue with the old version. Something like: "I see you used an outdated version of Lucide. Have you tried updating it to the latest version? Does this issue still occur?".
People will probably check them selves first and doesn't get annoyed by automated closed issue.

jguddas:
We could also add a new issue type that is called "I can't find / import this icon".

PR #3455, opened 2025-07-27
fix(icons): changed `monitor-speaker` icon

jguddas:
- [x] Other: Icon update

### Description
Increased size of circle.

## Before Submitting


karsa-mistmere:
`speaker` is 1000% better, but I'm not sure `monitor` is very legible any more. 🤔

ericfennis:
I agree with @karsa-mistmere 😕

jguddas:
> `speaker` is 1000% better, but I'm not sure `monitor` is very legible any more. 🤔

Wasn't all that legible before either tho ;)

karsa-mistmere:
> > `speaker` is 1000% better, but I'm not sure `monitor` is very legible any more. 🤔
> 
> Wasn't all that legible before either tho ;)

While I do agree with this, it's not become any better either. 😅

PR #3461, opened 2025-07-28
chore(metadata): Adjust navigation category

### LINKED ISSUE #3092: Dumbbell icon in navigation category
Hello,
I noticed that the dumbbell icon is also being listed under the navigation category. I'm not sure if this was intentional, but it doesn't seem to fit that category. Please let me know if it should be removed. I would be happy to update it.

Image

Image

ericfennis:
Closes #3092

### DIFF
--- a/docs/.vitepress/data/categoriesData.json
+++ b/docs/.vitepress/data/categoriesData.json
@@ -101,7 +101,7 @@
   },
   {
     "name": "navigation",
-    "title": "Navigation"
+    "title": "Navigation, Maps, and POIs"
   },
   {
     "name": "notifications",
--- a/icons/arrow-big-down-dash.json
+++ b/icons/arrow-big-down-dash.json
@@ -15,7 +15,6 @@
   ],
   "categories": [
     "arrows",
-    "navigation",
     "gaming",
     "files"
   ]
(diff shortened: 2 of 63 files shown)


icon.schema.json
```diff
@@ -84,6 +84,7 @@
           "notifications",
           "people",
           "photography",
+          "poi",
```
jguddas: We don't use abbreviations for any other category
```suggestion
          "point-of-interest",
```

karsa-mistmere:
Can we clarify what sort of icons we aim for [Navigation​](https://lucide.dev/icons/categories#navigation) to contain? 🤔

It's kinda vague.

ericfennis:
@karsa-mistmere Yeah wait you're right, this category is a bit bloated now. 
The intention of the navigation category was in context for navigation and maps, etc. But arrows and other routing navigation stuff is added here. So it is messed up.

I would say let's revert the POI idea, remove all arrows and stuff, and rename it to "Navigation, Maps, and POIs".

ericfennis:
@karsa-mistmere @jguddas Updated

jguddas:
Don't forget to update the icon Schema JSON file

ericfennis:
@jguddas Not needed, I only changed the title and removed the arrows from the `navigation` category.

PR #3462, opened 2025-07-28
fix(icons): Rename `location-edit` to `map-pin-pencil`

ericfennis:
Renames `location-edit` to `map-pin-pen`. To match the other `map-pin` icons.


icons/map-pin-pen.json
```diff
@@ -18,5 +18,8 @@
     "navigation",
     "travel",
     "account"
+  ],
+  "aliases": [
+    "location-edit"
```
karsa-mistmere: Can we add a deprecation reason as per #2861?

ericfennis: Thanks updated it!

PR #3470, opened 2025-07-29
chore(docs): Update vitepress deps

ericfennis:
- Updates VitePress
- Cleanup CSS
- Fixes the name styling in the header


karsa-mistmere (changes requested):
Linting fails, we should fix that before we merge.

karsa-mistmere (approved):

PR #3500, opened 2025-08-04
feat(icons): added 'panel-dashed' variants + update tags on existing icons

irvineacosta:
## Description
Added new variants to the panel-dashed icon set:
- panel-left-right-dashed
- panel-top-right-dashed
- panel-all-sides-dashed

Also updated the json files of existing icons to include:
- margin
- padding
- guide

### Icon use case 
These icons will be used to represent paddings or margins, for example when customizing a UI element:
image

### Alternative icon designs 
image
image

### Concept 

### Author, credits & license

- [x] I've based them on the following Lucide icons: panel-dashed icon set (e.g. panel-top-dashed)

### Naming 

### Design 

## Before Submitting 

### DIFF
--- a/icons/panel-all-sides-dashed.json
+++ b/icons/panel-all-sides-dashed.json
@@ -6,8 +6,9 @@
     "irvineacosta"
   ],
   "tags": [
+    "sidebar",
+    "primary",
     "drawer",
-    "dock",
     "show",
     "reveal",
     "padding",
@@ -17,13 +18,5 @@
   ],
   "categories": [
     "layout"
-  ],
-  "aliases": [
-    {
-      "name": "panel-bottom-inactive",
-      "deprecationReason": "alias.name",
-      "deprecated": true,
-      "toBeRemovedInVersion": "v1.0"
-    }
   ]
 }
--- a/icons/panel-all-sides-dashed.svg
+++ b/icons/panel-all-sides-dashed.svg
@@ -1,0 +1,25 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="currentColor"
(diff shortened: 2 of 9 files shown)


irvineacosta:
@karsa-mistmere @jguddas as discussed here's the PR that expands the existing panel-dashed icons 😁

jguddas:
1. You should not duplicate icons that are just rotated versions of other icons.
2. `panel-all-sides-dashed` does not look like panels, I would go with #3499 instead.

In short: I would remove all icons besides `panel-left-right-dashed`.

karsa-mistmere:
> 1. You should not duplicate icons that are just rotated versions of other icons.

Please don't state this as fact, we should postpone merging orientations, as we are in no way ready for handling the automation of mirrored and rotated variants.

Forcing people to use CSS transforms is _not_ an option, since browsers (and HTML in general) are not our only target.

irvineacosta:
> > 1. You should not duplicate icons that are just rotated versions of other icons.
> 
> Please don't state this as fact, we should postpone merging orientations, as we are in no way ready for handling the automation of mirrored and rotated variants.
> 
> Forcing people to use CSS transforms is _not_ an option, since browsers (and HTML in general) are not our only target.

@jguddas it's also not possible to rotate instances inside figma components without detaching them.

jguddas:
> > > 1. You should not duplicate icons that are just rotated versions of other icons.
> > 
> > Please don't state this as fact, we should postpone merging orientations, as we are in no way ready for handling the automation of mirrored and rotated variants.
> > 
> > Forcing people to use CSS transforms is _not_ an option, since browsers (and HTML in general) are not our only target.
> 
> @jguddas it's also not possible to rotate instances inside figma components without detaching them.

I changed my mind, I think it's good as long as they have a unique use case.

jguddas (approved):

PR #3524, opened 2025-08-08
docs: update lucide-static documentation about raw string imports

pascalduez:
## Problem
While using `lucide-static` for templating I noticed that the documentation examples are not currently working. 
```js
const {messageSquare} = require('lucide-static/lib');
```
Reasons:
* The `lib` folder [is being ignored](https://github.com/lucide-icons/lucide/blob/528ffd6fa505d535fae3c8e5383d7527cbbf5c57/packages/lucide-static/.gitignore#L2) at the packaging step by npm/pnpm.
* Made a try to run the `build` alias in that package sources and the resulting `lib` folder is actually empty, it contains an empty `icons` folder.
```bash
packages/lucide-static/lib
└── icons

2 directories, 0 files
```

## Proposed resolution
Since everything that is needed is already present in the `dist` folder, which is CJS and ESM builds, and [properly declared](https://github.com/lucide-icons/lucide/blob/528ffd6fa505d535fae3c8e5383d7527cbbf5c57/packages/lucide-static/package.json#L26-L27) in `package.json` I though of promoting this approach instead, which is very common.  

## Side notes
* Which implies the [buildLib script](https://github.com/lucide-icons/lucide/blob/9db4dc7e1c024632bc28aabead7d0f6e5fa7eb4b/packages/lucide-static/scripts/buildLib.mts) is probably broken or not useful anymore?
* Took the liberty to document ESM usage in Node.js by default, it's 2025 after all, and we need to get rid of commonjs. 
* Added an example for Vite.
* The [source declaration](https://github.com/lucide-icons/lucide/blob/9db4dc7e1c024632bc28aabead7d0f6e5fa7eb4b/packages/lucide-static/package.json#L25) is not working either for published packages since [it's excluded](https://github.com/lucide-icons/lucide/blob/2150ce75ce61ea204fa67a784187cc464ab22a55/packages/lucide-static/.npmignore#L6).


pascalduez:
Well, not stale I guess... 

@ericfennis @karsa-mistmere @jguddas

PR #3543, opened 2025-08-14
feat(packages/lucide): added support for providing a custom root element

### LINKED ISSUE #3537: Support passing a custom root element to `createIcons`
### Package

- [x] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [ ] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] lucide-astro
- [ ] Figma plugin
- [ ] all JS packages
- [ ] site

### Description

When calling `lucide.createIcons`, I'd like to have the ability to specify a custom root element, so that only a specific part of the DOM tree is searched for potential icons needing to be created. Currently, it will always search the entire document, even if it's known that only a small part of the page needs to be searched. 

While the performance impact of this is probably not very large in most cases, this seems like a low-hanging fruit to avoid doing a whole bunch of unnecessary computational work on bigger pages with lots of DOM nodes.

### Use cases

I'm working on a website using HTMX and Alpine, which regularly update the DOM. Currently, after any update that potentially adds an icon to the page, I need triggers that run `createIcons` on the entirety of the page, even if I know that only a small part of the DOM could have been updated (e. g. a notification or modal). 

I believe this would be relevant to anyone building interactive websites using similar technologies, or even old-school PHP/AJAX style architectures. 

### Checklist

karsa-mistmere:
closes #3537

## Description

Added support for providing a custom root element to `createIcons` in `lucide`, for better performance when updating small parts of the DOM.

## Before Submitting 

### DIFF
--- a/packages/lucide/src/lucide.ts
+++ b/packages/lucide/src/lucide.ts
@@ -3,32 +3,32 @@ import * as iconAndAliases from './iconsAndAliases';
 
 /**
  * Replaces all elements with matching nameAttr with the defined icons
- * @param {{ icons?: object, nameAttr?: string, attrs?: object, context?: Element|Document }} options
+ * @param {{ icons?: object, nameAttr?: string, attrs?: object, root?: Element | Document }} options
  */
 const createIcons = ({
   icons = {},
   nameAttr = 'data-lucide',
   attrs = {},
-  context = document,
-}: { icons?: object; nameAttr?: string; attrs?: object; context?: Element|Document } = {}) => {
+  root = document,
+}: { icons?: object; nameAttr?: string; attrs?: object; root?: Element | Document } = {}) => {
   if (!Object.values(icons).length) {
     throw new Error(
       "Please provide an icons object.\nIf you want to use all the icons you can import it like:\n `import { createIcons, icons } from 'lucide';\nlucide.createIcons({icons});`",
     );
   }
 
-  if (typeof context === 'undefined') {
+  if (typeof root === 'undefined') {
     throw new Error('`createIcons()` only works in a browser environment.');
   }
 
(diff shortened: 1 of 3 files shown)


packages/lucide/src/lucide.ts
```diff
@@ -3,27 +3,32 @@ import * as iconAndAliases from './iconsAndAliases';
 
 /**
  * Replaces all elements with matching nameAttr with the defined icons
- * @param {{ icons?: object, nameAttr?: string, attrs?: object }} options
+ * @param {{ icons?: object, nameAttr?: string, attrs?: object, context?: Element|Document }} options
```
ericfennis: Maybe `root` fits better?
```suggestion
 * @param {{ icons?: object, nameAttr?: string, attrs?: object, root?: Element|Document }} options
```

packages/lucide/src/lucide.ts
```diff
@@ -3,27 +3,32 @@ import * as iconAndAliases from './iconsAndAliases';
```
ericfennis, line 3: Some lint errors in this file

karsa-mistmere: Huh, my bad, I ran `pnpm lint` within `packages/lucide` instead of, apparently `lint:format` in root.

packages/lucide/src/lucide.ts
karsa-mistmere: I guess I'm fine with either. I wanted to avoid a potentially loaded keyword, but I guess `root` will work fine.

PR #3544, opened 2025-08-16
fix(icons): changed `a-arrow-*` icons

jguddas:
- [x] Other: Icon update

### Description
Changed arrow height to match A.

## Before Submitting


karsa-mistmere:
I'd prefer for the arrows to have a square bounding box as most of our other arrows do.

PR #3545, opened 2025-08-16
fix(icons): changed `ampersands` icon

jguddas:
- [x] Other: Icon update

### Description
Redesign, hmm…

## Before Submitting


karsa-mistmere:
This is not really the same thing as `&&` though. 🤔

Not even Fira Code is this drastic about the `&&` ligature. 😅

ericfennis:
Nope, I'm also against this change. The icon is not very clear anymore in this way

PR #3546, opened 2025-08-16
fix(icons): changed `banknote-arrow-up` icon

jguddas:
- [x] Other: Icon update

### Description
Made arrow bigger

## Before Submitting

### DIFF
--- a/icons/banknote-arrow-down.json
+++ b/icons/banknote-arrow-down.json
@@ -4,7 +4,8 @@
     "AnnaSasDev",
     "joffx",
     "ericfennis",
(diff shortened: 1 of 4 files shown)


karsa-mistmere:
I'm pretty sure I'd explored the larger arrow for these, but I'm still uneasy about this change, since it does decrease the legibility of the banknote quite a bit. 🤔

ericfennis:
I agree. 
I'm bothered by the unconnected circle in the middle

ericfennis:
I'm closing this one. Feels not like a good change.

PR #3547, opened 2025-08-16
fix(icons): changed `server` icon

jguddas:
- [x] Other: Icon update

### Description
Changed roundness and decreased size.
Now we have a gap violation, but I think that is better than the icon being to big.

## Before Submitting


karsa-mistmere:
Why the uneven rounding? 🤔

I'm not sure about which is more important, optical volume or the 2px gap. 🫤

Plus we might need to correlate this with some other icons, namely:
- https://lucide.dev/icons/router
- https://lucide.dev/icons/hard-drive
- https://lucide.dev/icons/hard-drive-download
- https://lucide.dev/icons/hard-drive-upload

I guess that'd lead to these ones not really matching in style: 
image

ericfennis:
I don't like the uneven rounding either 😕

ericfennis:
I'm not sure about this change. I think it makes the icons too small

ericfennis:
Let's keep it as it is. 2px gap is more important than the optical volume, also icons feel cramped now. This introduces less consistency.

PR #3576, opened 2025-08-22
fix(packages/lucide): replace elements inside `<template>` (#2635)

### LINKED ISSUE #2635: Lucide won't render `data-lucide` if inside `<template>`
### Package

- [X] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [ ] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

0.460.0

- [X] Yes
- [ ] No

### Browser

- [X] Chrome/Chromium
- [X] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [ ] Other/not relevant

### Operating system

- [ ] Windows
- [X] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [ ] Other/not relevant

### Description

I'm using the [Content Template element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) (`<template>`) to render parts of an HTML form that I use as a template (duh) within JavaScript. Elements with `data-lucide` inside the `template` tag do not render.

### Steps to reproduce

1. Minimal setup from guide (all icons)
2. The following input HTML:
    ```html
    <div>
       <template><i data-lucide="trash-2" class="size-6 stroke-red-600 mx-auto"></i></template>
       <i data-lucide="trash-2" class="size-6 stroke-red-600 mx-auto"></i>
    </div>
    ```
3. Results in:
    ```html
        <div>
          <template><i data-lucide="trash-2" class="size-6 stroke-red-600 mx-auto"></i></template>
          <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" data-lucide="trash-2" class="lucide lucide-trash-2 size-6 stroke-red-600 mx-auto"><path d="M3 6h18"></path><path d="M19 6v14c0 1-1 2-2 2H7c-1 0-2-1-2-2V6"></path><path d="M8 6V4c0-1 1-2 2-2h4c1 0 2 1 2 2v2"></path><line x1="10" x2="10" y1="11" y2="17"></line><line x1="14" x2="14" y1="11" y2="17"></line></svg>
    </div>
    ```

### Checklist

- [X] I have searched if someone has submitted a similar issue before and there weren't any. (Please make sure to also search closed issues, as this issue might already have been resolved.)

KhalidAlansary:
closes #2635
## Description
Find and replace elements inside `<template>` tags that are not normally selected with `document.querySelector`

## Before Submitting

### DIFF
--- a/packages/lucide/src/lucide.ts
+++ b/packages/lucide/src/lucide.ts
@@ -21,10 +21,13 @@ const createIcons = ({
     throw new Error('`createIcons()` only works in a browser environment.');
   }
 
-  const elementsToReplace = root.querySelectorAll(`[${nameAttr}]`);
-  Array.from(elementsToReplace).forEach((element) =>
-    replaceElement(element, { nameAttr, icons, attrs }),
-  );
+  const elementsToReplace = Array.from(root.querySelectorAll(`[${nameAttr}]`));
+  const templates = Array.from(root.querySelectorAll('template'));
+  templates.forEach((template) => {
+    const contentToReplace = Array.from(template.content.querySelectorAll(`[${nameAttr}]`));
+    elementsToReplace.push.apply(elementsToReplace, contentToReplace);
+  });
+  elementsToReplace.forEach((element) => replaceElement(element, { nameAttr, icons, attrs }));
 
   /** @todo: remove this block in v1.0 */
   if (nameAttr === 'data-lucide') {


karsa-mistmere:
All in all, I'm not really in favour of adding this feature: templates are meant to be templates, I think it makes more sense for icons to be replaced _after_ instantiating them, their contents aren't included in `querySelectorAll` for a reason.

If you really need to run `createIcons` on some templates ahead of time, you can already do so by specifying a custom root element, for example:

```ts
createIcons({ root: templateElement });
```

Or, to replace all icons in all templates:

```ts
document.querySelectorAll('template').forEach(root => createIcons({ root }));
```

Additionally, even if we were to add official support for this, it should be done in a backwards-compatible way — for instance, by introducing a boolean flag that controls whether icons inside templates should be replaced. This flag should default to false, so the current behaviour remains unchanged.

@ericfennis, @jguddas, what are your thoughts?

ericfennis:
Yeah, I agree with @karsa-mistmere on this one. It's contained for a reason.

I've added a new option: `replaceInsideTemplates`, to the createIcons function. And I also added some tests.

packages/lucide/src/lucide.ts
```diff
@@ -22,11 +32,16 @@ const createIcons = ({
...
-    elementsToReplace.push.apply(elementsToReplace, contentToReplace);
-  });
+
+  if (replaceInsideTemplates) {
+    const templates = Array.from(root.querySelectorAll('template'));
+
+    templates.forEach((template) => {
+      const contentToReplace = Array.from(template.content.querySelectorAll(`[${nameAttr}]`));
+      elementsToReplace.push.apply(elementsToReplace, contentToReplace);
+    });
```
karsa-mistmere: @ericfennis I wonder if maybe this feature should be made recursive? 🤔

In essence, something like:

```ts
if (replaceInsideTemplates) {
  Array.from(root.querySelectorAll('template')).forEach(template => createIcons({
    icons,
    nameAttr,
    attrs,
    root: template,
    replaceInsideTemplates,
  });
}

ericfennis: Good one, this makes the code a lot cleaner.

karsa-mistmere: > Good one, this makes the code a lot cleaner.

But then again, people who do need to do this could just do this themselves. 😄

ericfennis:
Renamed the option to `inTemplates`. Since the term "replace Icons" does not feel right. So `createIcons` -> `inTemplates`.
@karsa-mistmere I've applied your suggestion.

I've also added some documentation.

packages/lucide/src/lucide.ts
ericfennis: True, but it is also nice to extend the current API to enable this.

PR #3584, opened 2025-08-24
fix(icons): changed `panel-top-bottom-dashed` icon

jguddas:
- [x] Other: Icon update

### Description
Moved dashes to match rest of the icons.

## Before Submitting

### DIFF
--- a/icons/panel-left-right-dashed.json
+++ b/icons/panel-left-right-dashed.json
@@ -3,7 +3,8 @@
   "contributors": [
     "danielbayley",
     "ericfennis",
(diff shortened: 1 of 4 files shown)


jguddas:
Alternatively we could modify the other pannel icons.

karsa-mistmere:
> Alternatively we could modify the other panel icons.

Not sure we should break the harmony between `panel` and `calendar`:

image

karsa-mistmere (approved):

ericfennis (approved):

PR #3587, opened 2025-08-25
fix(icons/file): arcified folds

karsa-mistmere:
## Description

Added rounding to `file-*` fold corners & applied this change to other icons with folds.

Also merged, renamed & updated many `file-*` icons to have a more consistent feel, sticking to a guideline-compliant naming scheme.

- Moved most "external" addons to the bottom left corner and renamed them from `-2` => `-corner`
- Removed most icon variants where the fold wasn't visible and added a backwards compatible alias for them

## Before Submitting 


icons/file-check-corner.svg
```diff
@@ -9,8 +9,7 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
```
ericfennis, line 11: Just checking is this weird github diff issues?

icons/file-scan.svg
```diff
@@ -9,10 +9,10 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M20 10V7l-5-5H6a2 2 0 0 0-2 2v16a2 2 0 0 0 2 2h4" />
-  <path d="M14 2v4a2 2 0 0 0 2 2h4" />
+  <path d="M20 8a2.4 2.4 0 0 0-.706-1.704l-3.588-3.588A2.4 2.4 0 0 0 14 2H6a2 2 0 0 0-2 2v16a2 2 0 0 0 2 2h4.35" />
```
ericfennis: Hmm not sure about this one. Removing the line effects the file shape a lot.

icons/file-user.svg
```diff
@@ -9,8 +9,8 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M14 2v4a2 2 0 0 0 2 2h4" />
-  <path d="M15 18a3 3 0 1 0-6 0" />
-  <path d="M15 2H6a2 2 0 0 0-2 2v16a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V7z" />
-  <circle cx="12" cy="13" r="2" />
+  <path d="M6 22a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h8a2.4 2.4 0 0 1 1.704.706l3.588 3.588A2.4 2.4 0 0 1 20 8v12a2 2 0 0 1-2 2z" />
```
ericfennis, line 12: Hmm not sure about this design. Look a bit like someone looking from a window 🤔

icons/file-check-corner.svg
karsa-mistmere: Yeah, I'm afraid it'll keep happening when a lot of files change, no matter if you use `git mv`. :(

icons/file-scan.svg
karsa-mistmere: Huh, no, this wasn't done purposefully anyway, not sure where that line segment disappeared, it's fixed now.

icons/file-user.svg
karsa-mistmere: I agree that it feels unusual – albeit maybe mostly because it's not what we got used to.

This is very much in line with other user icons though, e.g.:
- https://lucide.dev/icons/circle-user-round
- https://lucide.dev/icons/square-user-round
- https://lucide.dev/icons/contact-round
- `book-user` as redesigned in #3171 by @jguddas

ericfennis: Hmm, true. But the design is not perfect, also for #3171. 
Also breaks consistency with other `file-*` variants. None of the smaller icons are connected with the "file outer shape".

jguddas: I think we can keep it the old way and move this change into the other PR

ericfennis (approved):

PR #3604, opened 2025-09-05
feat(packages): Added aria-hidden fallback for decorative icons to all packages

ericfennis:
It was already added to the React package in #2158

This makes sure we add to all other packages.

TODO:
- [x] Add notice in Docs

@karsa-mistmere FYI, I've also added it to the Angular package.


docs/guide/packages/lucide-preact.md
```diff
@@ -124,3 +124,19 @@ const App = () => {
...
+
+By default, we hide icons from screen readers using `aria-hidden="true"`.
+
+You can add accessibility attributes using aria-labels.
+
+```jsx
+import { Camera } from 'lucide-preact';
+
+const App = () => {
+  return <Camera aria-label="Camera icon" />;
```
jguddas: That's not good alt text, you don't need to have "icon" in there.

packages/lucide-angular/src/lib/lucide-angular.component.ts
```diff
@@ -133,6 +142,14 @@ export class LucideAngularComponent implements OnChanges {
     if (typeof this.name === 'string') {
       icoElement.classList.add(`lucide-${this.name.replace('_', '-')}`);
     }
+
+    // Forward aria-* and role from host to svg
+    Object.entries(restAttributes).forEach(([attr, value]) => {
+      if (attr.startsWith('aria-') || attr === 'role' || attr === 'title') {
+        this.renderer.setAttribute(icoElement, attr, value as string);
+      }
+    });
```
karsa-mistmere: Yeah, this is... not really optimal, I'm afraid and will lead to DOM that looks something like this:

```html
<lucide-icon title="Camera" aria-label="Camera" role="alert">
  <svg title="Camera" aria-label="Camera" role="alert" ...>
    ...
  </svg>
</lucide-icon>
```

i.e. every single ARIA, role and title attribute will get duplicated, which is probably not something we want. IMO Angular icon generation doesn't even really need to be touched, we can document what we recommend, but there's really no need to copy the attributes, people can add accessible labels to their icons on the component host element, and the SVG itself can remain hidden.

ericfennis: Hmm oke, seems good idea.  So, adding a permanent aria-hidden on the SVG element and letting the host element handle the aria-labels/titles

docs/guide/packages/lucide-preact.md
ericfennis: This is an example. But this is what screenreaders will read. Just "camera" is not really helping IMO.

packages/lucide-angular/src/lib/lucide-angular.component.ts
ericfennis: @karsa-mistmere I've checked this an it is only applied to the svg and not the host element.

docs/guide/packages/lucide-astro.md
```diff
@@ -185,3 +185,19 @@ import LucideIcon from './LucideIcon.astro';
...
+By default, we hide icons from screen readers using `aria-hidden="true"`.
+
+You can add accessibility attributes using aria-labels.
+
+```jsx
+---
+import { Check } from '@lucide/astro';
+---
+
+<Check aria-label="Task completed" />
```
ericfennis, line 207: @jguddas I've changed it to this

jguddas: Nice

packages/lucide-angular/src/lib/lucide-angular.component.ts
karsa-mistmere: > @karsa-mistmere I've checked this an it is only applied to the svg and not the host element.

Not sure how you checked, but I've just built and installed the package, and no, it does get applied to both elements:

image

karsa-mistmere: Sidenote: for `@lucide/angular` maybe we should switch to an `svg[lucideIcon]` attribute selector from the current element selectors:

```diff
- <lucide-icon [img]="ChevronLeftIcon"></lucide-icon>
+ <svg [lucideIcon]="ChevfronLeftIcon"></svg>
```

That way, there wouldn't be any wrapping element to consider, and we could use host bindings for most stuff instead of having to manipulate the DOM manually.

ericfennis:
@karsa-mistmere Thanks for the fixes. Is this ready to go?

packages/lucide-angular/src/lib/lucide-angular.component.ts
karsa-mistmere: > @karsa-mistmere Thanks for the fixes. Is this ready to go?

No, unfortunately this behaviour remains unchanged.

ericfennis:
@karsa-mistmere Sorry, I was a bit confused with your changes and comments.
But I finally got it working. The changes are only applied to the root element (`lucide-angular, lucide-icon, i-lucide, span-lucide`).
If no aria attribute is added, `aria-hidden="true"` is added.  So no other attributes are added to the SVG itself.

PR #3617, opened 2025-09-11
feat(icons): add `github-actions`

danielbayley:
## Description

GitHub Actions representative icon.

### Icon use case 

Represents any functionality related to GitHub Actions, or e.g. a grouping of [`workflow`](https://lucide.dev/icons/workflow)s. Related to the many [`git-*`](https://lucide.dev/icons?search=git) icons.

### Concept 

### Author, credits & license

- [x] I've based them on the following design: 

image

### Naming 

- [ ] I've named icons by what they are rather than their use case.

### Design 

## Before Submitting 


karsa-mistmere:
Hey @danielbayley,

1) The inevitable legal stuff: do we know who made the original you based this on, how it is licensed, and if it's protected by trademark or copyright? I'm wary about adding a remix without knowing the origins.

2) Regardless of the previous, the name GitHub _is_ protected, so we should try to avoid naming it as such, maybe something like `git-pipeline` instead?

3) With the omission of internal details, I feel like much of the original intent is lost. It's pretty busy as is but it's still kind of failing to convey that CI pipeline character. 🤔

4) The optical volume is a tad too large.

karsa-mistmere:
5. I'm only now noticing, but there's also a gap violation in the top left part:
image

danielbayley:
1\. I think it has been abstracted enough to not be a copyright issue. It’s not a brand…

2\. `git-actions`?

4/5. Fixed.

karsa-mistmere:
> 1. I think it has been abstracted enough to not be a copyright issue. It’s not a brand…

Well, I understand your argument, but… we've literally documented the derivation right here in this PR. 😅
That's exactly the kind of thing that _doesn't_ magically make copyright concerns disappear.

> **_4_**/5. Fixed.

Not sure what I’m supposed to be seeing here—the optical volume looks exactly the same to me as before 🤔.

(I had to dig the original version out of my e-mails since, despite previous requests, you force-pushed over it 🙄.)

Issue #3557, opened 2025-08-18
No brand logos (e.g. Discord, Google, X etc), read this


karsa-mistmere:
## About brand logos

Lucide **does not accept** brand logos, and we do not plan to add them in the future. This is due to a combination of **legal restrictions**, **design consistency concerns**, and **practical maintenance reasons**.

[Click here to read our official statement about brand logos in Lucide.](https://github.com/lucide-icons/lucide/blob/main/BRAND_LOGOS_STATEMENT.md)

PR #3639, opened 2025-09-19
fix(icons): Remove brand icons

ericfennis:
# Description

See announcement: #3557
Relevant discussions: #670 #94  

Contributes to: #1687


karsa-mistmere (changes requested):
We should also update https://github.com/lucide-icons/lucide/blob/main/icon.schema.json#L59 & https://github.com/lucide-icons/lucide/blob/main/icon.schema.json#L137

(@jguddas and https://github.com/jguddas/lucide-studio/blob/9705849cad83c463dc09aafb158349ad644eaf41/src/components/ContributionDialog/category-options.ts#L72 )

ericfennis:
@karsa-mistmere Thanks for your feedback!

karsa-mistmere (approved):

PR #3641, opened 2025-09-19
chore(packages): Remove umd exports

ericfennis:
# Description

For all non-framework packages, we are going to remove UMD support for packages

- UMD is legacy
UMD was created to support environments before ES modules and modern bundlers were widely adopted. Today, every serious frontend workflow (Vite, Webpack, Rollup, esbuild, Next.js, etc.) expects ESM or CJS. UMD adds complexity without solving a real problem.

- It bloats the package
Supporting UMD means generating extra build artifacts, which increases package size and maintenance overhead. That’s code we’d ship, but nobody actually needs.

- Bundlers should tree-shake the icons you only use.

- It’s the wrong default in 2025
Supporting UMD today encourages outdated practices. We’d rather push the ecosystem forward than spend time using old tooling.

### DIFF
--- a/packages/lucide/package.json
+++ b/packages/lucide/package.json
@@ -51,7 +51,7 @@
     "rollup": "^4.40.0",
     "rollup-plugin-dts": "^6.2.1",
     "typescript": "^5.8.3",
-    "vite": "^6.3.4",
+    "vite": "^6.3.6",
     "vitest": "^3.1.3"
   }
 }


jguddas:
That is a pretty major breaking change, this can only really be merged as a V1, right?

ericfennis:
@jguddas Yess correct

ericfennis:
But I will keep the support for the regular Lucide package.
I'm not even sure if the others work. Long time ago I tested that.

PR #3663, opened 2025-10-01
feat(icons): refine `scroll`[`-text`]

danielbayley:
## Description

Refine `scroll` and `scroll-text` icons.

## Before Submitting 


karsa-mistmere:
I feel like the added sharp corners don't quite capture the Lucide feel. 🤔

The redesign overall leans a bit more into a fantasy-like style compared to what we have now, and it also seems pretty close to Octicons' https://www.primer.style/octicons/icon/log-16/

karsa-mistmere:
Based on @ericfennis' response and mine, I'm closing this in favour of the status quo.

PR #3718, opened 2025-10-23
feat(docs): add keyboard shortcut for search

dzonatan:
## Description

Keyboard-geek here.
I miss this shortcut every time I open lucide.dev to quickly find the appropriate icon.

image

Depending on the OS it will either show CMD + K for Mac, or CTRL + K for Windows.
Also added `/` forward slash which is a very common shortcut for this matter.

Disclosure: I'm not experienced with vue, therefore some AI was used.
No feelings if this gets rejected.

## Before Submitting 

### DIFF
--- a/docs/.vitepress/theme/components/base/FakeInput.vue
+++ b/docs/.vitepress/theme/components/base/FakeInput.vue
@@ -3,12 +3,20 @@ import createLucideIcon from 'lucide-vue-next/src/createLucideIcon'
 import { search }  from '../../../data/iconNodes'
 
 const SearchIcon = createLucideIcon('search', search)
+
+defineProps({
+  shortcut: {
+    type: String,
+    required: false
+  }
+})
 </script>
 
 <template>
   <button class="fake-input">
     <component :is="SearchIcon" class="search-icon"/>
     <slot/>
+    <kbd v-if="shortcut" class="shortcut">{{ shortcut }}</kbd>
   </button>
 </template>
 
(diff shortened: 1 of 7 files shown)


docs/.vitepress/theme/components/base/FakeInput.vue
```diff
@@ -33,4 +41,17 @@ const SearchIcon = createLucideIcon('search', search)
   border-color: var(--vp-c-brand);
   background: var(--vp-c-bg-alt);
 }
+
+.shortcut {
+  margin-left: auto;
+  padding: 2px 6px;
+  font-size: 12px;
```
jguddas, line 48: Are we doing font size in x everywhere?

jguddas (commented):
Nice

docs/.vitepress/theme/components/base/FakeInput.vue
dzonatan: looks like it, 18 lines above `.fake-input` also sets font size in pixels.
but I can change to relative unit if you say so, just let me know.

jguddas:
Should we hide the shortcut on mobile?

dzonatan:
> Should we hide the shortcut on mobile?

Yeah, I guess it makes sense to hide. 
Is there a preferred way to identify mobile devices? `pointer: coarse` media query, user agent matching, or screen size (>=960px) matching?

ericfennis:
Nice addition!!

ericfennis:
Maybe also nice to add it to the search bar on the icons page.

dzonatan:
As suggested: added to the icons overview page and hidden on mobile.
Let me know if I should squash those commits.

jguddas:
It feels a bit weird that it disappears on focus.
My suggestion is to hide it entirely on small screens and keep it always visible on big screens.

dzonatan:
> It feels a bit weird that it disappears on focus. My suggestion is to hide it entirely on small screens and keep it always visible on big screens.

I intentionally did this as it was getting in the way of this clear button which is specific to "search" type input.
image
Not sure what's better, adding extra padding or hiding it on focus.

karsa-mistmere:
> > It feels a bit weird that it disappears on focus. My suggestion is to hide it entirely on small screens and keep it always visible on big screens.
> 
> I intentionally did this as it was getting in the way of this clear button which is specific to "search" type input. image Not sure what's better, adding extra padding or hiding it on focus.

Adding extra padding is definitely the way. 👍

dzonatan:
added padding.
localhost_5173_icons_mac
localhost_5173_icons_win

ericfennis (approved):
@dzonatan Excellent work!

PR #3730, opened 2025-10-24
Fix default prop

ericfennis:
Fixes the UMD build Lucide package.

Accidentally removed the default prop of the `createIcons` function in https://github.com/lucide-icons/lucide/pull/3576.


jguddas:
Maybe it's time to write some tests?

ericfennis:
@jguddas haha jump, thought the same.

PR #3746, opened 2025-10-31
fix(icons): rearange `circle`-icons path and circle order

adamlindqvist:
## Description
Update the order of the svg elements of icons with a circle "frame" for it to be consistent where `circle` always is the first element. This makes sure that consumers are able to set a custom background on a icon with css.

### Example:

```css
.icon {
  color: #000000;

  circle {
    fill: #ff0000;
  }
}
```

```jsx
import { Clock9 } from "lucide-react";

<Clock9 className="icon" size={32} strokeWidth={1.5} />
```

**Before:**

image

**After:**

image

## Before Submitting 

### DIFF
--- a/icons/towel-rack.json
+++ b/icons/towel-rack.json
@@ -1,0 +1,30 @@
+{
+  "$schema": "../icon.schema.json",
+  "contributors": [
+    "danielbayley",
+    "jguddas",
+    "karsa-mistmere"
+  ],
+  "tags": [
+    "flannel",
+    "bathroom",
+    "toiletries",
+    "sanitation",
+    "clean",
+    "fresh",
+    "dry",
+    "laundry",
+    "laundrette",
+    "hospitality",
+    "housekeeping",
+    "room service",
+    "spa break",
+    "health club",
+    "amenities",
+    "hanging"
+  ],
+  "categories": [
+    "home",
+    "travel"
+  ]
+}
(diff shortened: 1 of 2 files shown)


adamlindqvist:
> Actually this should be the other way around so people do not use fills and we accidentally break it for them.

I'm sorry I don't think I follow on that comment, do you mean this changes should be reverted? 

I know about the comment in https://lucide.dev/guide/advanced/filled-icons: "Fills are officially not supported. However, all SVG properties are available on all icons. Fill can still be used and will work fine on certain icons.". But I'm not sure what you want to change?

karsa-mistmere:
> > Actually this should be the other way around so people do not use fills and we accidentally break it for them.
> 
> I'm sorry I don't think I follow on that comment, do you mean this changes should be reverted?
> 
> I know about the comment in https://lucide.dev/guide/advanced/filled-icons: "Fills are officially not supported. However, all SVG properties are available on all icons. Fill can still be used and will work fine on certain icons.". But I'm not sure what you want to change?

I’d be interested to hear @ericfennis’ take on this.

Personally, I don’t fully agree with @jguddas’ concern here — I think the PR is fine as it stands. While fills aren’t officially supported, I don’t see harm in structuring icons in a way that doesn’t actively work against being filled, especially given that fills are already documented as something that may work on certain icons.

To me, this doesn’t meaningfully increase the risk of breakage, but it does improve consistency and flexibility.

ericfennis:
I agree with @karsa-mistmere, this is not really breaking anything. I'm okay with adding this.

ericfennis (approved):

PR #3690, opened 2025-10-15
feat(icons): added `ai` icon

markoidan:
## What is the purpose of this pull request?
- [x] New Icon

### Description
Added new `ai` icon.

### Icon use case

ai 
chatbot
assistant

### Concept

- [x] I have provided valid use cases for each icon.
- [x] I have not added any a brand or logo icon.
- [x] I have not used any hate symbols.
- [x] I have not included any religious or political imagery.

### Author, credits & license

- [ ] The icons are solely my own creation.
- [ ] The icons were originally created in #<issueNumber> by @<githubUser>
- [ ] I've based them on the following Lucide icons: 
- [ ] I've based them on the following design: 

### Naming

- [x] I've read and followed the [naming conventions](https://lucide.dev/guide/design/icon-design-guide#naming-conventions)
- [x] I've named icons by what they are rather than their use case.
- [x] I've provided meta JSON files in `icons/[iconName].json`.

### Design

- [ ] I've read and followed the [icon design guidelines](https://lucide.dev/guide/design/icon-design-guide)
- [ ] I've made sure that the icons look sharp on low DPI displays.
- [ ] I've made sure that the icons look consistent with the icon set in size, optical volume and density.
- [ ] I've made sure that the icons are visually centered.
- [ ] I've correctly optimized all icons to three points of precision.

## Before Submitting
- [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [ ] I've checked if there was an existing PR that solves the same issue.

PR #3759, opened 2025-11-05
Studio/stars

markoidan:


karsa-mistmere:
Duplicates https://lucide.dev/icons/sparkles and #3690

Please do not submit the same design twice without fixing any of the underlying issues.

Edit: also locking conversations on this one, please reply to the issues outlined in #3690

PR #3825, opened 2025-11-26
feat(scripts): Enable icons dir in CLI arguments

Garbanas:
* Allow to override icons dir through CLI argument '--icons'
* Enable rescursive mapping of SVG child nodes
* Provide complete SVG with all child nodes to template function

## Description
Providing the entire SVG including child nodes to the template functions enables usage of more advanced SVG icon use cases, like using SVG masks.

Example:
```html
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" viewBox="0 0 24 24">
    <circle cx="12" cy="12" r="9" fill="currentColor" mask="url(#a)"/>
    <mask id="a">
        <path fill="#fff" d="M0 0h24v24H0z"/>
        <path fill="none" stroke="#000" d="M9 12.75 11.25 15 15 9.75"/>
    </mask>
</svg>
```
icons

If mapped correctly, modifiying stroke-width will result in a differently sized checkmark symbol on a circle, where the circle is full color and the checkmark is masked out (transparent).

This way dynamic "solid" icon variants are possible.

## Before Submitting

### DIFF
--- a/packages/astro/src/types.ts
+++ b/packages/astro/src/types.ts
@@ -16,7 +16,7 @@ export type SVGAttributes = HTMLAttributes<'svg'>;
 
 export type IconNode = IconNodeChild[];
 
-type IconNodeChild = [elementName: SVGElements, attrs: SVGAttributes];
+type IconNodeChild = [elementName: SVGElements, attrs: SVGAttributes, children: IconNodeChild[]];
 
 // All possible svg elements according to the Astro definitions
 type SVGElements =
--- a/packages/lucide-angular/src/icons/types/index.ts
+++ b/packages/lucide-angular/src/icons/types/index.ts
@@ -1,5 +1,5 @@
 type HtmlAttributes = { [key: string]: string | number };
-export type LucideIconNode = readonly [string, HtmlAttributes];
+export type LucideIconNode = readonly [string, HtmlAttributes, LucideIconData];
 export type LucideIconData = readonly LucideIconNode[];
 export type LucideIcons = { [key: string]: LucideIconData };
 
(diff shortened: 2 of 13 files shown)


karsa-mistmere:
Thanks a lot for the PR and the detailed example — I can definitely see the potential use case here as we've already explored this in #458.

However, Lucide currently doesn’t support these kinds of structures, and **_that’s fully intentional_**. Introducing nested SVG elements like `<mask>` or `<defs>` with hardcoded `id` attributes leads to **invalid HTML when the same icon is rendered more than once**, since `id`s must be unique across the entire document.

So if we allow arbitrary child nodes, simply enabling recursive mapping wouldn’t be enough — every renderer/package (React, Vue, Svelte, Angular, Figma plugin, etc.) would also have to implement a **reliable ID–namespacing or prefixing strategy** to guarantee uniqueness. Without that, we’d end up with collisions and broken masks/gradients across the board.

That means:
- Including child SVG nodes is not a problem *by itself*
- **But** supporting them properly requires a full spec/strategy for handling `id`s safely across all packages
- Otherwise icons with `<mask>` / `<defs>` would render fine once, but break as soon as more than one copy is used

So while we’re not currently able to merge this as-is, masks and “solid variants” are definitely something we’d like to explore in the future — just needs to be done in a way that works safely across packages and multiple instances.

jguddas:
> Lucide currently doesn’t support these kinds of structures, and that’s fully intentional.

👍 it also just keeps things so much more simple.

We have thought about how to support fills a lot. My vote is having them as separate instances.
Adding masks just adds to much complexity IMO.

### `square-filled` icon example:

```ts
import { createLucideIcon } from "lucide-react";
export const SquareFilledIcon = createLucideIcon("square", [
  [
    "rect",
    { x: "3", y: "3", width: "18", height: "18", rx: "2", fill: "currentColor", key: "h1oib" },
  ],
]);
```

PR #3836, opened 2025-12-01
docs(dev): Fix code sample for vanilla JS

wavebeem:
## Description

The `.append` method of elements takes plain text, not HTML. This updates the code to use `.insertAdjacentHTML` instead.

## Before Submitting

### DIFF
--- a/docs/.vitepress/lib/codeExamples/createCodeExamples.ts
+++ b/docs/.vitepress/lib/codeExamples/createCodeExamples.ts
@@ -13,14 +13,15 @@ const getIconCodes = (): CodeExampleType => {
       title: 'Vanilla',
       code: `\
 <script>
-import { createIcons } from 'lucide';
-import { $CamelCase } from '@lucide/lab';
+import { createIcons, $CamelCase } from 'lucide';
+
 createIcons({
   icons: {
     $CamelCase
   }
 });
 </script>
+
 <i data-lucide="$Name"></i>\
   `,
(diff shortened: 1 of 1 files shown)


docs/.vitepress/lib/codeExamples/createLabCodeExamples.ts
```diff
@@ -22,7 +22,7 @@ createIcons({
   }
 });
 
-document.body.append('<i data-lucide="$Name"></i>');\
+document.body.insertAdjacentHTML('beforeend', '<i data-lucide="$Name"></i>');\
```
karsa-mistmere: TBH, maybe we should skip this entire statement and have the code example be a HTML block containing something like:

```html
<script>
import { createIcons } from 'lucide';
import { $CamelCase } from '@lucide/lab';

createIcons({
  icons: {
    $CamelCase
  },
});
</script>

<i data-lucide="$Name"></i>
```

I think for vanilla JS this would make much more sense than appending to the DOM.

ericfennis: Agreed

wavebeem:
Updated per code review feedback 🙏🏻

karsa-mistmere (approved):
> Updated per code review feedback 🙏🏻

Thanks! I also updated the non-lab code sample to use the same logic, from my part this is now good to go.

PR #3851, opened 2025-12-04
fix(site): Small adjustments color picker and add clear button search bar

ericfennis:
## Adjustments site

- Color pick with "normal" value
Screenshot 2025-12-04 at 15 04 31

- Searchbar with a clear button

image

- Removed the Flutter package from the homepage

### DIFF
--- a/docs/.vitepress/theme/components/base/ColorPicker.vue
+++ b/docs/.vitepress/theme/components/base/ColorPicker.vue
@@ -13,7 +13,7 @@ const emit = defineEmits(['update:modelValue']);
 
 const value = computed({
   get: () => {
-    if (props.modelValue === 'currentColor') {
+    if (props.modelValue == null || props.modelValue === 'currentColor') {
       return isDark.value ? '#ffffff' : '#000000';
     }
 


docs/.vitepress/theme/components/base/Input.vue
```diff
@@ -86,6 +122,14 @@ defineExpose({
   padding-left: 52px;
 }
 
+.clear-button {
+  position: absolute;
+  right: 56px;
+  top: 9px;
+  background-color: var(--vp-button-alt-bg);
+  padding: 4px;
```
karsa-mistmere: The clear button feels a bit intrusive to me, should we maybe only display the background on hover?

image

```suggestion
  padding: 4px;
  transition: background-color .25s;
```

docs/.vitepress/theme/components/base/ColorPicker.vue
```diff
@@ -1,23 +1,31 @@
...
+const { isDark } = useData();
+
+const emit = defineEmits(['update:modelValue']);
 
 const value = computed({
-  get: () => props.modelValue,
-  set: (val) => emit('update:modelValue', val)
-})
+  get: () => {
+    if (props.modelValue === 'currentColor') {
```
karsa-mistmere: How about?
```suggestion
    if (!props.modelValue || props.modelValue === 'currentColor') {
```

ericfennis: This can be annoying if you want to clear the field. But we can make it explicit.

docs/.vitepress/theme/components/base/Input.vue
ericfennis: Good point!

docs/.vitepress/theme/components/base/ColorPicker.vue
ericfennis: :P 
https://github.com/user-attachments/assets/9bd97bd3-26b8-4d9a-b8a3-847106bd71ad

ericfennis: Add check if value is `undefined` or `null`

Issue #3744, opened 2025-10-30
lucide-react-native adding ~13k lines of repeated license text to production output bundle expo-router/web


gorbypark:
### Package

- [ ] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [ ] lucide-react
- [x] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] lucide-astro
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

0.548.0

### Can you reproduce this in the latest version?

- [x] Yes
- [ ] No

### Browser

- [ ] Chrome/Chromium
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [x] Other/not relevant

### Operating system

- [ ] Windows
- [ ] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [x] Other/not relevant

### Description

When using lucide-react-native in an expo-router based expo project on the web (metro bundling for web), if you do `npx expo export --platform web` and then inspect the `entry_xxxx.js` bundle, there is about 13,000 lines of text that is just the repeated lucide license text.  I haven't counted it up, but I'm assuming it's one license per lucide icon?

I'm guessing when serving the bundle the servers compression will take care of mostly compressing it away, but it does seem a bit excessive!

```
const u=(0,((e=t)&&e.__esModule?e:{default:e}).default)("ZoomIn",[["circle",{cx:"11",cy:"11",r:"8",key:"4ej97u"}],["line",{x1:"21",x2:"16.65",y1:"21",y2:"16.65",key:"13gj7c"}],["line",{x1:"11",x2:"11",y1:"8",y2:"14",key:"1vmskp"}],["line",{x1:"8",x2:"14",y1:"11",y2:"11",key:"durymu"}]])},5122,[3462]);
__d(function(g,r,i,a,m,_e,d){"use strict";Object.defineProperty(_e,'__esModule',{value:!0}),Object.defineProperty(_e,"default",{enumerable:!0,get:function(){return u}});var e,t=r(d[0]);
/**
   * @license lucide-react-native v0.548.0 - ISC
   *
   * This source code is licensed under the ISC license.
   * See the LICENSE file in the root directory of this source tree.
   */
```
[...]

Issue #1559, opened 2023-09-12
Import Icon as a separate module


RobertSasak:
### Package

- [ ] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [ ] lucide-react
- [X] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] all JS packages
- [ ] site

## Problem

I would like to reduce overall bundle size. React native(metro) does not support tree shaking so it is up to developer to only import modules that are needed. There is no app that need all icons and only a few are needed at the time. 

## Solution

Export all icons as a separate sub modules of the _lucide-react-native_ package.

## Workaround

It is currently possible to import separate icons from dist subfolder.
```ts
import HelpCircleIcon from 'lucide-react-native/dist/esm/icons/help-circle'
import QrCodeIcon from 'lucide-react-native/dist/esm/icons/qr-code'
```
However those do not have a type and typescript throws error
```
Could not find a declaration file for module 'lucide-react-native/dist/esm/icons/arrow-left'. 
```

### Use cases

Import only icons that are actually needed.
```ts
import HelpCircleIcon from 'lucide-react-native/dist/esm/icons/help-circle'
import QrCodeIcon from 'lucide-react-native/dist/esm/icons/qr-code'
```

### Checklist

- [X] I have searched the existing issues to make sure this bug has not already been reported.

RobertSasak:
Here is a type definition file to fix a typescript error. 
```ts
// types/lucide-react-native.d.ts
declare module 'lucide-react-native/dist/esm/icons/*' {
  import { LucideIcon } from 'lucide-react-native'

  export = LucideIcon
}
```

fmendez89:
> Here is a type definition file to fix a typescript error.
> 
> ```ts
> // types/lucide-react-native.d.ts
> declare module 'lucide-react-native/dist/esm/icons/*' {
>   import { LucideIcon } from 'lucide-react-native'
> 
>   export = LucideIcon
> }
> ```

Hi @RobertSasak ,
When I import any icon, TS error arise. For example for PlusCircleIcon. 

'PlusCircleIcon' only refers to a type, but is being used as a value here.ts(2693)

Do I need to import all icons on the type definition? Like this?

```ts
// types/lucide-react-native.d.ts
[...]

PR #3868, opened 2025-12-09
fix(lucide-react-native): remove icons namespace export to enable tree-shaking

jtomaszewski:
## Summary

Remove the `export * as icons from './icons'` namespace export from `lucide-react-native` entry files to enable proper tree-shaking.

Related: https://github.com/lucide-icons/lucide/issues/1559 , https://github.com/lucide-icons/lucide/issues/3744

## Problem

The namespace re-export defeats tree-shaking because bundlers cannot statically analyze which exports from the namespace are actually used. This causes all 1600+ icons to be included in the final bundle even when only a few are imported.

For example, in a React Native/Expo app using Metro bundler:
- **Before**: Bundle size ~10.5MB (all icons included)
- **After**: Bundle size ~8.3MB (only used icons included)

## Solution

Remove the `icons` namespace export while keeping individual icon exports available via `export * from './icons'`.

## Breaking Change

The `icons` namespace export is no longer available. Users should import icons directly:

```tsx
// Before (no longer works)
import { icons } from 'lucide-react-native';
const Icon = icons.Activity;

// After (recommended)
import { Activity } from 'lucide-react-native';
```

This is **perhaps** a minor breaking change as the namespace import pattern was rarely used and defeats the purpose of tree-shaking.

## Testing

Tested in a production Expo app with Metro bundler - bundle size reduced by ~20% after this change.


ericfennis:
Good findings, but I'm not entirely comfortable removing this feature, since it is a breaking change.

I've added the `exports` property to the `package.json` to add a path to import the icons entry into projects.
This keeps this feature, but with a different import.
Also updated the docs with the new import.

```tsx
import * as icons from 'lucide-react-native/icons';

interface IconProps {
  name: keyof typeof icons;
  color?: string;
  size?: number;
}

const Icon = ({ name, color, size }: IconProps) => {

  return <LucideIcon color={color} size={size} />;
};

export default Icon;
```

ericfennis (approved):

PR #2471, opened 2024-09-17
fix(icons): changed `*-check` icons

jguddas:
## What is the purpose of this pull request?
- [x] Other: Icon update

### Description
Increased size of check and closed gaps.

## Before Submitting
- [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [x] I've checked if there was an existing PR that solves the same issue.

PR #3875, opened 2025-12-10
fix(icons): unify check, cross and plus sizes across icons

karsa-mistmere:
closes #2471

## Description

Unifies `check` (✔︎), `cross` (×) and `plus` (+) sub-icon appearances across icons. As privately discussed in the maintainer channel on Discord regarding changes made in #2471, any main icon group should use a single set of these sub-icons so that they remain consistent:

image

E.g. either of these would be valid sets (as opposed to the current mix-and-match):

image

Previously, `circle`, `square` & `badge` used 12by12 based `plus` and `cross` elements, but 10×10 based `check`s.

All changes made in this PR and the reason:
- Bumped up the size of the checkmark in `circle-check`, `square-check` and `badge-check` so that they'd be consistent with pluses and crosses in respective icon variants.
- decreased the size of the `x` in `clipboard-x`, `volume-x`, `copy-x`, `git-pull-request-closed` to be consistent with other icons in the same group
- decreased the size of the checkmark in `clock-check` to be consistent with `clock-plus` and `clock-x`
- increased the size of the `x` in `mail-x` to be consistent with `mail-plus`
- increased the size of the checkmark in `spell-check` for better visual hierarchy and balance

## Before Submitting 


ericfennis:
Nice work.
Not sure about the changes of the cross for `git-pull-request-closed` and `volume-x` 🤔

karsa-mistmere:
> Nice work. Not sure about the changes of the cross for `git-pull-request-closed` and `volume-x` 🤔

@ericfennis: they were featured in the Figma file in #2471.

1. IMO `git-pull-request-closed` is very much a no-brainer: all of our other `git` icons use the smaller pluses, e.g. https://lucide.dev/icons/git-branch-minus, https://lucide.dev/icons/git-branch-plus, https://lucide.dev/icons/git-pull-request-create, https://lucide.dev/icons/git-pull-request-create-arrow

2. `volume-x` _is_ kind of up for debate, I guess, but: should we ever decide to add `volume-plus` and `volume-minus` (both of which are pretty basic icons imo) only the smaller `×` is going to work, in comparison, I think the smaller `x` works just fine in unison with our other volume icons and doesn't look out of place at all:

image<br>

image<br>

image

jguddas:
I prefer the bigger x's.
image

karsa-mistmere:
> I prefer the bigger x's. image

They aren't really compatible with our existing design language though, as we've already discussed this on Discord.

jguddas:
I like the 8x8 if they are placed in the center, they are quite legible, sharp and feel like a good weight compared to the surrounding icon.

It's important to not have too much space between shapes, if it's closer to 2px I like it more.

image

jguddas (changes requested):
The 9x9 x's are not sharp and feel a bit small.

jguddas:
> decreased the size of the checkmark in clock-check to be consistent with clock-plus and clock-x

Feels like it should be fixed in `clock-plus` and `clock-x`, no?

jguddas:
In the 8x8 set the `plus` feels to big, weird since actually it's smaller than a rotated x would be.
Honestly, I think it's fine, at least it's sharp that way and has a good gap weight.
image

karsa-mistmere:
> In the 8x8 set the `plus` feels to big, weird since actually it's smaller than a rotated x would be. Honestly, I think it's fine, at least it's sharp that way and has a good gap weight. image

I'm wagering a guess that you probably mean the 12×12 set, since that is what is depicted, assuming this is correct: this the plus that we already use in these contexts. While we _could_ switch to the 10x10 set globally (which is the next step down on a pixel perfect grid), but that's exactly what Eric, You and I decided _not_ to do on Discord. 🙂

> The 9x9 x's are not sharp and feel a bit small.

Again, I'm assuming you mean the 10×10 grid variants. Yeah, I wish they could be sharper, but using the same center point as `plus` by nature means that the crosses are going to be less sharp than optimal – placing them on an even sized square grid would allow them to appear much more crisp:

image<br>
image

...but alas, I really don't think this is a viable option, as using these as sub-icons would look utterly chaotic:
image

As for their size, both crosses are rounded to the nearest half pixel as their plus counterparts, and both are actually already a tiny bit larger than them:
image

> > decreased the size of the checkmark in clock-check to be consistent with clock-plus and clock-x
> 
> Feels like it should be fixed in `clock-plus` and `clock-x`, no?

No, since every single similarly themed icon uses the smaller set, so using the larger crosses and pluses would look inconsistent with Lucide as a whole.

ericfennis:
> I prefer the bigger x's. image

I'm with @jguddas on this one. The bigger X-es looks better

karsa-mistmere:
> I'm with @jguddas on this one. The bigger X-es looks better

But I'm afraid @jguddas is taking things out of context.

First of all, save for a few, our `-x` icons _already use this cross_, and not all can accommodate the larger one, e.g.:
- https://lucide.dev/icons/archive-x
- https://lucide.dev/icons/map-pin-x-inside
- https://lucide.dev/icons/search-x
- https://lucide.dev/icons/user-x

More importantly though, using a larger `x` would make these variants appear utterly inconsistent with their `-plus` and `-check` variants (compare top 3 rows [this PR] with bottom 3 rows [@jguddas's proposal]):

@ 2x:
image

@ 1x:
image

jguddas:
I have thought about this a bit more, I feel like this needs to be decided on a case by case basis.
Can we split the PR up into multiple, maybe have one for `checks` and maybe even multiple for each case of `-x`?

jguddas (changes requested):
Let's split up this PR 🙏

karsa-mistmere:
> I have thought about this a bit more, I feel like this needs to be decided on a case by case basis. Can we split the PR up into multiple, maybe have one for `checks` and maybe even multiple for each case of `-x`?

I'm sorry but this makes zero sense to me. Splitting this up would put these even further out of context. The entire point is that we need icons with checkmarks, crosses and pluses to be consistent _with each other_.

jguddas:
> that we need icons with checkmarks, crosses and pluses to be consistent with each other.

There are 25+ x-icons, to get them to feel cohesive across `folder-x`, `file-x`, … will not be doable in a single PR.
Especially since we can't use the same x-shape everywhere, due to the need of space / visual weight.

My vote is to adjust the size across variants with a target of **"Everything should feel like shape-group-\* and square-\*"** and then we have a PR were we discuss `clock-*` and if bumping up the `clock-plus` and `clock-x` is more cohesive or if we should decrease the `clock-check` instead.

Example: https://github.com/lucide-icons/lucide/pull/4004

jguddas (approved):

Issue #926, opened 2023-01-28
"Copy character" button for site


foxt:
When using Lucide as a font (i.e. TTF), it currently only compensates for usage in CSS. It'd be nice to be able to use in software such as Figma or XD. For this it'd be nice to have a "Copy character" button, similar to the font awesome website.

ericfennis:
Yeah that would be nice, but we can't do this one before we fix the "stable hex character" problem, see #514.

lutzseverino:
There is no maintainable, scalable way of creating a Figma component that displays Lucide icons if neither the font provides ligatures or the website provides a "copy character" button.

I'm trying to create one right now for a project we're working on, and this is a deal breaker. To make use of the component I'm creating, I have to tell teammates to upload the font to a font inspector, copy the hexadecimal, go to Google for a website that provides the Unicode so they can copy it... a whole hassle.

So... bump!

lutzseverino:
Will wait for a day or so, but since a bot marked this as stale, I might create a new issue.

jguddas:
Instead of creating a duplicate issue we just have to close as duplicate, feel free to create a PR instead.

The-Lost-Light:
Even displaying unicode would be much more convenient

janTatesa:
Or atleast show what's the encoded code in https://unpkg.com/lucide-static@0.541.0/font/info.json

janTatesa:
Though there is a problem that every programming language has a different syntax for unicode characters. What if there was a dropdown menu where you choose what programming language to copy the character for?

PR #3876, opened 2025-12-10
chore(dev): Enable ligatures in font build configuration

dcxo:
## Description

This PR enables support for text ligatures in the Lucide icon font by setting `addLigatures: true` in the `svgtofont` configuration. This allows icons to be referenced by their names (eg, `check-check`) ensuring more stable usage across various environments, especially non-web ones.

I've got a pretty specific use case for this: I'm planning to use Lucide in my custom Linux shell. 
The problem with using Unicode codes is: if I update the font, the Unicode values might change, which will break everything. 

Ligatures fix that, because the icon names stay the same no matter what.

This change also indirectly addresses the issue raised in #926. The original problem was the lack of a copy character button for Unicode codes. With ligatures, we no longer need to worry about copying Unicode values. Instead, users can easily copy the icon name (which is already available as a button), and it will work the same way across different languages and platforms. This eliminates the problems caused by different Unicode syntaxes in various environments.

This should not break anything, the Unicode codes are still working, so no problem for the current static web font.

## Before Submitting 

### DIFF
--- a/tools/build-font/main.ts
+++ b/tools/build-font/main.ts
@@ -101,6 +101,7 @@ async function init() {
       useCSSVars: false,
       outSVGReact: false,
       outSVGPath: false,
+      addLigatures: true,
       svgicons2svgfont: {
         fontHeight: 1000, // At least 1000 is recommended
         normalize: false,


ericfennis (commented):
@dcxo Unfortunately, this will not completely resolve the issue we have with code points in the font.
Also, this is not fixing #926 you referenced. I've removed it from the PR description

But I've checked with https://fontdrop.info, and this works great. 
Just a note, it adds +7% to the font file size. 
But I think this is a nice DX improvement since you can write the name of the icon instead of using classNames or unicode characters.

What this is not fixing is when we change an icon name, the used ligature will break, and probably will just show text.

I'm working on a different PR fixing this properly by adding support for aliases and stable codepoints.

PR #3889, opened 2025-12-14
fix(site): Fix and unify color-picker font-size

taimar:
The color-picker value was previously rendered at 14px, which caused the default light theme value (#000000) to be cut off. The "Stroke width" and "Size" inputs already use 13px, so this change aligns the color-picker value to 13px as well and resolves the cut-off issue.

Before:
before

After:
after


ericfennis (approved):
Thank you!

PR #3881, opened 2025-12-12
feat(packages): added new `@lucide/icons` package

karsa-mistmere:
## Description

This PR introduces a new `@lucide/icons` package that serves as the canonical, metadata-rich icon source and shared build API for the Lucide v1 ecosystem.

### Future-proof icon exports

Icons are now exported in an extended, structured format that exposes more information than the current children-only exports:

- icon name  – this enables consistent `lucide-<name>` CSS class generation in third-party packages and internal consumers such as `@lucide/angular`.

- **size / dimensions** – allows future support for multiple icon sizes or variants while keeping the current 24×24 default intact.

This format is intentionally **extensible** and can accommodate additional metadata in the future (e.g. variants, tags, categories) without breaking consumers.

### Dynamic icon imports

The package supports dynamic icon access, similar to the existing behaviour in `lucide-react`, enabling runtime icon selection without requiring static imports of individual icons.

### Prefixed and suffixed exports

Icons can be exported with prefixes or suffixes, mirroring the behaviour already established in `lucide-react`.

### Centralized build API

`@lucide/icons` also exposes shared helper functions for converting icon definitions into different output formats:

- `buildLucideIconNode`  
  Converts an icon definition into Lucide’s internal svgson-like node format.

- `buildLucideIconElement`  
  Produces a native [`SVGElement`](https://developer.mozilla.org/en-US/docs/Web/API/SVGElement).

- `buildLucideSvg`  
  Generates an SVG string.

- `buildLucideDataUri`  
  Generates a base64-encoded SVG data URI.

Centralising these helpers — particularly `buildLucideIconNode`—allows us to consolidate icon build logic in one place, ensuring consistent behaviour across first-party and third-party packages, and reducing duplication between framework implementations.

## Before Submitting 

- [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [x] I've checked if there was an existing PR that solves the same issue.

PR #3897, opened 2025-12-19
feat(packages/angular): add new @lucide/angular package

karsa-mistmere:
## Description

Adds new standalone-first Angular package that better matches the ecosystem established by `lucide-react`:
- one standalone component per icon
- dynamic icon component with IconNode/name input that accepts:
  - icon components
  - icon nodes from the `lucide`/`@lucide/icons` packages
  - icon names for icons provided through `provideLucideIcons` via dependency injection
- default icon configuration provided through `provideLucideConfig` via dependency injection
- support for all configuration options established in earlier packages
- support for providing an accessible label via an `<svg:title>` element. Unless this accessible label is provided, the SVGs will automatically add `aria-hidden="true"`

## TODO
- [x] Add package documentation to site
- [x] Add more documentation to code
- [x] Explore the possibility of a migration schematic from `lucide-angular`

## TODO, but requires #3881
- [ ] Add `@lucide/angular/dynamic` entry point for dynamic icon imports using `@lucide/icons`
- [ ] Add `@lucide/angular/material` entry point for [registering icon data via `MatIconRegistry`](https://material.angular.dev/components/icon/overview#registering-icons)

## Before Submitting 

### DIFF
--- a/packages/angular/MIGRATION.md
+++ b/packages/angular/MIGRATION.md
@@ -63,10 +63,10 @@ import { provideLucideIcons, LucideAirVent, LucideAlarmClock } from '@lucide/ang
 export const appConfig: ApplicationConfig = {
   providers: [
     // ...
-    provideLucideIcons([
+    provideLucideIcons(
       LucideAirVent,
       LucideAlarmClock,
-    ]),
+    ),
   ]
 };
 ```


packages/angular/README.md
```diff
@@ -0,0 +1,77 @@
...
+  ![NPM Downloads](https://img.shields.io/npm/dw/lucide-angular)
+  [![GitHub](https://img.shields.io/github/license/lucide-icons/lucide)](https://lucide.dev/license)
+</div>
+
+<p align="center">
+  <a href="https://lucide.dev/guide/">About</a>
+  ·
+  <a href="https://lucide.dev/icons/">Icons</a>
+  ·
+  <a href="https://lucide.dev/guide/packages/lucide-angular">Documentation</a>
```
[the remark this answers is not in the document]
karsa-mistmere:
```suggestion
  [![npm](https://img.shields.io/npm/v/@lucide/angular?color=blue)](https://www.npmjs.com/package/@lucide/angular)
  ![NPM Downloads](https://img.shields.io/npm/dw/@lucide/angular)
  [![GitHub](https://img.shields.io/github/license/lucide-icons/lucide)](https://lucide.dev/license)
</div>

<p align="center">
  <a href="https://lucide.dev/guide/">About</a>
  ·
  <a href="https://lucide.dev/icons/">Icons</a>
  ·
  <a href="https://lucide.dev/guide/packages/angular">Documentation</a>
```

packages/angular/src/lucide-dynamic-icon.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+    it('should not set aria-hidden when title is set', () => {
+      title.set('Foobar');
+      fixture.detectChanges();
+      expect(getSvgAttribute('aria-hidden')).toBeUndefined;
+    });
+    it('should set aria-hidden if no title is provided', () => {
+      title.set(undefined);
+      fixture.detectChanges();
+      expect(getSvgAttribute('aria-hidden')).toBeUndefined;
+    });
```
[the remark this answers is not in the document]
karsa-mistmere:
```suggestion
    it('should not set aria-hidden when title is set', () => {
      title.set('Foobar');
      fixture.detectChanges();
      expect(getSvgAttribute('aria-hidden')).toBeUndefined();
    });
    it('should set aria-hidden if no title is provided', () => {
      title.set(undefined);
      fixture.detectChanges();
      expect(getSvgAttribute('aria-hidden')).toBeUndefined();
    });
```

packages/angular/src/lucide-icons.ts
```diff
@@ -0,0 +1,64 @@
+import { InjectionToken, Provider } from '@angular/core';
+import { LucideIconData, LucideIcons } from './types';
+import { isLucideIconComponent, LucideIconComponentType } from './types';
+import { toKebabCase } from './utils/to-kebab-case';
+
+/**
+ * Injection token for providing Lucide icons by name.
+ *
+ * @internal Use {@link provideLucideConfig}
+ */
```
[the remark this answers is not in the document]
karsa-mistmere:
```suggestion
 * @internal Use {@link provideLucideIcons}
 */
```

packages/angular/src/lucide-icons.ts
```diff
@@ -0,0 +1,64 @@
...
+        acc[toKebabCase(icon.iconName)] = icon.iconData;
+        return acc;
+      }, {} as LucideIcons),
+    };
+  } else {
+    return {
+      provide: LUCIDE_ICONS,
+      useValue: Object.entries(icons).reduce((acc, [name, icon]) => {
+        if (isLucideIconComponent(icon)) {
+          acc[icon.iconName] = icon.iconData;
```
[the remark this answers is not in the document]
karsa-mistmere: New icon data `iconName` props are already kebab case. The entire "converting to kebab case" methodology is somewhat murky, maybe we should only accept new icon data format from `@lucide/icons` here, that already contains the unique name of the icon.

packages/angular/tsconfig.json
```diff
@@ -0,0 +1,38 @@
...
+    "strictTemplates": true,
+  },
+  "files": [],
+  "references": [
+    {
+      "path": "./tsconfig.lib.json",
+    },
+    {
+      "path": "./tsconfig.spec.json",
+    },
```
karsa-mistmere, line 36:
```suggestion
      "@lucide/angular": ["./dist"]
    },
    "strict": true,
    "noImplicitOverride": true,
    "noPropertyAccessFromIndexSignature": true,
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": true,
    "skipLibCheck": true,
    "isolatedModules": true,
    "declaration": false,
    "downlevelIteration": true,
    "experimentalDecorators": true,
    "importHelpers": true,
    "target": "ES2022",
    "module": "preserve"
  },
  "angularCompilerOptions": {
    "enableI18nLegacyMessageIdFormat": false,
    "strictInjectionParameters": true,
    "strictInputAccessModifiers": true,
    "strictTemplates": true
  },
  "files": [],
  "references": [
    {
      "path": "./tsconfig.lib.json",
    },
    {
      "path": "./tsconfig.spec.json",
    }
```

packages/angular/src/lucide-dynamic-icon.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+      const hostFixture = TestBed.createComponent(TestHostComponent);
+      hostFixture.componentRef.setInput('icon', testIcon);
+      hostFixture.detectChanges();
+      hostFixture.componentRef.setInput('icon', testIcon2);
+      hostFixture.detectChanges();
+      const rect = hostFixture.debugElement.query(By.css('rect')).nativeElement;
+      expect(rect).toBeInstanceOf(SVGElement);
+      expect(rect.outerHTML).toBe('<rect x="1" y="1" width="22" height="22"></rect>');
+    });
+  });
```
ericfennis, line 131: Can you also write a test that the local props made to the icon are overwriting the globally defined properties on the ContextProvider

packages/angular/src/lucide-icon-base.ts
```diff
@@ -0,0 +1,146 @@
...
+  selector: 'svg[lucideIcon]',
+  templateUrl: './lucide-icon.html',
+  host: {
+    ...defaultAttributes,
+    class: 'lucide',
+    '[attr.width]': 'size().toString(10)',
+    '[attr.height]': 'size().toString(10)',
+    '[attr.stroke]': 'color()',
+    '[attr.stroke-width]': 'computedStrokeWidth()',
+    '[attr.aria-hidden]': '!title()',
```
ericfennis, line 44: This should also check if there is any `aria-label` attribute is present.

ericfennis (commented):
Nice work @karsa-mistmere, this looks very promising!
Some comments

ericfennis:
@karsa-mistmere FYI, updated from `next` branch, fixed conflicts and failing checks

ericfennis:
@karsa-mistmere Any updates on this? Can I help with something?
I assume I need to fix the "next" tag deployment of `@lucide/icons` right?

karsa-mistmere:
> @karsa-mistmere Any updates on this? Can I help with something? I assume I need to fix the "next" tag deployment of `@lucide/icons` right?

Yeah, but first, I still want to add the aliases in `@lucide/icons`, as we discussed, I'm on it already!

packages/angular/src/types.ts
```diff
@@ -0,0 +1,47 @@
...
+export type LucideIconComponentType = Type<{
+  title: Signal<Nullable<string>>;
+  size: Signal<Nullable<number>>;
+  color: Signal<Nullable<string>>;
+  strokeWidth: Signal<Nullable<number>>;
+  absoluteStrokeWidth: Signal<Nullable<boolean>>;
+}> & {
+  iconName: string;
+  iconData: LucideIconData;
+};
```
ericfennis:
```suggestion
export interface LucideProps {
  title: Signal<Nullable<string>>;
  size: Signal<Nullable<number>>;
  color: Signal<Nullable<string>>;
  strokeWidth: Signal<Nullable<number>>;
  absoluteStrokeWidth: Signal<Nullable<boolean>>;
}

/**
 * Represents a Lucide icon component that has `iconName` and `iconData` signals inherited from `LucideIconBase` and respective static members accessible without instantiating the component.
 */
export interface LucideIcon extends Type<LucideProps> {
  iconName: string;
  iconData: LucideIconData;
};
```

ericfennis: Does this also supports other svg and html attributes out of the box, or does it need more typing?

packages/angular/src/lucide-icon-base.ts
karsa-mistmere: In Angular components there isn't a generic "props object" you can statically introspect at runtime the way you might in React (`{...props}` + conditional logic on `props['aria-label']`). Attribute bindings are **explicit and compiled ahead of time**.

Because of this – and since `aria-label` (or other `aria-*` attributes) isn't currently exposed as an input on this component – there isn't anything reliable to check in component code unless we start introducing additional explicit inputs.

Even then, Angular does not provide a synchronous, framework-supported way to introspect arbitrary bound attributes that might be present on a component instance. The only alternative would be DOM inspection (e.g. via `ElementRef`), which would add unnecessary complexity and would add quite performance-heavy runtime overhead for what should be a simple, predictable and performant component.

If we want to support accessible labeling, the Angular-native approach is to:
- accept `title` (and/or `ariaLabel`) as an explicit input, and
- set `aria-hidden` based on those explicit signals.

So the current `aria-hidden: !title()` is intentionally simple and deterministic: if you provide a title, it's exposed to assistive tech; if you don't, it's treated as decorative.

packages/angular/src/types.ts
karsa-mistmere: Not automatically — and that's by design in Angular.

As also noted above, Angular doesn't provide a "spread props onto an element" mechanism in templates. Every bindable thing needs to be declared and bound explicitly (`[attr.foo]`, `[class.bar]`, etc).

This isn't just a typing concern — it's fundamental to how Angular's template compiler and binding system operate. There's no equivalent of React's `{...rest}` pattern that would forward arbitrary inputs or attributes to the underlying `<svg>` element.

While we certainly _could_ add more supported attributes, each one has to be an explicit API decision + explicit binding (signal inputs in particular are statically declared and strongly shaped; they cannot be instantiated dynamically). Modelling the component as “accept arbitrary attributes and forward them” would be:
- not idiomatic Angular (templates are intentionally explicit), and
- at odds with signal-based inputs, which are designed for explicit, strongly typed bindings rather than catch-all attribute bags

---

Regarding the naming: the package already contains a separate `LucideIcon` that represents the dynamic icon component. It would be reasonable to rename that to `LucideDynamicIcon` (while keeping the `svg[lucideIcon]` selector for simplicity's sake).

That would make the distinction clearer:
- `LucideIcon` → the type of a generated icon component (static metadata + typed inputs)
- `LucideDynamicIcon` → the runtime renderer/directive that selects an icon by name

packages/angular/src/lucide-dynamic-icon.spec.ts
karsa-mistmere: By ContextProvider, I assume you mean the `LUCIDE_CONFIG` injection token, right?

I added some tests on the usage itself (although there were already unit tests on the provider function itself).

karsa-mistmere:
@ericfennis: added some extensive rewrites:

- as per https://github.com/lucide-icons/lucide/pull/3897#discussion_r2640158469, I switched from separate node+name to self-describing `LucideIconData` objects similar to `@lucide/icons` – so no more toKebabCase hackery anywhere in the code apart from some legacy helpers, and as an added bonus there is now backwards compatible alias support for name based icon providers 🚀
- as per https://github.com/lucide-icons/lucide/pull/3897#discussion_r2840064045 I renamed `LucideIconComponentType` => `LucideIcon`, and `LucideIcon` => `LucideDynamicIcon`
- I added backwards compatible CSS class support as already implemented in `@lucide/icons`
- I switched to the `vector-effect: non-scaling-stroke` implementation from using computed stroke widths
- as requested, I added some unit tests on `LUCIDE_CONFIG` provider usage

packages/angular/src/types.ts
ericfennis: Thanks for the explanation, makes sense.

ericfennis:
@karsa-mistmere Awesome work!

PR #3920, opened 2025-12-26
fix(site): add missing titles and a title template

taimar:
Multiple pages in the Lucide docs don’t define a unique page title, which makes open browser tabs hard to distinguish – it’s unclear whether you’re looking at the same page or different ones. For example, both the home page and the icons page show “Lucide” as the title.

This PR adds the missing titles. It also introduces a title template, replacing the default pipe character (`|`) with an en-dash (`–`) for clearer separation between the page name and the site name, and a nicer overall look.

Before:
before screenshot showing tab name for icons page: Lucide

After:
after screenshot showing tab name for icons page: Icons – Lucide


ericfennis (approved):
Nice thank you!

PR #3922, opened 2025-12-29
feat(lucide-react): expose SVG stroke attributes as typed props

omghante:
This PR exposes commonly-used SVG stroke attributes as first-class typed props in `lucide-react`.
### What's Added
| Prop | Type | Purpose |
|------|------|---------|
| `strokeLinecap` | `'butt' \| 'round' \| 'square'` | Line end shape |
| `strokeLinejoin` | `'bevel' \| 'miter' \| 'round'` | Corner join shape |
| `strokeDasharray` | `string \| number` | Dashed/dotted patterns |
| `strokeDashoffset` | `string \| number` | Dash pattern offset |
| `opacity` | `number` | Overall transparency |
| `strokeOpacity` | `number` | Stroke-only transparency |
### Motivation
These are valid SVG attributes already supported via rest props, but adding explicit typing:
- **Improves TypeScript autocomplete** and discoverability
- **Documents** the most common customization patterns
- **Reduces friction** for developers
### Design Choices
- **No new behavior** — pure pass-through to SVG
- **No validation** — browser handles invalid values  
- **No default changes** — fully backward compatible
- **Conditional spreading** — props only applied when explicitly provided
### Example Usage
```tsx
import { Heart } from 'lucide-react';
// Dashed stroke
<Heart strokeDasharray="5 5" />
// Semi-transparent icon
<Heart opacity={0.5} />
// Square line caps instead of round
<Heart strokeLinecap="square" />
```
### Files Changed
packages/lucide-react/src/types.ts — Added typed props with JSDoc
packages/lucide-react/src/Icon.ts — Added conditional prop spreading
packages/lucide-react/tests/Icon.strokeAttributes.spec.tsx — Added unit tests

## Before Submitting 

### DIFF
--- a/packages/lucide-react/src/Icon.ts
+++ b/packages/lucide-react/src/Icon.ts
@@ -32,6 +32,12 @@ const Icon = forwardRef<SVGSVGElement, IconComponentProps>(
       className = '',
       children,
       iconNode,
+      strokeLinecap,
+      strokeLinejoin,
+      strokeDasharray,
+      strokeDashoffset,
+      opacity,
+      strokeOpacity,
       ...rest
     },
     ref,
@@ -47,6 +53,13 @@ const Icon = forwardRef<SVGSVGElement, IconComponentProps>(
         strokeWidth: absoluteStrokeWidth ? (Number(strokeWidth) * 24) / Number(size) : strokeWidth,
         className: mergeClasses('lucide', className),
         ...(!children && !hasA11yProp(rest) && { 'aria-hidden': 'true' }),
+        // Pass through SVG stroke attributes only when explicitly provided
+        ...(strokeLinecap && { strokeLinecap }),
+        ...(strokeLinejoin && { strokeLinejoin }),
+        ...(strokeDasharray !== undefined && { strokeDasharray }),
+        ...(strokeDashoffset !== undefined && { strokeDashoffset }),
+        ...(opacity !== undefined && { opacity }),
+        ...(strokeOpacity !== undefined && { strokeOpacity }),
         ...rest,
       },
       [
--- a/packages/lucide-react/src/types.ts
+++ b/packages/lucide-react/src/types.ts
@@ -23,6 +23,43 @@ type ElementAttributes = RefAttributes<SVGSVGElement> & SVGAttributes;
 export interface LucideProps extends ElementAttributes {
   size?: string | number;
   absoluteStrokeWidth?: boolean;
+
+  /**
+   * SVG stroke-linecap attribute.
+   * Controls the shape at the end of open subpaths.
+   */
+  strokeLinecap?: 'butt' | 'round' | 'square';
+
+  /**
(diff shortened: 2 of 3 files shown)


jguddas:
Nice but also a little bit dangerous.

karsa-mistmere:
These props are not supported for a reason, see [our icon design principles.](https://lucide.dev/guide/design/icon-design-guide)

jguddas:
> These props are not supported for a reason, see [our icon design principles.](https://lucide.dev/guide/design/icon-design-guide)

Aren't we supporting overriding those props?

karsa-mistmere:
> > These props are not supported for a reason, see [our icon design principles.](https://lucide.dev/guide/design/icon-design-guide)
> 
> Aren't we supporting overriding those props?

Stroke width, yes. Line cap and join no.

omghante:
> These props are not supported for a reason, see [our icon design principles.](https://lucide.dev/guide/design/icon-design-guide)


i will try to fix it as per the principles

omghante:
> Nice but also a little bit dangerous.

i am curious about it will u let me know that why it is dangerous....?

karsa-mistmere:
> > Nice but also a little bit dangerous.
> 
> i am curious about it will u let me know that why it is dangerous....?

As far as I'm concerned, it would allow – or worse, encourage – people to instantiate non-compliant icons, and I really don't think we want that.

> > These props are not supported for a reason, see [our icon design principles.](https://lucide.dev/guide/design/icon-design-guide)
> 
> i will try to fix it as per the principles

To be honest, since we only have strokes as is, support for stroke opacity separately makes no sense, as it renders inferior to overall opacity, and stroke dashes could also result in non-compliant designs, so I think the only one that would actually make sense for us is `opacity`, but that one should already be passed along as is in `...rest`.

jguddas:
Okay, I did a little test, the types are already there.

image

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcARFDvmQNwBQokscA3nAMLBS4A2WAkrggA7OAF84REnADkPAK65gAEywBaSnhgz6dLAA8m8IcNTwAovuTg+cALxwAFAEp7APjgAeTtz6CRcGbEANZYADLAwli4yGB2ZGRwAPRuQA

karsa-mistmere:
> Okay, I did a little test, the types are already there.
> 
> image
> https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcARFDvmQNwBQokscA3nAMLBS4A2WAkrggA7OAF84REnADkPAK65gAEywBaSnhgz6dLAA8m8IcNTwAovuTg+cALxwAFAEp7APjgAeTtz6CRcGbEANZYADLAwli4yGB2ZGRwAPRuQA

Yes, that's what I said. 🙂

> `opacity` [...] should already be passed along as is in `...rest`.

PR #3940, opened 2026-01-04
fix(icons): renamed `triangle-right` to `right-angled-triangle`

jguddas:
## Description

Renamed icon, this makes it so this icon isn't in the same bucked as the other `triangle` icons, this isn't a variant of the icon `triangle`, so naming it like this should make it less confusing.

## Before Submitting 


karsa-mistmere:
Still, according to our naming conventions, modifiers should come after the shape, so `triangle-right-angled` would be the correct name.

jguddas:
> Still, according to our naming conventions, modifiers should come after the shape, so `triangle-right-angled` would be the correct name.

That's my point, it doesn't really feel like it's a modified triangle icon, it feels separate, maybe we should fix that.

karsa-mistmere:
> > Still, according to our naming conventions, modifiers should come after the shape, so `triangle-right-angled` would be the correct name.
> 
> That's my point, it doesn't really feel like it's a modified triangle icon, it feels separate, maybe we should fix that.

It is first and foremost a triangle, so that would still be a base name, even if it's not _the_ `triangle`.

PR #3943, opened 2026-01-04
fix(icons): renamed `text-select` to `square-dashed-text`

jguddas:
## Description

Renamed icon.

## Before Submitting 


karsa-mistmere:
@jguddas `square-dashed-text.json` has linting issues. :(

danielbayley (approved):

ericfennis (approved):
Approved, but linting issues should be fixed

matyasf:
Note: Such renames might break downstream applications. Would it be possible, that such renames are released as a breaking change warranting a new major version release? Thanks!

karsa-mistmere:
> Note: Such renames might break downstream applications. Would it be possible, that such renames are released as a breaking change warranting a new major version release? Thanks!

I'm not quite sure what breaking changes you mean. We do add backwards compatible aliases for each of these renames, so any previous exports will remain usable.

The only thing that should change between versions is CSS classes will change from `lucide lucide-text-select` to `lucide lucide-square-dashed-text`.

PR #3977, opened 2026-01-11
fix(icons): changed `gift` icon

jguddas:
- [x] Other: Icon update

### Description
Increased size.

## Before Submitting

### DIFF
--- a/icons/gift.json
+++ b/icons/gift.json
@@ -4,7 +4,8 @@
     "colebemis",
     "csandman",
     "ericfennis",
-    "jguddas"
+    "jguddas",
+    "karsa-mistmere"
   ],
   "tags": [
     "present",


icons/gift.svg
```diff
@@ -9,8 +9,8 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <rect x="3" y="8" width="18" height="4" rx="1" />
   <path d="M12 8v13" />
```
jamiemlaw: Should be `<path d="M12 7v14" />` to keep the gap closed after the resize

ericfennis (approved):

icons/gift.svg
karsa-mistmere, line 1: I think it feels a bit too wide now. 🤔

Also, with the lid staying the same height but becoming much wider, it creates a bit of an optical illusion where the lid ends up _looking_ wider than the box compared to the previous iteration — even though the proportions are technically closer.

What do you think about slightly shrinking the lid horizontally, something like this?

Image<br>

Image

<a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=%3Cpath+d=%22M12+7v14%22+/%3E%3Cpath+d=%22M20+11v8a2+2+0+0+1-2+2H6a2+2+0+0+1-2-2v-8%22+/%3E%3Cpath+d=%22M7.5+7a1+1+0+0+1+0-5A4.8+8+0+0+1+12+7a4.8+8+0+0+1+4.5-5+1+1+0+0+1+0+5%22+/%3E%3Crect+x=%223%22+y=%227%22+width=%2218%22+height=%224%22+rx=%221%22+/%3E&name=gift">icons<br/>Open lucide studio</a>

PR #1848, opened 2024-02-03
feat(icon): add UserKey icon

mittalyashu:
## What is the purpose of this pull request?

- [x] New Icon
- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other:

### Description

A user icon with key

### Icon use case

It can be used to show user authentication or passkey

### Alternative icon designs

CleanShot 2024-02-03 at 18 47 56@2x

### Concept 

- [x] I have provided valid use cases for each icon.
- [x] I have not added any a brand or logo icon.
- [x] I have not used any hate symbols.
- [x] I have not included any religious or political imagery.

### Author, credits & license

- [x] The icons are solely my own creation.
- [ ] The icons were originally created in #<issueNumber> by @<githubUser>
- [x] I've based them on the following Lucide icons: 
- [ ] I've based them on the following design: 

### Naming 

- [ ] I've read and followed the [naming conventions](https://lucide.dev/guide/design/icon-design-guide#naming-conventions)
- [ ] I've named icons by what they are rather than their use case.
- [x] I've provided meta JSON files in `icons/[iconName].json`.

### Design 

- [ ] I've read and followed the [icon design guidelines](https://lucide.dev/guide/design/icon-design-guide)
- [ ] I've made sure that the icons look sharp on low DPI displays.
- [ ] I've made sure that the icons look consistent with the icon set in size, optical volume and density.
- [ ] I've made sure that the icons are visually centered.
- [ ] I've correctly optimized all icons to two points of precision.

## Before Submitting 

- [ ] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [ ] I've checked if there was an existing PR that solves the same issue.

Issue #4041, opened 2026-01-29
Passkey icon


lode:
### Icon name

passkey, or user-key

### Use cases

It would be nice to have a icon for usage next to passkey flows.

According to https://www.passkeycentral.org/design-guidelines/principles a user with a key is the main icon they advice to use.

### Design ideas

<img width="900" height="900" alt="Image" src="https://github.com/user-attachments/assets/48cb923a-7dcb-4d8e-945d-14ff0c6f5dd1" />

And in style of Lucide, could look a lot like `user-search`:

<img width="24" height="24" alt="Image" src="https://github.com/user-attachments/assets/36e5527d-5c26-45b2-b363-c81d64ccc6bf" />

### Checklist

- [x] I have searched if someone has submitted a similar issue before and there weren't any. (Please make sure to also search closed issues, as this issue might already have been resolved.)
- [x] I have searched existing icons to make sure it does not already exist and I didn't find any.
- [x] I am not requesting a brand logo and the art is not protected by copyright, see more at https://github.com/lucide-icons/lucide/blob/main/BRAND_LOGOS_STATEMENT.md
- [x] I am not requesting an icon that includes religious, war/violence related, political imagery or hate symbols.
- [x] I have provided appropriate use cases for the icon(s) requested.

Rayyan-101:
I put together a first draft concept for a `passkey` icon in Lucide style.

The design follows the recommended “user + key” metaphor from the passkey guidelines and is visually close to existing icons like `user-search`, to stay consistent with the Lucide set.

I’d love to get feedback on the direction before opening a PR.

Preview:

<img width="459" height="452" alt="Image" src="https://github.com/user-attachments/assets/4d73e11e-feb3-4e39-9ea3-4f6f07b4d15a" />

karsa-mistmere:
We had a `user-key` PR in #1848 that was closed automatically when @mittalyashu deleted his feature branch, I've now posted a PR including all proposed changes in that PR.

PR #4044, opened 2026-01-30
feat(icons): adds `user-key` and `user-round-key`, updates other `-key` icons to match

karsa-mistmere:
closes #1848 #4041

- [x] New Icon
- [x] Other: Icon update

### Description
- Added new `user-key` & `user-round-key` icons.
- Updated `-key` icons with designs based on #1848

### Icon use case

See #4041

### Alternate design ideas
image<br>

image

### Concept

### Author, credits & license

- [x] The icons were originally created in #1848
- [x] I've based them on the following Lucide icons: `user`, `user-round`

### Naming

### Design

## Before Submitting


lode:
I've read the previous PR discussion on the direction of the icon, but I think the outcome is not good. The key is now pointing upwards, while the official recommendation is a key pointing downwards. I think relating to how others are doing this is more important than specific visual esthetics.

lode:
See:
- https://learn.microsoft.com/en-us/windows/apps/develop/security/design
- https://www.passkeycentral.org/design-guidelines/
- https://developers.google.com/identity/passkeys/ux/user-interface-design
- https://web.dev/articles/passkey-google-ux
- https://www.yubico.com/blog/new-fido-alliance-design-guidelines-key-takeaways-for-passkeys-and-a-phishing-resistant-user-future/

karsa-mistmere:
> while the official recommendation is a key pointing downwards

Can you point me to this exact recommendation? I can't find it anywhere in your links.

We could flip it upside down, except then `rotate-ccw-key` would actually become `rotate-cw-key`:

image

Personally, I still prefer the diagonal key, it feels way more consistent with our base key icons.

lode:
> Can you point me to this exact recommendation? I can't find it anywhere in your links.

Sorry, maybe my wording is a bit off. I don't think there's anything that "official". But these design recommendations from all the bigger parties that are behind webauthn all state something similar to 'stick to the standards'. For example in https://www.passkeycentral.org/design-guidelines/principles:

> Use proven passkeys messaging and icons before and after OS dialogs.
> Before triggering the passkeys OS dialogs, display passkeys messages, icons, and actions related to the status of the current task.

Further you can just see it in the design they all use, always the same. Clearly to help end users understand it is the same thing.

It is a bit like the [share icon](https://lucide.dev/icons/share-2). It would be weird if we designed that rotated 180 degrees spreading out to the left instead of to the right.

> Personally, I still prefer the diagonal key, it feels way more consistent with our base key icons.

I think the same applies. But also, especially when shown as a small icon, I get a big weird associations with that variant, as if it is the persons hand sticking up.

Also, diagonal isn't always what's done here. Again, I see following conventions is what's done. For example in [an editing pen with top _right_ to left bottom](https://lucide.dev/icons/user-round-pen) but the [search magnifying glass with top _left_ to bottom right](https://lucide.dev/icons/user-round-search).

karsa-mistmere:
> > Personally, I still prefer the diagonal key, it feels way more consistent with our base `key` icons.
> 
> I think the same applies. But also, especially when shown as a small icon, I get a big weird associations with that variant, as if it is the persons hand sticking up.
> 


> Also, diagonal isn't always what's done here. Again, I see following conventions is what's done. For example in [an editing pen with top _right_ to left bottom](https://lucide.dev/icons/user-round-pen) but the [search magnifying glass with top _left_ to bottom right](https://lucide.dev/icons/user-round-search).

Yes, and our `pen` and `search` icons do in fact match these exact orientations. :) 

_That_ was the point I was trying to make with "being consistent with our base `key` icons".

lode:
Thanks! I understand what you mean.

I still think this icon should be different, alike share, this is more about aligning with others.

A bit ugly maybe, but another solution is two icons. One following the direction of the keys (`user-key`), and one specific for passkeys (`user-key-round` or `user-passkey`).

Spleefies:
I prefer the diagonal keys.

ericfennis (approved):

PR #3553, opened 2025-08-17
fix(icons): changed `receipt-text` icon

jguddas:
## What is the purpose of this pull request?
- [x] Other: Icon update

### Description
Arcified icon.

## Before Submitting
- [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [ ] I've checked if there was an existing PR that solves the same issue.

PR #4075, opened 2026-02-17
fix(icons): changed `receipt` icon

karsa-mistmere:
- [x] Other: Icon update

### Description
Updates outlines with the one used by `receipt-text` since #3553 

## Before Submitting


karsa-mistmere:
(Not a fan of these existing at all, but if they do, they should be consistent in style)

PR #4135, opened 2026-03-11
fix(astro): add Astro v6 compatibility

iseraph-dev:
## Problem
`@lucide/astro` does not support Astro v6, which was released alongside Lucide v1. Installing the package in an Astro v6 project results in unmet peer dependency warnings, as the peer dependency range only allows `^4 || ^5`.

## Summary
- Updated `peerDependencies` to accept Astro v6: `"astro": "^4 || ^5 || ^6"`
- Updated `devDependencies` for post-v1 compatibility:
  - `astro` → `^6.0.8` (test against latest major)
  - `vite` → `^7.3.1` (required by Astro v6)
  - `vitest` → `^4.1.1` (compatible with Vite 7)

No runtime code changes were needed — the Astro Container API used in the components remains unchanged across v4/v5/v6.

## Test plan
- [x] All 14 existing tests pass against Astro v6 with Vite 7
- [x] Runtime code uses only stable `astro/compiler-runtime` APIs, compatible with v4/v5/v6
- [x] No breaking changes in Astro v6 affect icon component rendering

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -57,8 +57,9 @@
     "linkedom": "^0.18.5",
     "prettier": "^3.4.2",
     "typescript": "^5.8.3",
-    "vite": "^6.3.6",
-    "astro": "^6.0.0"
+    "vite": "^7.3.1",
+    "vitest": "^4.1.1",
+    "astro": "^6.0.8"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


fkrzski (approved):

iseraph-dev:
@ericfennis Would it be possible to merge this into the upcoming release?

PR #4193, opened 2026-03-23
ci(ci.yml): Fix release flow

ericfennis:
* Updated the package list in the release workflow to use simplified package names (e.g., `angular`, `astro`, `svelte`, `icons`, `vue`) instead of scoped or commented-out package names. This makes the matrix easier to read and maintain.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,7 +86,7 @@ jobs:
         working-directory: packages/${{ matrix.package }}
 
       - name: Publish
-        run: pnpm publish --access public --no-git-checks --ignore-scripts --tag ${{ inputs.tag }} --access public
+        run: pnpm publish --access public --no-git-checks --ignore-scripts --tag ${{ inputs.tag }}
         working-directory: packages/${{ matrix.package }}
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -76,16 +74,20 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Set new version
-        run: pnpm --filter ${{ matrix.package }} version --new-version ${{ needs.prepare.outputs.VERSION }} --no-git-tag-version
+        run: pnpm version --new-version ${{ needs.prepare.outputs.VERSION }} --no-git-tag-version
```
karsa-mistmere, line 77: TBH, I'd consider keeping these steps as they are, the current version feels like a more standard monorepo approach, and instead just add `--fail-if-no-match`.

We may even only need to add it to the first pnpm command, if that one fails, the rest won't even execute:

```suggestion
        run: pnpm --fail-if-no-match --filter ${{ matrix.package }} version --new-version ${{ needs.prepare.outputs.VERSION }} --no-git-tag-version
```

PR #4201, opened 2026-03-24
fix(github/workflows): revert release workflow & add --fail-if-no-match

karsa-mistmere:
Reverted some changes made in #4193 

Also added `"analytics": false` to `packages/angular/angular.json` to prevent prompt about it during `pnpm build`.

## Before Submitting 

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -83,7 +83,7 @@ jobs:
         run: pnpm --filter ${{ matrix.package }} test
 
       - name: Publish
-        run: pnpm --filter ${{ matrix.package }} publish --access public --no-git-checks --ignore-scripts --tag ${{ inputs.tag }} --access public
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -74,20 +74,16 @@ jobs:
...
 
       - name: Test
-        run: pnpm test
-        working-directory: packages/${{ matrix.package }}
+        run: pnpm --filter ${{ matrix.package }} test
 
       - name: Publish
-        run: pnpm publish --access public --no-git-checks --ignore-scripts --tag ${{ inputs.tag }}
-        working-directory: packages/${{ matrix.package }}
+        run: pnpm --filter ${{ matrix.package }} publish --access public --no-git-checks --ignore-scripts --tag ${{ inputs.tag }} --access public
```
ericfennis: double `--access public`

```suggestion
        run: pnpm --filter ${{ matrix.package }} publish --access public --no-git-checks --ignore-scripts --tag ${{ inputs.tag }}
```

PR #4234, opened 2026-03-31
fix(@lucide/svelte): aria-hidden="true" was never set

blt-r:
When using icon components like `<AirVent />` the aria-hidden attribute was never set. 

That is because in `Icon.svelte` we have `!children && !hasA11yProp(props) && { 'aria-hidden': 'true' }`, so if children prop is present (and is truthy), the attribute is not set. But in the generated icon components there was:

```svelte
<Icon name="air-vent" {...props} iconNode={iconNode}>
  {@render props.children?.()}
</Icon>
```

Because of how svelte works this results in `Icon` receiving some `children` prop even if this `children` in outer component was undefined. Even though that snippet is empty in terms of markup, it's still present, so the `!children` check fails. The solution is to just do this:

```svelte
<Icon name="air-vent" {...props} iconNode={iconNode} />
```

Because `children` is simply a prop like any other it will already be forwarded, no need to do that explicitly.

I've also added tests for that and updated the snapshots for the tests, as previously they were testing for incorrect behavior. 

## Before Submitting 

### DIFF
--- a/packages/svelte/scripts/exportTemplate.mts
+++ b/packages/svelte/scripts/exportTemplate.mts
@@ -36,8 +36,6 @@ const iconNode: IconNode = ${JSON.stringify(children)};
  */
 </script>
 
-<Icon name="${iconName}" {...props} iconNode={iconNode}>
-  {@render props.children?.()}
-</Icon>
+<Icon name="${iconName}" {...props} iconNode={iconNode} />
 `;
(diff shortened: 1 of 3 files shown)


blt-r:
Hey, @ericfennis sorry for the ping. Can you take a look at this please

PR #4235, opened 2026-03-31
fix(angular): copy package.json to dist after build

hfroget-305:
## Description

Fixes `ERR_PNPM_NO_IMPORTER_MANIFEST_FOUND` in CI by ensuring `package.json` exists in `packages/angular/dist/` after the Angular build completes.

The `publishConfig.directory` is set to `dist`, which makes pnpm look for a manifest there. However, `ng-packagr` does not always produce one, causing the build to fail. This PR adds a `postbuild` script that copies `package.json` into `dist/` and chains it into the `build` command.

### Changes
- Added `"postbuild": "cp package.json dist/package.json"` script
- - Updated `"build"` to run `pnpm postbuild` after `pnpm build:ng`


karsa-mistmere:
> ## Description
> Fixes `ERR_PNPM_NO_IMPORTER_MANIFEST_FOUND` in CI by ensuring `package.json` exists in `packages/angular/dist/` after the Angular build completes.
> 
> The `publishConfig.directory` is set to `dist`, which makes pnpm look for a manifest there. However, `ng-packagr` does not always produce one, causing the build to fail. This PR adds a `postbuild` script that copies `package.json` into `dist/` and chains it into the `build` command.
> 
> ### Changes
> * Added `"postbuild": "cp package.json dist/package.json"` script
> * * Updated `"build"` to run `pnpm postbuild` after `pnpm build:ng`

Hey @hfroget-305, thanks for taking the time to look into this. 🙇‍♂️

However, this issue has actually already been resolved in #4213. The issue wasn't related to `ng-packagr` at all, but rather `pnpm clean` accidentally being executed between the `build` and `publish` steps, which removed the `dist` folder altogether.

So I don't think it's necessary to introduce the additional `postbuild` step here. Also, just to note, `ng-packagr` _should_ consistently generate a `package.json` in `dist`, so it would be good to double-check that assumption.

Thanks again for the contribution and digging into this, but I'll be closing this as as duplicate of #4213

Issue #4244, opened 2026-04-01
LucideProvider size prop should accept string | number to match individual icon props


daedalus28:
### Bug description

`LucideProps.size` accepts `string | number`, but `LucideConfig` (used by `LucideProvider` via `Partial<LucideConfig>`) types `size` as just `number`. This means you can't pass string values like `"1em"` to the provider without a type cast, even though:

1. Individual icons accept `size="1em"` just fine
2. The runtime implementation passes the value straight through to context with no coercion — strings work correctly

The [global styling docs](https://lucide.dev/guide/react/advanced/global-styling) recommend using `LucideProvider` over CSS specifically so that individual icons can still override via props. A common use case is setting `size="1em"` globally so icons scale with surrounding text, but TypeScript rejects this:

```tsx
// TS2322: Type 'string' is not assignable to type 'number'
<LucideProvider size="1em">
```

### Workaround

```tsx
<LucideProvider size={'1em' as unknown as number}>
```

### Expected behavior

`LucideConfig.size` should be `string | number` to match `LucideProps.size`:

```diff
 type LucideConfig = {
-    size: number;
+    size: string | number;
     color: string;
     strokeWidth: number;
     absoluteStrokeWidth: boolean;
     className: string;
 };
```

### Version

`lucide-react@1.7.0`

swastik7805:
I've looked into this and the fix is straightforward — `LucideConfig.size` needs to be `string | number` to match `LucideProps.size`. Also noticed `color` was typed as `string | number` which is incorrect since color values are always strings.

```ts
size: string | number;
color: string;
```

Raised a PR to bring `lucide-react` in sync: #4253

daedalus28:
@karsa-mistmere pinging you on this issue since you commented on the PR. This definitely seems like a rough edge of the API.

karsa-mistmere:
> [@karsa-mistmere](https://github.com/karsa-mistmere) pinging you on this issue since you commented on the PR. This definitely seems like a rough edge of the API.

The PR response stands, so the PR in and of itself does not solve the issue, we're going to need more time.

In the meantime I suggest using contextual CSS for rem based sizes.

PR #4253, opened 2026-04-02
fix(packages/angular): allow string inputs for size

swastik7805:
Closes issues #4288 #4287

## Description
Allows strokeWidth to be specific as float string, e.g.:

Root Bug:

```tsx
export interface LucideConfig {
  /**
   * Stroke color.
   * @default currentColor
   */
  color: string;
  /**
   * Width and height.
   * @default 24
   */
  size: number;
```

Previously this would get interpreted as `1`.

Core update (other iclude `lucide-icon-base.spec.ts` , `lucide-icon-base.ts` and  `types.ts`):

```tsx
export interface LucideConfig {
  /**
   * Stroke color.
   * @default currentColor
   */
  color: string;
  /**
   * Width and height.
   * @default 24
   */
  size: number | string;
```

This unblocks the recommended global styling pattern from the docs:

Note. This also allows `size` to be specified as float, which is not useful.
But I thought that there is no harm in that, so I didn't guard against it.

## Before Submitting

### DIFF
--- a/packages/angular/src/lucide-config.ts
+++ b/packages/angular/src/lucide-config.ts
@@ -13,7 +13,7 @@ export interface LucideConfig {
    * Width and height.
    * @default 24
    */
-  size: number;
+  size: string | number;
   /**
    * Stroke width
    * @default 2
--- a/packages/angular/src/lucide-icon-base.spec.ts
+++ b/packages/angular/src/lucide-icon-base.spec.ts
@@ -113,12 +113,6 @@ describe('LucideIconBase', () => {
       expect(getSvgAttribute('width')).toBe('18');
       expect(getSvgAttribute('height')).toBe('18');
     });
-    it('should use default on invalid string', () => {
-      size.set('large');
-      fixture.detectChanges();
-      expect(getSvgAttribute('width')).toBe('24');
-      expect(getSvgAttribute('height')).toBe('24');
-    });
   });
 
   describe('strokeWidth', () => {
@@ -132,9 +126,9 @@ describe('LucideIconBase', () => {
       expect(getSvgAttribute('stroke-width')).toBe('1.41');
     });
     it('should allow string stroke width', () => {
-      strokeWidth.set('1px');
+      strokeWidth.set('0.125rem');
       fixture.detectChanges();
-      expect(getSvgAttribute('stroke-width')).toBe('1');
+      expect(getSvgAttribute('stroke-width')).toBe('0.125rem');
     });
   });
 
--- a/packages/angular/src/lucide-icon-base.ts
+++ b/packages/angular/src/lucide-icon-base.ts
@@ -13,20 +13,6 @@ import { LucideIconData, Nullable } from './types';
 import defaultAttributes from './default-attributes';
 import { lucideIconTemplate } from './lucide-icon-template';
 
-function transformNumericStringInput(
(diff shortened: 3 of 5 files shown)


karsa-mistmere:
Hey @SwastikIIIT,

For `@lucide/angular`, I see no harm in allowing string inputs for `size` and `strokeWidth`, I took the liberty to update the Angular code base to fix some leftover issues in https://github.com/lucide-icons/lucide/pull/4253/changes/ad55057fcbd06fc85700a50ed228c04519a52669

For `lucide-react` and our other packages however, life is not so easy, since we currently use the `size` prop to calculate the stroke width when using `absoluteStrokeWidth=true`, see https://github.com/lucide-icons/lucide/blob/main/packages/lucide-react/src/Icon.ts#L41

Using a `rem` value here will not work, since `Number('1rem')` will yield `NaN`.

I'd suggest reverting the changes made to `lucide-react` for the time being, until we migrate all packages from using these calculated stroke widths to the implementation in `@lucide/angular` (setting `vector-effect: non-scaling-stroke` on child elements).

daedalus28:
@karsa-mistmere The reason for adding this was #4244 which is specifically about typing in `lucide-react`'s provider. Perhaps we can just document that passing in a number doesn't work with `absoluteStrokeWidth`? If `size` was set via css (which is also in the docs), you'd still have the same issue, right? Size can also be passed as `1em` at the individual icon level which I imagine would have this issue as well.

I think the key takeaway is that either we're saying lucide react just doesn't support non px size values (in which case the typing on the individual icons should change too) or we should let it through on the provider.

swastik7805:
Hi @daedalus28 ,
My changes on lucide-react pckage for allowing string as well number input in size works with issue stated in #4244.
Can u check and pass ur comment on it?

daedalus28:
@swastik7805 It looks good to me, but I am not familiar with the lucide codebase 😅

swastik7805:
hi @karsa-mistmere ,
so the issue #4244 will not be resolved till all the packages are migrated as in `package/lucide-react` size is used to calculate `strokeWidth`,therefore,changing type of `size` in that package may interfere with other functionalities.

karsa-mistmere:
> @karsa-mistmere The reason for adding this was #4244 which is specifically about typing in `lucide-react`'s provider. Perhaps we can just document that passing in a number doesn't work with `absoluteStrokeWidth`? If `size` was set via css (which is also in the docs), you'd still have the same issue, right? Size can also be passed as `1em` at the individual icon level which I imagine would have this issue as well.
> 
> I think the key takeaway is that either we're saying lucide react just doesn't support non px size values (in which case the typing on the individual icons should change too) or we should let it through on the provider.

Icons currently allow string inputs not because we want support for sizes with explicit units, but to support numeric strings, e.g. `size="48"` instead of `size={48}`.

Documenting that `absoluteStrokeWidth` and non-numeric strings sizes are not compatible would certainly be an option, but I think we'd be better off just migrating to `vector-effect: non-scaling-stroke`, which does work OOTB with `rem` sizes.

karsa-mistmere:
Reopened this, since it does still contain valuable fixes for `@lucide/angular` that I pushed earlier, see https://github.com/lucide-icons/lucide/pull/4253#issuecomment-4182277923

jguddas (approved):

PR #4267, opened 2026-04-04
fix(@lucide/svelte): proper doc comments for svelte components

blt-r:
The doc comments for the icon components like `<AirVent />` are broken. The actual doc comment with information about the icon and preview is not present, instead there's just license in doc comment.

The reason why the real doc comment wasn't present is because of incorrect syntax. In this PR I swapped it for proper svelte html-style doc comment.

The reason why the license was present in the doc comment is because it was inserted into the reexporting file and jsdoc parser applied it to the `export { default } from "./air-vent.svelte";` item. To fix that I added jsdoc `@file` declaration to the license doc comment so that it applied to the file itself.

In `appendBlockComments.mts` there was also some old code leftover from Svelte 4 trying to add jsdoc comment to the component declaration. It did'n work and wasn't needed so I removed it.

## Before/after:

before-affter

## P.S.
It seemed very wasteful to include entire text of the license in every single file in the `./dist`. I checked, and other packages (react, preact, vue, solid) just have:

```javascript
/**
 * @license package-name v0.0.1 - ISC
 *
 * This source code is licensed under the ISC license.
 * See the LICENSE file in the root directory of this source tree.
 */
```

Should I change the svelte package to do the same?

## Before Submitting 

### DIFF
--- a/packages/svelte/scripts/appendBlockComments.mts
+++ b/packages/svelte/scripts/appendBlockComments.mts
@@ -7,6 +7,8 @@ import { getJSBanner } from './license.mts';
 const currentDir = getCurrentDirPath(import.meta.url);
 const targetDirectory = path.join(currentDir, '../dist');
 
+const jsBanner = getJSBanner();
+
 const files = await readdir(targetDirectory, {
   recursive: true,
   encoding: 'utf-8',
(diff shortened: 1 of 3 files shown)


blt-r:
Hey @karsa-mistmere can you take a look at this please

ericfennis (approved):
You're right, Thanks for fixing this. I've adjusted the text!

PR #4269, opened 2026-04-04
fix(site): ensure Carbon Ad widget dismisses correctly

### LINKED ISSUE #3999: Close (X) icon is not clickable / does not work on lucide.dev website
### Browser

- [ ] Chrome/Chromium
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [ ] Other/not relevant

### Operating system

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [ ] Other/not relevant

### Description

Close (X) icon is not clickable / does not work on lucide.dev website

**Expected behavior**

Clicking on the X (close) icon should close or dismiss the floating widget.
Actual behavior
The icon is visible and styled as clickable, but clicking it does nothing.

**Environment**

OS: Windows 10
Browser: Firefox 
Screen size: Desktop
Reproducible: Always

**Additional context**
This can be confusing for users, as the close icon follows common UI patterns but does not perform the expected action.

Screenshots attached for reference.

Image

### Steps to reproduce

1. Open https://lucide.dev/icons
2. Look at the floating widget in the bottom-right corner of the screen
3. Click on the X (close) icon
4. Nothing happens — the widget is not closed

### Checklist

swastik7805:
Closes #3999 

## Description
This PR fixes a bug on the `lucide.dev` website where the floating Carbon Ad widget in the bottom-right corner could not be dismissed by clicking the Close (X) icon.

### The Problem
The issue stemmed from how JavaScript evaluates truthiness combined with session storage serialization. 
The ad was hidden using the following inline logic: `:class="{ 'hide-ad': !(showAd && carbonLoaded) }"`

However, when the `showAd` state was saved to `sessionStorage`, the boolean `false` was sometimes serialized and retrieved as the string `"false"`. Because non-empty strings are truthy in JavaScript, the string `"false"` evaluated to `true`. This caused the hide condition to fail, leaving the ad stuck on the screen despite user interaction.

### The Solution
Instead of adding complex workarounds to the template logic, this PR fixes the root cause at the storage level. 

By explicitly passing `{ serializer: StorageSerializers.boolean }` to the `useSessionStorage` hook, we ensure the value is correctly parsed back into a strict boolean (`true`/`false`) rather than a string. This elegantly resolves the truthiness issue and allows the existing inline logic to work perfectly.

**Changes made:**
* Updated `useSessionStorage` in `CarbonAdOverlay.vue` to use `StorageSerializers.boolean`.
* Cleaned up/reverted unnecessary import reordering.
### Evidence

https://github.com/user-attachments/assets/607bceeb-b67d-4d24-9467-18d479d58831

## Before Submitting

### DIFF
--- a/docs/.vitepress/theme/components/icons/CarbonAdOverlay.vue
+++ b/docs/.vitepress/theme/components/icons/CarbonAdOverlay.vue
@@ -5,13 +5,13 @@ import IconButton from '../base/IconButton.vue';
 import VPDocAsideCarbonAds from 'vitepress/dist/client/theme-default/components/VPDocAsideCarbonAds.vue';
 import { x } from '../../../data/iconNodes';
 import Icon from '@lucide/vue/src/Icon';
-import { onMounted, ref } from 'vue';
+import { computed, onMounted, ref } from 'vue';
 
 const { theme } = useData();
-const showAd = useSessionStorage('show-carbon-ads', true, {
-  serializer: StorageSerializers.boolean,
-});
+const showAd = useSessionStorage('show-carbon-ads', true);
 const carbonLoaded = ref(true);
+const isHidden = computed(() => !showAd.value || !carbonLoaded.value);
+const closeAd = () => showAd.value = false;
 
 defineProps<{
   drawerOpen: boolean;
@@ -30,13 +30,13 @@ onMounted(() => {
   <div
     :class="{
       'drawer-open': drawerOpen,
-      'hide-ad': !(showAd && carbonLoaded),
+      'hide-ad': isHidden,
     }"
     class="floating-ad"
     v-if="theme.carbonAds"
   >
     <IconButton
-      @click="showAd = false"
+      @click="closeAd"
       class="hide-button"
     >
       <Icon
@@ -99,6 +99,7 @@ onMounted(() => {
   position: absolute;
   top: 8px;
   right: 8px;
+  z-index: 3;
   background-color: transparent;
 }
 </style>


swastik7805:
Hi team (@karsa-mistmere  @jguddas  @ericfennis )! 
The fix for the Carbon Ad close button is up. I've included a video in the PR body so you can see the fix in action. Let me know if everything looks good or if any tweaks are needed. Thanks

docs/.vitepress/theme/components/icons/CarbonAdOverlay.vue
```diff
@@ -5,12 +5,21 @@ import IconButton from '../base/IconButton.vue';
 import VPDocAsideCarbonAds from 'vitepress/dist/client/theme-default/components/VPDocAsideCarbonAds.vue';
 import { x } from '../../../data/iconNodes';
 import Icon from '@lucide/vue/src/Icon';
-import { onMounted, ref } from 'vue';
+import { onMounted, ref, computed } from 'vue';
+
 
 const { theme } = useData();
 const showAd = useSessionStorage('show-carbon-ads', true);
```
karsa-mistmere: Hey @swastik7805,

Thanks for looking into this, the issue really is the lack of serialization, but the proper solution would be to just do this:

```suggestion
const showAd = useSessionStorage('show-carbon-ads', true, {
  serializer: StorageSerializers.boolean,
});
```

Then `isHidden` and `closeAd` are not necessary, and the rest of the code can be restored as it originally was.

swastik7805: @karsa-mistmere  thanks providing this alternative fix.I have updated the PR .Its ready for review.

swastik7805:
Hi @karsa-mistmere ,
I have incorporated the changes requested by u.
Could you please review and merge this when you have a moment?

karsa-mistmere:
Yeah, @swastik7805, I'm afraid this doesn't solve the issue after all, at least the error persists on the latest deployment: https://lucide-micddsqzo-lucide.vercel.app/icons/  😅

swastik7805:
> Yeah, @swastik7805, I'm afraid this doesn't solve the issue after all, at least the error persists on the latest deployment: https://lucide-micddsqzo-lucide.vercel.app/icons/ 😅

😲yup u are right.Both custom functions and serialisation is not working for this issue.
Need to dive deep into it to figure out the exact solution.

karsa-mistmere:
> > Yeah, @swastik7805, I'm afraid this doesn't solve the issue after all, at least the error persists on the latest deployment: https://lucide-micddsqzo-lucide.vercel.app/icons/ 😅
> 
> 😲yup u are right.Both custom functions and serialisation is not working for this issue. Need to dive deep into it to figure out the exact solution.

Hey @swastik7805,

I've managed to pinpoint the source of error in #4275 (sorry, I needed to test it in a Vercel deployment), so I'm closing this as a duplicate.

PR #4305, opened 2026-04-20
feat: add new package lucide-taro-base64 for Taro inline SVG workaround

jacket-sikaha:
## Title

feat: add new package `lucide-taro-base64` for Taro inline SVG workaround

### Summary

This PR introduces a new official package: **`lucide-taro-base64`**  
It solves the **inline SVG unsupported issue** in Taro applications, specifically optimized for **`taro-plugin-html`**.

### Problem

Taro **does not support native inline `<svg />` tags**, so the official `lucide-taro` package cannot render icons normally.

### Solution

`lucide-taro-base64` is implemented **by following the exact API design of `lucide-react`**:

- Rewrote the core `createLucideIcon` function
- Each icon is a **fully functional React component**
- Dynamically generates SVG and converts it to a **DataURL (base64 alternative)** at runtime
- Renders to an `` tag under the hood
- Works seamlessly with `taro-plugin-html`: the plugin converts `` to Taro components and renders via DataURL

### Implementation Highlights

- **Same API as `lucide-react` & `lucide-taro`** (drop-in replacement)
- **Dynamic props support**: `color`, `size`, `strokeWidth`, `fill`, `className`, `style`, etc.
- **ForwardRef support** for component ref
- **TypeScript fully typed**
- **Tree-shakable**
- **Zero extra dependencies**
- **Consistent icon names** with all Lucide packages

### Usage Example

```tsx
import { View } from '@tarojs/components';
import { ArrowBigLeft, CookingPot, Square, User } from 'lucide-taro-base64';
import './index.less';

export default function Index() {
  return (
    <View className="index">
      <div className="text-[#acc855] text-[100px]">Hello world!</div>
      <div className="flex w-full flex-col items-center">
        <User
          color="blue"
          className="size-10 block"
        />
        <CookingPot
          color="red"
          className="size-52 block"
          strokeWidth={3}
        />
        <ArrowBigLeft
          color="#c8559e"
          className="size-32 block"
          fill="#ef4444"
        />
        <div className="flex w-2/3 items-center justify-center gap-2 rounded-full border-2 border-red-500 bg-transparent p-4 transition-all hover:bg-red-500/5 active:scale-95">
          <Square
            style={{ width: 32, height: 32 }}
            color="#ef4444"
            fill="#ef4444"
          />
          <div className="text-base font-medium text-red-500">停止</div>
        </div>
      </div>
    </View>
  );
}
```

### How it works internally

1. Icon component receives props
2. `createLucideIcon` generates SVG string dynamically
3. Converts to `data:image/svg+xml` DataURL
4. Renders as ``
5. `taro-plugin-html` transforms `` into Taro compatible components

### Compatibility

- Taro 3+
- taro-plugin-html
- All Taro platforms (H5, WeChat Mini Program, App, etc.)

---


karsa-mistmere:
Hey @jacket-sikaha,

Thanks a lot for putting this together — you've clearly put quite a bit of thought and effort into this.

That said, I don't believe this is something we'd want to bring into the core Lucide repository.

The main reason is that this is a fairly platform-specific workaround (Taro + taro-plugin-html + DataURL rendering), and we try to keep the core packages focused on broadly applicable, canonical implementations. Solutions like this tend to be better suited as community-maintained packages, where they can evolve independently without us having to support a very niche runtime constraint long-term.

If you'd like to publish and maintain this as a separate package, we're more than happy for you to do so — and we can even point people your way when similar use cases come up.

PR #4320, opened 2026-04-24
docs(readme): Update readme files

### LINKED ISSUE #4316: The switch from `lucide-vue-next` to `@lucide/vue` isn't communicated clearly.
The npm download numbers quite clearly show that not many users have made the switch from the old `lucide-vue-next` to the new `@lucide/vue` npm package, most likely because they weren't properly informed about the change.

Not only is there no obvious info or warning about the change anywhere, but even the README still features only the old package.

ericfennis:
Closes #4316 
Closes #4306

## Changes:
- Update badges that not worked or outdated
- Updates the outdated list of package names
- Update some text


A-Loot:
It might also make sense to add a note or warning to the old package to inform users about the change.

e.g.:

> [!IMPORTANT]
> This package has been moved to `@lucide/vue`. `lucide-vue-next` will no longer receive any updates.

ericfennis:
@A-Loot Yeah we will do that. We've decided to wait a bit longer for this. We also some other packages we need to deprecate.

PR #4326, opened 2026-04-27
fix(icons): changed `option` icon

jamiemlaw:
- [x] Other: Icon update

### Description
Arcified corners

## Before Submitting

### DIFF
--- a/icons/option.svg
+++ b/icons/option.svg
@@ -9,6 +9,6 @@
   stroke-linecap="round"
   stroke-linejoin="round"
 >
-  <path d="M3 3h6l6 18h6" />
   <path d="M14 3h7" />
+  <path d="M3 3h5.28a1 1 0 0 1 .948.684l5.544 16.632a1 1 0 0 0 .949.684H21" />
 </svg>


karsa-mistmere:
Maybe we could bump it up to even 2px and grid-align it? 🤔

<a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=%3Cpath+d=%22M15+3h6%22+/%3E%3Cpath+d=%22M3+3h4a2+2+0+0+1+1.854+1.248l6.292+15.504A2+2+0+0+0+17+21h4%22+/%3E&name=option">icons<br/>Open lucide studio</a>

jamiemlaw:
> Maybe we could bump it up to even 2px and grid-align it? 🤔

I was toying with the idea of whether to align the arcs, or to keep the lines as sharp as possible. For short lines, I typically side with the former; for long lines, the latter. In this case, I leant towards line sharpness.

I hadn't considered a 2px radius because I had intended to match the curvature on `bold`, `type`, etc.

But looking at your updated design, I like it.

I'll leave the PR open a little longer in case there's more feedback, but I think you made a good point.

ericfennis (approved):
I'm fine with both.

ericfennis:
To move on, I'm merging this one

jguddas (approved):

PR #4348, opened 2026-05-06
feat(icons): added sticky note variants

Barakudum:
## Description

Adds a couple [`sticky-note`](https://lucide.dev/icons/sticky-note) variants.

- `sticky-note-plus`
  <a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=%3Cpath%20d%3D%22M15%203v5a1%201%200%200%200%201%201h5%22%20%2F%3E%3Cpath%20d%3D%22M18%2015v6%22%20%2F%3E%3Cpath%20d%3D%22M21%2011.5V9a2.4%202.4%200%200%200-.706-1.706l-3.588-3.588A2.4%202.4%200%200%200%2015%203H5a2%202%200%200%200-2%202v14a2%202%200%200%200%202%202h6.5%22%20%2F%3E%3Cpath%20d%3D%22M21%2018h-6%22%20%2F%3E">icons</a>
- `sticky-note-minus`
  <a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=%3Cpath%20d%3D%22M15%203v5a1%201%200%200%200%201%201h5%22%20%2F%3E%3Cpath%20d%3D%22M21%2014V9a2.4%202.4%200%200%200-.706-1.706l-3.588-3.588A2.4%202.4%200%200%200%2015%203H5a2%202%200%200%200-2%202v14a2%202%200%200%200%202%202h6.5%22%20%2F%3E%3Cpath%20d%3D%22M21%2018h-6%22%20%2F%3E">icons</a>
- `sticky-note-check`
  <a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=%3Cpath%20d%3D%22m15%2019%202%202%204-4%22%20%2F%3E%3Cpath%20d%3D%22M15%203v5a1%201%200%200%200%201%201h5%22%20%2F%3E%3Cpath%20d%3D%22M21%2013V9a2.4%202.4%200%200%200-.706-1.706l-3.588-3.588A2.4%202.4%200%200%200%2015%203H5a2%202%200%200%200-2%202v14a2%202%200%200%200%202%202h6%22%20%2F%3E">icons</a>
- `sticky-note-x`
[...]

PR #4360, opened 2026-05-12
feat(metadata): Adds support for `$extends` property in meta file

ericfennis:
## Introduction

Currently, we define all sorts of information about icons in metajson files.
But we have a lot of duplicated content. For example, `alarm-clock-minus` has some of the same tags as `alarm-clock` and `minus`. This creates a lot of overhead, defining them over and over again. For example, in #4348.

## Added new properties

- `$extends.tags` - inherit tags from defined icons
- `$extends.categories` - inherit categories from defined icons
- `$extends.contributors` - inherit contributors from defined icons
- `$extends` - inherit tags, categories and contributors from defined icons, not sure if this one is very useful

##How to Use:##

```json
{
  "name": "alarm-clock-minus",
  "contributors": [
      "mittalyashu",
      "ericfennis",
      "jguddas",
      "karsa-mistmere"
   ],
  "$extends.contributors": ["alarm-clock"],
  "tags": [
     "combined",
   ],
  "$extends.tags": ["alarm-clock", "minus"],
  "categories": [
     "devices",
  ],
  "$extends.categories": ["alarm-clock"]
}
```


karsa-mistmere:
Neat idea, it'd reduce the burden of having to duplicate super long lists of tags, although I'd probably approach this like seen below, it feels a bit cleaner to edit and maintain:

```json
{
  "name": "alarm-clock-minus",
  "contributors": [
      "$extends:alarm-clock",
      "mittalyashu",
      "ericfennis",
      "jguddas",
      "karsa-mistmere"
   ],
  "tags": [
     "$extends:alarm-clock",
     "$extends:minus",
     "combined"
   ],
  "categories": [
     "$extends:alarm-clock",
     "devices"
  ]
}
```

We'll have to be careful about introducing circular references though. 🙂

ericfennis:
@karsa-mistmere Jup, exactly, there are also multiple syntax possibilities we can take. I like your idea as well.

Another idea was an object, but I didn't like that since it was close to the defined properties
```json
{
    $extends: {
       "tags: [],
       "categories": [],
       "contributors": []
    }
}
```

jguddas (changes requested):
I don't like this, not KISS, now if I change one file it might have invisible down stream effects, I would rather have duplication.
My suggestion would be to move away from the tags entirely and migrate towards using use case embeddings for most things.

We could even just auto generate the tags using AI based on the use case.

Strong vote for moving towards the new use case field.
And maybe add a visual description field as well.

karsa-mistmere:
> I don't like this, not KISS, now if I change one file it might have invisible down stream effects, I would rather have duplication. My suggestion would be to move away from the tags entirely and migrate towards using use case embeddings for most things.
> 
> We could even just auto generate the tags using AI based on the use case.
> 
> Strong vote for moving towards the new use case field. And maybe add a visual description field as well.

I agree with part of this. Tags that primarily describe use cases often lack sufficient context, so moving those towards richer use-case descriptions makes sense.

However, I'd be cautious about dropping support for tags entirely, as we currently also use them to store synonyms or alternative names/interpretations (e.g. users/group/people or rat/mouse/mice/gerbil/rodent).

I think that this latter use doesn't naturally fit into use-case descriptions. Trying to include every synonym and alternate name in prose will often result in convoluted, repetitive, or even tautological use cases.

I wonder if it would make sense to split the current tags field into something like `synonyms` (or `alt-names`) and `use-cases`, with each serving a single, well-defined purpose.

jguddas:
> I wonder if it would make sense to split the current tags field into something like synonyms (or alt-names) and use-cases, with each serving a single, well-defined purpose.

We already have aliases, maybe it fits in there somehow?

---

For the extends syntax, it would be easier for me from a tooling perspective to add an optional `extendsContributors?: string[]` field.
That way we don't break any contracts, for people like me who already display contributors, and thus would have failing UI with `contributors: ["$extends:house", "jguddas"]` failing to render, `Error: no such user "$extends:house"`.

ericfennis:
@jguddas So my initial implementation 😅

jguddas:
Also maybe, no extending contributors, if I edit an icon, an icon that in the past branched of should not get my name on it as my new changes aren't in it

PR #4393, opened 2026-05-20
feat(site): persist home page customizer settings in localStorage

### LINKED ISSUE #4358: Remember customizer settings on website
### Package

- [ ] lucide
- [ ] lucide-angular (old version)
- [ ] @lucide/angular (new version)
- [ ] @lucide/astro
- [ ] lucide-flutter
- [ ] lucide-preact
- [ ] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-static
- [ ] lucide-svelte (old version)
- [ ] @lucide/svelte (new version)
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] lucide-astro
- [ ] @lucide/icons
- [ ] Figma plugin
- [ ] all JS packages
- [x] site
- [ ] other/not relevant

### Description

Image

It would be a great UX improvement for my workflow if these settings were stored and restored upon returning to the website.

### Use cases

I am using Lucide icons for a game project in Godot, and the way I use it is by going to the site, searching for the icon I need and downloading the SVG. However, I often forget to set the customization options I need before downloading. Since the default is 24px size, the icons are super pixellated - which means going back and redownloading.

PS: I tested the Godot Lucide asset someone made, but I find that downloading the entire set and keeping it in the repo is overkill for my use.

### Checklist

reory:
closes #4358

## Description
This PR resolves a UX issue where user customization choices (Color, Stroke Width, Size, and Absolute Stroke Width) on the home page customizer would reset upon a page refresh or navigation change.

By replacing the standard reactive states with `useLocalStorage` from `@vueuse/core`, these configurations now gracefully persist across sessions.

## Before Submitting - [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).

### DIFF
--- a/docs/.vitepress/theme/components/home/HomeIconCustomizer.vue
+++ b/docs/.vitepress/theme/components/home/HomeIconCustomizer.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { ref, watch } from 'vue'
-import { syncRef, useCssVar } from '@vueuse/core'
+import { syncRef, useCssVar, useLocalStorage } from '@vueuse/core'
 import HomeContainer from './HomeContainer.vue'
 import RangeSlider from '../base/RangeSlider.vue'
 import InputField from '../base/InputField.vue'
@@ -11,10 +11,10 @@ import Switch from '../base/Switch.vue'
 
 
 const iconContainer = ref<HTMLElement | null>()
-const color = ref('currentColor')
-const strokeWidth = ref(2)
-const size = ref(24)
-const absoluteStrokeWidth = ref(false)
+const color = useLocalStorage('lucide-custom-color', 'currentColor')
+const strokeWidth = useLocalStorage('lucide-custom-strokeWidth', 2)
+const size = useLocalStorage('lucide-custom-size', 24)
+const absoluteStrokeWidth = useLocalStorage('lucide-custom-absoluteStroke', false)
 
 const colorCssVar = useCssVar(
   '--customize-color',
--- a/docs/.vitepress/theme/components/home/HomeIconCustomizer.vue
+++ b/docs/.vitepress/theme/components/home/HomeIconCustomizer.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { ref, watch } from 'vue'
-import { syncRef, useCssVar, useLocalStorage } from '@vueuse/core'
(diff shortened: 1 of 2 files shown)


karsa-mistmere:
Hey @reory,

I think there may have been a slight misunderstanding here.

The original issue in #4358 proposed persisting the customizer settings _on the [Icons page](https://lucide.dev/icons/)￼._

This PR currently persists the settings for the home page demo instead, which is a different component/workflow.

Personally, I'm not sure there's much value in persisting the home page demo settings, since they don't affect the downloaded assets in any way.

reory:
Hi @karsa-mistmere 
Ahh I see, let me work on this misunderstanding right away and get a fix.
I have reverted the homepage changes and moved the `useLocalStorage` logic directly into the `useIconStyle.ts`. I also tested the icons locally, the stroke slider and the size and it works well and importantly persists across refreshes on the main `/icons/` directory as requested.

Let me know if this looks good to you. 
Thanks.

reory:
@karsa-mistmere 
Hi, I have updated the branch and fixed the Prettier formatting issues that were causing the linter to fail. Ready for another review, when you have time.

docs/.vitepress/theme/components/home/HomeIconCustomizer.vue
```diff
@@ -9,8 +9,9 @@ import ResetButton from '../base/ResetButton.vue'
 import HomeIconCustomizerIcons from './HomeIconCustomizerIcons.vue'
 import Switch from '../base/Switch.vue'
 
-
 const iconContainer = ref<HTMLElement | null>()
+
+// Reverted back to standard refs so they don't persist on refresh
```
karsa-mistmere: Let's revert these changes:

```suggestion

const iconContainer = ref<HTMLElement | null>()
```

karsa-mistmere (changes requested):
Hey @reory,

Thanks again for working on this!

I tested the PR locally, and there still seem to be a couple of issues:

1. `absoluteStrokeWidth` does not appear to persist after refresh.
2. On page refresh, the default values are rendered first, then replaced by the `localStorage` values a moment later. This causes a very visible state jump/flicker, that I think we should try to avoid if possible.

https://github.com/user-attachments/assets/523b45e7-203d-422d-9e14-7f6adb265eb8

PR #4409, opened 2026-05-22
feat(packages): extract icon build logic into `@lucide/shared`

### LINKED ISSUE #4564: @lucide/vue: `aria-hidden` is missing for icons
### Package

- [ ] lucide
- [ ] lucide-react
- [x] @lucide/vue
- [ ] @lucide/svelte
- [ ] @lucide/angular
- [ ] @lucide/astro
- [ ] lucide-react-native
- [ ] lucide-preact
- [ ] lucide-solid
- [ ] lucide-static
- [ ] @lucide/icons
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

1.25.0

- [x] Yes
- [ ] No

### Browser

- [ ] Chrome/Chromium
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [ ] Other/not relevant

### Operating system

- [ ] Windows
- [ ] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [ ] Other/not relevant

### Description

It seems that the `aria-hidden` property is missing from the source code.
Vue example (first code block): https://lucide.dev/guide/vue/basics/color
Any other example (first code block): https://lucide.dev/guide/lucide/basics/color

### Steps to reproduce

1. Import any icon
2. Check HTML output in devtools

### Checklist

karsa-mistmere:
## Description

Fixes #4564

This MR consolidates duplicated icon-build logic into `@lucide/shared` and aligns package APIs around a shared icon contract.

- Moved duplicated icon build logic into `@lucide/shared`
- This enables backwards compatible CSS class support for all packages (adding `lucide-{alias}` classes for all aliases)
- Updated packages to use shared build/types utilities
- Standardised `Icon` usage around `icon={iconData}` (with `iconNode` fallback where needed)
- Added `nonScalingStroke` support across affected packages
- Deprecated `absoluteStrokeWidth` (in favour of `nonScalingStroke`)
- Aligned d.ts/tsconfig setup for shared type resolution
- Updated tests/snapshots for new behaviour and API changes

## Before merge
- [ ] We should decide whether the vanilla `lucide` package should be updated, although this would likely come with some breaking changes. Alternatively, we could develop a new `@lucide/vanilla` package that uses the shared build utilities & exposes much of the same API as our framework packages
- [ ] `@lucide/svelte` currently adds an extra `.lucide-icon` CSS class on icons. This is kept, but I think we should consider dropping this to be in line with all the rest of the packages.
- [x] `@lucide/vue` currently adds `lucide lucide-{name.toCamel.ToKebab}-icon lucide-{name}-icon` CSS classes, this PR introduces a breaking change by adding the same CSS classes as in all other packages (i.e. dropping the `lucide-{name.toCamel.ToKebab}-icon` one), since I deemed class lists that would otherwise appear to be simply too much, e.g.:
   - `lucide lucide-circle-question-mark lucide-circle-question-mark-icon lucide-help-circle lucide-help-circle-icon lucide-circle-help lucide-circle-help-icon`
   - `lucide lucide-circle-arrow-out-down-right lucide-circle-arrow-out-down-right-icon lucide-arrow-down-right-from-circle lucide-arrow-down-right-from-circle-icon`

## Before Submitting 


jguddas (approved):

jguddas (approved):

packages/lucide-react-native/src/Icon.ts
```diff
@@ -44,39 +74,51 @@ const Icon = forwardRef<SVGSVGElement, IconComponentProps>(
...
+      color: color ?? contextColor,
+      width: width ?? size ?? contextSize,
+      height: height ?? size ?? contextSize,
+      strokeWidth: strokeWidth ?? contextStrokeWidth,
+      absoluteStrokeWidth: absoluteStrokeWidth ?? contextAbsoluteStrokeWidth,
+      nonScalingStroke: nonScalingStroke ?? contextNonScalingStroke,
+      className: mergeClasses(contextClass, className),
+      hasA11yProp: hasAccessibleProp,
+      attributes: rest,
+    });
```
ericfennis, line 97: Let's wrap this in a `useMemo`.
```suggestion
    const [, svgAttributes, builtIconNode = []] = useMemo(() = > buildLucideIconForReact(icon, {
      color: color ?? contextColor,
      width: width ?? size ?? contextSize,
      height: height ?? size ?? contextSize,
      strokeWidth: strokeWidth ?? contextStrokeWidth,
      absoluteStrokeWidth: absoluteStrokeWidth ?? contextAbsoluteStrokeWidth,
      nonScalingStroke: nonScalingStroke ?? contextNonScalingStroke,
      className: mergeClasses(contextClass, className),
      hasA11yProp: hasAccessibleProp,
      attributes: rest,
    }), [...deps]);
```

packages/lucide-preact/src/Icon.ts
```diff
@@ -18,49 +24,56 @@ interface IconComponentProps extends LucideProps {
...
+    color: color ?? contextColor,
+    width: width ?? size ?? contextSize,
+    height: height ?? size ?? contextSize,
+    strokeWidth: strokeWidth ?? contextStrokeWidth,
+    absoluteStrokeWidth: absoluteStrokeWidth ?? contextAbsoluteStrokeWidth,
+    nonScalingStroke: nonScalingStroke ?? contextNonScalingStroke,
+    className: mergeClasses(contextClass, classes as string),
+    hasA11yProp: Boolean(children) || hasA11yProp(rest),
+    attributes: rest,
+  });
```
ericfennis, line 71:
```suggestion
  const [name, svgAttributes, builtIconNode = []] = useMemo(() => buildLucideIconNode(icon, {
    color: color ?? contextColor,
    width: width ?? size ?? contextSize,
    height: height ?? size ?? contextSize,
    strokeWidth: strokeWidth ?? contextStrokeWidth,
    absoluteStrokeWidth: absoluteStrokeWidth ?? contextAbsoluteStrokeWidth,
    nonScalingStroke: nonScalingStroke ?? contextNonScalingStroke,
    className: mergeClasses(contextClass, classes as string),
    hasA11yProp: Boolean(children) || hasA11yProp(rest),
    attributes: rest,
  }), [..deps]);
```

packages/lucide-react/src/Icon.ts
```diff
@@ -19,45 +26,66 @@ interface IconComponentProps extends LucideProps {
...
+      color: color ?? contextColor,
+      width: width ?? size ?? contextSize,
+      height: height ?? size ?? contextSize,
+      strokeWidth: strokeWidth ?? contextStrokeWidth,
+      absoluteStrokeWidth: absoluteStrokeWidth ?? contextAbsoluteStrokeWidth,
+      nonScalingStroke: nonScalingStroke ?? contextNonScalingStroke,
+      className: mergeClasses(contextClass, className),
+      hasA11yProp: hasAccessibleProp,
+      attributes: rest,
+    });
```
ericfennis, line 79: This one as well
```suggestion
    const [name, svgAttributes, builtIconNode = []] = useMemo(() => buildLucideIconForReact(icon, {
      color: color ?? contextColor,
      width: width ?? size ?? contextSize,
      height: height ?? size ?? contextSize,
      strokeWidth: strokeWidth ?? contextStrokeWidth,
      absoluteStrokeWidth: absoluteStrokeWidth ?? contextAbsoluteStrokeWidth,
      nonScalingStroke: nonScalingStroke ?? contextNonScalingStroke,
      className: mergeClasses(contextClass, className),
      hasA11yProp: hasAccessibleProp,
      attributes: rest,
    }), [...deps]);
```

packages/lucide-solid/src/Icon.tsx
```diff
@@ -1,66 +1,59 @@
...
 
+  const hasAccessibleProp = Boolean(localProps.children) || hasA11yProp(rest);
+
+  const icon: LucideIconData = localProps.icon ?? {
+    node: localProps.iconNode ?? ([] as LucideIconNode[]),
+    size: 24,
+    aliases: [],
+  };
+
+  const [, svgAttributes, builtIconNode = []] = buildLucideIconNode(icon, {
```
ericfennis: We need to test if this is still reactive in Solid.

packages/vue/src/Icon.ts
```diff
@@ -29,54 +60,44 @@ const Icon: FunctionalComponent<LucideProps & IconProps> = (
...
-      return (
-        (Number(strokeWidthValue) * 24) / Number(size ?? contextSize ?? defaultAttributes.width)
-      );
-    }
+  const { class: propsClass, ...restProps } = props;
+
+  const defaultSlot = slots.default?.();
 
-    return strokeWidthValue;
+  const [, svgAttributes, builtIconNode = []] = buildLucideIconNode(icon, {
```
ericfennis, line 78: Is `computed` not needed here?

packages/lucide-react-native/src/Icon.ts
```diff
@@ -44,39 +74,51 @@ const Icon = forwardRef<SVGSVGElement, IconComponentProps>(
...
-        : strokeWidth ?? contextStrokeWidth;
+    const hasAccessibleProp = Boolean(children) || hasA11yProp(rest);
+
+    const [, svgAttributes, builtIconNode = []] = buildLucideIconForReact(icon, {
+      color: color ?? contextColor,
+      width: width ?? size ?? contextSize,
+      height: height ?? size ?? contextSize,
+      strokeWidth: strokeWidth ?? contextStrokeWidth,
+      absoluteStrokeWidth: absoluteStrokeWidth ?? contextAbsoluteStrokeWidth,
+      nonScalingStroke: nonScalingStroke ?? contextNonScalingStroke,
```
ericfennis, line 93: We need to test if `non-scaling-stroke` is working in React Native. Have you tested this?

ericfennis (changes requested):
Nice work overal. 
Very big change, so will do a second review later. I will need some more time for testing if everything correctly working as expected and build outputs are correct.

For now there are couple of things we need to address and check:
- [x] Test non-scaling-stroke in React Native, not sure if this working as expected.
- [ ] Wrap some `buildLucideIconForReact` in `useMemo` to prevent re-execution every render.
- [x] Check if reactivity is still working in Vue and Solid.
- [ ] Adjust documentation, replace `absoluteStrokeWidth ` with `nonScalingStroke`

packages/lucide-react/src/Icon.ts
karsa-mistmere: @ericfennis: Wrapping this in `useMemo` is not really viable as-is.

`rest` would need to be included in the dependency list because it is forwarded to `buildLucideIconNode` as SVG attributes. Omitting it could leave rendered attributes or event handlers stale.

However, object rest destructuring creates a new `rest` object on every render, so including it would invalidate the memo every time anyways, since `useMemo` compares dependencies using `Object.is`, which only considers objects equal when they share the same reference in memory.

Trying to fingerprint `rest` would not be a safe alternative either, since it may contain functions and other non-serializable values, and computing such a fingerprint could easily cost more than the work we are trying to avoid.

I see two reasonable options:

1. Leave this as-is – the builder is currently quite small and inexpensive: it resolves defaults, merges classes and attributes, and maps the icon node. Memoizing it would likely add complexity without meaningful savings, and so far, we've been doing the same operations unmemoized anyways.

2. Remove `attributes` from `buildLucideIconNode` and merge forwarded attributes afterward. That would make the builder easier to memoize because it would no longer depend on `rest`:

```ts
const [name, builtAttributes, builtIconNode = []] = useMemo(
  () =>
    buildLucideIconNode(icon, {
      color: color ?? contextColor,
      width: width ?? size ?? contextSize,
      height: height ?? size ?? contextSize,
      strokeWidth: strokeWidth ?? contextStrokeWidth,
      absoluteStrokeWidth:
        absoluteStrokeWidth ?? contextAbsoluteStrokeWidth,
      nonScalingStroke:
        nonScalingStroke ?? contextNonScalingStroke,
      className: mergeClasses(contextClass, classes as string),
      hasA11yProp: hasAccessibleProp,
    }),
  [
    icon,
    color,
    contextColor,
    width,
    height,
    size,
    contextSize,
    strokeWidth,
    contextStrokeWidth,
    absoluteStrokeWidth,
    contextAbsoluteStrokeWidth,
    nonScalingStroke,
    contextNonScalingStroke,
    contextClass,
    classes,
    hasAccessibleProp,
  ],
);

const svgAttributes = { ...builtAttributes, ...rest };
```

The caveat is that `buildLucideIconNode` could no longer inspect forwarded attributes directly, should we need that in the future.

packages/lucide-solid/src/Icon.tsx
karsa-mistmere: Totally fair – it wasn't.

Fixed it and added some rudimentary unit tests for this.

packages/vue/src/Icon.ts
karsa-mistmere: I've added regression tests for reactive props and provided context, and they pass without wrapping this in `computed`.

My understanding is that in Vue, the `FunctionalComponent` body is the render function, so it reruns when reactive state used during render changes.

packages/lucide-react-native/src/Icon.ts
karsa-mistmere: Only via snack.expo.dev, but it worked there. `react-native-svg` has had support for `vector-effect` [for quite some time](https://github.com/software-mansion/react-native-svg/pull/955) now.

karsa-mistmere:
@ericfennis: I've now also upgraded `lucide-react`'s documentation with `nonScalingStroke`, let me know if these changes are OK with you so that I can upgrade the documentation for the rest of the packages.

ericfennis (approved):

PR #4413, opened 2026-05-26
fix(site): empty search results on first load for tag-only matches

taimar:
## Summary

Fixes an occasional bug on `/icons` where searches like "clear" show "No results" even though matching icons exist.

## Root cause

The search index includes `tags` as a search key, and several queries (e.g. "clear") only match via tags — not by name or alias. Tags and categories are fetched lazily on `@focus` of the search input. If `searchQueryDebounced` fires before `/api/tags` resolves, Fuse runs against the bare `props.icons` (no tag data) and finds zero matches, so `NoResults` is rendered. The page works after a refresh because the URL contains `?search=…`, which flips `useFetch.immediate` to `true` and loads tags in parallel with mount.

## Fix

- Suppress `NoResults` while tag/category metadata is still in flight (`isSearchMetadataLoading`).
- Also trigger the fetch from a `watch(searchQuery)` so the metadata loads even when `@focus` never fires (e.g., autofill, programmatic input, restored query).
- Guard `loadSearchMetadata` on `isFetching` to avoid duplicate requests now that it can be called from both the focus handler and the watch.
- Apply the same change to `IconsOverview.vue` and `IconsCategoryOverview.vue`.


karsa-mistmere (changes requested):
Hey @taimar,

Nice catch, and the root cause makes sense.

One small concern: `isSearchMetadataLoading` only suppresses `NoResults` while `isFetchingTags` / `isFetchingCategories` is already true, but I guess there could still be a short render window where `searchQuery` is non-empty, metadata is missing, but the fetch has not started yet. 🤔

Maybe the `isSearchMetadataLoading` implementations should instead just directly reflect whether the metadata required for searching is available yet, e.g.:

```ts
const isSearchMetadataLoading = computed(
  () =>
    searchQuery.value.length > 0 &&
    (tags.value == null || categories.value == null),
);
``` 

That way we avoid showing a false negative regardless of the exact fetch timing.

taimar:
@karsa-mistmere it should be good now

docs/.vitepress/theme/components/icons/IconsCategoryOverview.vue
```diff
@@ -117,6 +121,7 @@ const categoriesList = computed(() => {
     }, []);
 });
 const searchPlaceholder = useSearchPlaceholder(searchQuery, searchResults);
+const isSearchMetadataLoading = computed(() => searchQuery.value.length > 0 && tags.value == null);
```
karsa-mistmere: Shouldn't this also include the categories?

```suggestion
const isSearchMetadataLoading = computed(
  () => searchQuery.value.length > 0 && (tags.value == null || categoriesMap.value == null),
);
```

taimar: I guess symmetry with the icons page is worth more than guarding against the narrow case where /api/categories lags behind /api/tags and a brand search momentarily lands on a blank area instead of the rejection panel. Applied.

/edit: linting error is not from this PR

taimar:
@karsa-mistmere are any additional changes needed or could it move forward?

taimar:
@karsa-mistmere @ericfennis gentle bump – anything still blocking this one? Happy to adjust if needed.

ericfennis (approved):

ericfennis:
LGTM! Nice work @taimar

PR #4423, opened 2026-05-30
fix(metadata): add missing use-cases prop on play-off.json

karsa-mistmere:
## Before Submitting 

### DIFF
--- a/icons/play-off.json
+++ b/icons/play-off.json
@@ -5,7 +5,10 @@
     "karsa-mistmere",
     "Ahmed-Dghaies"
   ],
-  "use-cases": [],
+  "use-cases": [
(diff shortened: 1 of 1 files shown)


icons/play-off.json
```diff
@@ -5,6 +5,7 @@
     "karsa-mistmere",
     "Ahmed-Dghaies"
   ],
+  "use-cases": [],
```
ericfennis:
```suggestion
  "use-cases": [
    "Indicates that playback is disabled for video, live stream, or audio file",
    "Indicates that video, live stream, or audio file cannot be played."
  ],
```

ericfennis (approved):

PR #4488, opened 2026-06-18
fix(docs): fix color input clipping

Hsiii:
## Description

Adjust icon customizer color control:
- make color field full-width to avoid clipping input value longer than 6 chars
- use the existing mono font token to prevent `#ffffffff` and `#00000000` taking considerable different lengths to make layout easier to maintain
- long hex values are capped eight characters instead of infinite to prevent confusion

|Before|After|
|---|---|
|image|image|

## Before Submitting 


ericfennis (approved):
Nice work @Hsiii .
Works well, like it!

PR #4491, opened 2026-06-22
fix(astro): add Astro v7 compatibility

iseraph-dev:
## Description

[Astro 7](https://astro.build/) has been released. `@lucide/astro` already works with it unchanged, but its `peerDependencies` range stops at `^6`, so Astro 7 projects get an unmet peer-dependency warning on install.

This widens the peer range to include `^7`

- `peerDependencies.astro`: `^4 || ^5 || ^6` → `^4 || ^5 || ^6 || ^7`

## Before Submitting

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -61,6 +61,6 @@
     "astro": "^6.1.10"
   },
   "peerDependencies": {
-    "astro": "^4 || ^5 || ^6"
+    "astro": "^4 || ^5 || ^6 || ^7"
   }
 }


jguddas (approved):

iseraph-dev:
I initially tried upgrading the package to Astro 7/Vite 8, but that triggers a lockfile re-resolution that changes the dependency layout across the monorepo and causes additional snapshot test failures in unrelated packages. Since the issue is tied to the shared workspace dependency graph rather than `@lucide/astro` itself, I opted for the minimal change: expanding the peer dependency range to include Astro 7 without modifying the lockfile. This keeps the workspace stable and all tests green. A fully validated Astro 7 setup would require a larger workspace-wide migration to Vite 8 and Vitest alignment.

ericfennis (approved):

PR #4496, opened 2026-06-24
feat(icons): add `square-off` icon

october-learns:
## Description
Add `square-off` icon 

### Icon use case 
For adding a check box which status is in 'not applicable' 

### Concept 

### Author, credits & license
- [x] I've based them on the following Lucide icons: `square` , `circle-off`

### Naming 
- [x] I've read and followed the [naming conventions](https://lucide.dev/contribute/icon-design-guide#naming-conventions)

### Design 
- [x] I've read and followed the [icon design guidelines](https://lucide.dev/contribute/icon-design-guide)

## Before Submitting


ericfennis:
Just curious for the use-case, since this may seem like the wrong direction from a UX perspective. Maybe better to show something "disabled" with opacity or something.

Are you trying to add a checkbox to a user interface with this icon?
Or do you have a form configurator where you can add checkboxes?

karsa-mistmere:
I think what @october-learns is talking about here is a checkbox of [indeterminate state](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/indeterminate), which is commonly represented as either of the two: https://lucide.dev/icons/square-slash or https://lucide.dev/icons/square-minus

ericfennis:
The indeteminate state is to indicate if something is between on or off right? 
What @october-learns mentioned is "not applicable", this seems more like disabled state.

karsa-mistmere:
> The indeteminate state is to indicate if something is between on or off right? What @october-learns mentioned is "not applicable", this seems more like disabled state.

The exact phrasing what that the "status is in 'not applicable'", that's what I'm taking to mean an indeterminate status.

But yes, we need more feedback on this one.

october-learns:
Hi there, thank you for reviewing this request!
I'm currently working on a form configurator for a digitalization project, and I'd like to introduce a checkbox type that supports three states: _true_, _false_, and _not applicable_.
Form preparers have access to configure the forms, which must replicate the printed versions exactly as originally designed. Some forms are shared across several purposes, but a few of the questions differ between them. Allowing preparers to define whether each question is applicable isn't very user-friendly, as it introduces complex conditional logic. Instead, for questions that don't apply, the form user can simply select the "_not applicable_" state.
The reason this can't be handled with the square-slash or square-minus icons is that those tend to convey "_disabled by design_" rather than a deliberate user choice of "_not applicable_."
Hope this makes sense.

karsa-mistmere:
> Instead, for questions that don't apply, the form user can simply select the "_not applicable_" state. The reason this can't be handled with the square-slash or square-minus icons is that those tend to convey "_disabled by design_" rather than a deliberate user choice of "_not applicable_." Hope this makes sense.

Thanks for the additional context!

I still think this maps to the existing concept of a three-state (or "mixed"/indeterminate) checkbox, rather than requiring a new icon. While your application assigns the semantic meaning of "Not applicable" to that third state, the icon itself is only responsible for representing that there *is* a third state – it doesn't generally communicate *why* that state exists or whether it was chosen by the user versus being determined by the application.

For that reason, I think `square-minus` is still the appropriate representation here. This also aligns with established UI patterns, where the indeterminate checkbox state is commonly shown with a horizontal line, see the documentation for the HTML `indeterminate` checkbox state:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/indeterminate

Because of this, I don't think `square-off` works for _this_ particular use case.

That being said, I'm very open to adding `square-off` as an icon, since it's a quite generic one, with probably loads of applicable use cases, it's just that this particular one is not very legitimate.

icons/square-off.json
```diff
@@ -0,0 +1,36 @@
...
+  ],
+  "categories": [
+    "shapes",
+    "design"
+  ],
+  "use-cases": [
+    "For adding a check box which status is in 'not applicable' ",
+    "indicating a checkbox in a not applicable state",
+    "marking a disabled or unavailable option in forms",
+    "denoting an unchecked or toggled-off square control"
```
karsa-mistmere: Let’s rework these to reflect disabled, invalid, or unavailable states rather than "not applicable". With those changes, I think this PR will be good to go.

```suggestion
    "indicating a disabled or unavailable state",
    "marking something as blocked or prohibited",
    "showing a deactivated interface element",
    "representing a closed or inactive component"
```

karsa-mistmere (changes requested):

icons/square-off.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve" style="fill-rule:evenodd;clip-rule:evenodd;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:1.5" viewBox="0 0 24 24"><path d="M0 0h24v24H0z" style="fill:none"/><path d="M22 22 2 2M20.414 20.414A2 2 0 0 1 19 21H5a2 2 0 0 1-2-2V5c0-.552.224-1.052.586-1.414M9 3h10a2 2 0 0 1 2 2v10" style="fill:none;stroke:#000;stroke-width:2.08px"/></svg>
```
karsa-mistmere: Does not follow our code conventions, see https://lucide.dev/contribute/icon-design-guide#code-conventions

```suggestion
<svg
  xmlns="http://www.w3.org/2000/svg"
  width="24"
  height="24"
  viewBox="0 0 24 24"
  fill="none"
  stroke="currentColor"
  stroke-width="2"
  stroke-linecap="round"
  stroke-linejoin="round"
>
  <path d="M20.4 20.4a2 2 0 01-1.4.6H5a2 2 0 01-2-2V5a2 2 0 01.59-1.41" />
  <path d="M21 15.3V5a2 2 0 00-2-2H8.7" />
  <path d="M22 22 2 2" />
</svg>
```

karsa-mistmere (changes requested):

karsa-mistmere (approved):

ericfennis (approved):

PR #4517, opened 2026-07-02
feat(site): integrate Lab icon filter with search controls

### LINKED ISSUE #4489: Integrate icon visibility toggle into search filter
### Package

- [ ] lucide
- [ ] lucide-react
- [ ] @lucide/vue
- [ ] @lucide/svelte
- [ ] @lucide/angular
- [ ] @lucide/astro
- [ ] lucide-react-native
- [ ] lucide-preact
- [ ] lucide-solid
- [ ] lucide-static
- [ ] @lucide/icons
- [ ] Figma plugin
- [ ] all JS packages
- [x] site
- [ ] other/not relevant

### Description

Add a filter control next to sorting control to let user choose whether to include external lib icons.

### Use cases

The current visibility toggle lives in sidebar, but it directly affect search result and icon count shown in search bar:

Image

It might be more intuitive for users if this control is integrated near search bar (e.g. for me, it took me a while to figure out how that control affected the page).

It might also be worth considering enabling Lab icons by default. Since the goal of lucide lab is to explore icon use cases, expose them directly could help user to discover and try them out. 

If this sounds good, I'd be happy to submit a PR!

### Checklist

Hsiii:
## Description

Resolves #4489.

Moves the Lab visibility control from the sidebar into the icon search toolbar so it sits with the search and sort controls it affects.

---

Before: the external icon toggle lived in the sidebar, away from the search result count.

image
image

---

After: the toolbar includes an All icons / Main library filter next to search and sort, includes Lab icons by default, and uses the arrow-up-down sort glyph for better readability.

## Before Submitting


taimar:
The select makes this a two-click interaction every time (open the dropdown to see, then choose), where the checkbox was a single click. And since it’s a binary state, a single checkbox feels simpler than two separate options.

Also, category choices are already on the left, so the sidebar placement was consistent with the other filtering options.

sago-cream:
Thanks for the feedback. I agree that requiring two clicks for a binary option is less ideal.

My thought was that, since this control directly affects the search results, it might fit better alongside the search and sort controls. Meanwhile, the current category controls feel more like a view switch than a filter, they let users browse in category mode rather than filtering the search results.

Would it make sense to keep it as a simple checkbox/toggle, but move it next to the search and sort controls instead?

ericfennis:
Hi @Hsiii, 
Sorry, I need to close this. This is not really the way to go. Since the lab is "experimental", it should not be visible by default.
I also agree with what @taimar said.

What maybe makes more sense is to save the checkbox in localStorage so it persists when you visit the next time.

sago-cream:
Hi @ericfennis,

Thanks for the feedback. I understand the reasoning.

I’ll also close #4489.

PR #4519, opened 2026-07-03
fix(packages/angular): move child rendering logic to template from effect

karsa-mistmere:
Closes #4515

## Problem

`@lucide/angular` currently renders SVG child nodes imperatively in `LucideIconBase` using `Renderer2`.

While this works during normal client rendering, it breaks Angular SSR hydration. The server renders the SVG paths correctly, but because they are created outside Angular's template, hydration cannot claim them. The client then renders them again, resulting in duplicated SVG child nodes.

## Solution

Move SVG child rendering into Angular's template using computed signals.

This allows hydration to recognise and claim the existing server-rendered nodes instead of recreating them, making SSR and client rendering produce the same logical view.

## Implementation

- Replaced the imperative render effect with declarative template rendering using `@for` and `@switch`.
- `LucideIconBase` now exposes computed signals for:
  - `iconNodes`, including `vector-effect="non-scaling-stroke"` when `absoluteStrokeWidth` is enabled.
  - `iconClasses` for the dynamic `lucide-*` classes.
- Removed all `Renderer2`-based rendering.
- Icon `key` values are now used solely for `@for` tracking and are no longer rendered as SVG attributes.

While the resulting template is somewhat verbose, this approach is fully hydration-native and relies exclusively on Angular's built-in rendering primitives.

## Before Submitting 

### DIFF
--- a/packages/angular/package.json
+++ b/packages/angular/package.json
@@ -54,6 +54,7 @@
     "@angular/core": "^21.2.17",
     "@angular/forms": "^21.0.0",
     "@angular/platform-browser": "^21.0.0",
+    "@angular/platform-server": "^21.2.17",
     "@angular/router": "^21.0.0",
     "@lucide/build-icons": "workspace:*",
     "@lucide/helpers": "workspace:*",
--- a/packages/angular/src/lucide-dynamic-icon.spec.ts
+++ b/packages/angular/src/lucide-dynamic-icon.spec.ts
@@ -22,7 +22,12 @@ describe('LucideDynamicIcon', () => {
   let component: LucideDynamicIcon;
   let fixture: ComponentFixture<LucideDynamicIcon>;
   let icon: WritableSignal<LucideIconInput | null | undefined>;
-  const getSvgAttribute = (attr: string) => fixture.nativeElement.getAttribute(attr);
+  const expectSvgClasses = (classes: string[]) => {
+    for (const cssClass of classes) {
+      expect(fixture.nativeElement.classList.contains(cssClass)).toBe(true);
+    }
+  };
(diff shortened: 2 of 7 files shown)


daroczig:
This is great :raised_hands: I've tested in our SSR app and works wonderfully. Thanks a ton :bow: 

---

My only concern would be the list of hardcoded elements of the PR, but as you said:

> Since Lucide only supports a small, fixed set of SVG elements and attributes ...

It's probably not an issue at all.

---

Thank you very much again, this is truly fantastic, and I'm looking forward to see it in prod soon :blush:

ericfennis (approved):

ramtinsoltani:
@ericfennis @daroczig @karsa-mistmere Question for the maintainers: is the attribute list in this switch statement (`packages/angular/src/lucide-icon-template.ts:9`) meant to be the new strict, permanent contract for `LucideIconData`, or is it an incidental side effect of the effect-to-template migration that's still open to expanding?

Right now fill is only bound on path/circle (not rect, ellipse, line, polygon, polyline), and stroke, stroke-width, stroke-linecap, stroke-linejoin, fill-opacity, class, and id aren't bound on any element type. That's fine for the built-in icon set since those icons never use per-node attribute overrides, they inherit everything from the root svg. But `LucideIconData` is a documented public API for custom icons (https://lucide.dev/guide/angular/advanced/icon-provider#registering-custom-icons), and this change silently drops any of those attributes with no console warning if someone was relying on them (fill icons using stroke-width: 0 overrides, multi-color icons with per-shape stroke/fill).

If this is intentional going forward, it'd help to document the supported attribute list explicitly so custom icon authors know the constraints upfront. If it's not intentional, happy to help test a fix, since right now there's no way to tell from the outside which one it is.

Issue #4424, opened 2026-05-30
`LucideProvider` is undefined in `lucide-react-native`


AhmedBaset:
### Package

- [ ] lucide
- [ ] lucide-react
- [ ] @lucide/vue
- [ ] @lucide/svelte
- [ ] @lucide/angular
- [ ] @lucide/astro
- [x] lucide-react-native
- [ ] lucide-preact
- [ ] lucide-solid
- [ ] lucide-static
- [ ] @lucide/icons
- [ ] lucide-vue-next (deprecated)
- [ ] lucide-svelte (deprecated)
- [ ] lucide-angular (deprecated)
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

1.17.0

### Can you reproduce this in the latest version?

- [x] Yes
- [ ] No

### Browser

- [ ] Chrome/Chromium
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [x] Other/not relevant

### Operating system

- [ ] Windows
- [x] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [ ] Other/not relevant

### Description

the published tarball of `lucide-react-native` doesn't contain `LucideProvider`

https://npmx.dev/package-code/lucide-react-native/v/1.17.0/dist%2Fesm%2Fcontext.mjs
```js
/**
 * @license lucide-react-native v1.17.0 - ISC
 *
 * This source code is licensed under the ISC license.
 * See the LICENSE file in the root directory of this source tree.
 */

import { createContext, useContext } from 'react';

const LucideContext = createContext({
  size: 24,
  color: "currentColor",
  strokeWidth: 2,
  absoluteStrokeWidth: false
});
const useLucideContext = () => useContext(LucideContext);

export { useLucideContext };
//# sourceMappingURL=context.mjs.map
```

the package exports it but it somehow doesn't exist in `./context`
```
// lucide-react-native.mjs
export { LucideProvider, useLucideContext } from './context.mjs';
```

Since LucideProvider is undefined, React errors:
> Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

### Steps to reproduce

1. `Import { LucideProvider } from "lucide-react-native";`
2. Render it

### Checklist

[...]

PR #4528, opened 2026-07-09
fix(lucide-react-native): restore the icons entry point in the JS bundles

### LINKED ISSUE #4527: [react-native] `./icons` entry point missing from 1.22.0 and 1.23.0 packages
The `./icons` entry point is broken in the last two `lucide-react-native` publishes (1.22.0 and 1.23.0).

`package.json` exports still declares it:

```json
"./icons": {
  "types": "./dist/icons.d.ts",
  "import": "./dist/esm/icons/index.mjs",
  "browser": "./dist/esm/icons/index.mjs",
  "require": "./dist/cjs/icons/index.js"
}
```

but `dist/esm/icons/index.mjs` and `dist/cjs/icons/index.js` are missing from the published tarball. The individual icon files are all there, just not the index barrels.

Easy to verify without installing:

```
$ tar -tzf lucide-react-native-1.23.0.tgz | grep "icons/index"
(nothing)

$ tar -tzf lucide-react-native-1.21.0.tgz | grep "icons/index"
package/dist/cjs/icons/index.js
package/dist/cjs/icons/index.js.map
package/dist/esm/icons/index.mjs.map
package/dist/esm/icons/index.mjs
```

So `import * as icons from "lucide-react-native/icons"` fails to resolve. In Metro (Expo / React Native) it looks like this:

```
WARN  The package lucide-react-native contains an invalid package.json configuration.
Reason: The resolution for ".../lucide-react-native/icons" defined in "exports" is .../dist/esm/icons/index.mjs, however this file does not exist. Falling back to file-based resolution.

Unable to resolve module lucide-react-native/icons from src/components/Icon.tsx
```

One nasty detail: `dist/icons.d.ts` is still published, so TypeScript resolves fine and this only blows up at bundle time. It took out one of our release builds before we noticed.

1.21.0 is the last good version, pinning back to it works as a workaround.

pvinis:
Closes #4527

## Description

#4497 removed `src/icons/index.ts` from the JS bundle inputs. The barrel is a pure re-export module, so with `preserveModules` rollup optimizes it away when it is not an entry point. `dist/esm/icons/index.mjs` and `dist/cjs/icons/index.js` stopped being emitted, while `package.json` still declares the `./icons` export. `dist/icons.d.ts` is still built, so TypeScript resolves fine and this only surfaces at bundle time.

Just adding the entry back would reintroduce #4424: the config built each entry as a separate rollup config writing into the same output directory, so the icons build could overwrite shared modules with tree-shaken copies that were missing exports like `LucideProvider`.

Instead, this builds all entries in a single rollup config per format. One module graph, so shared modules are emitted once with everything any entry needs, and the barrel comes back because entry points are never optimized away.

Verified with a local `pnpm build`:

- `dist/esm/icons/index.mjs` and `dist/cjs/icons/index.js` are emitted again, same shape as the 1.21.0 publish
- `dist/esm/context.mjs` and `dist/cjs/context.js` still export `LucideProvider` and `useLucideContext`, so #4424 stays fixed
- `pnpm test` passes

## Before Submitting

### DIFF
--- a/packages/lucide-react-native/rollup.config.mjs
+++ b/packages/lucide-react-native/rollup.config.mjs
@@ -5,7 +5,7 @@ import pkg from './package.json' with { type: 'json' };
 const packageName = 'LucideReact';
 const outputFileName = 'lucide-react-native';
 const outputDir = 'dist';
-const inputs = ['src/lucide-react-native.ts'];
+const inputs = ['src/lucide-react-native.ts', 'src/icons/index.ts'];
 const bundles = [
   {
     format: 'cjs',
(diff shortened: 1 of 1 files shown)


ericfennis (approved):
Nice work, thanks!

PR #4558, opened 2026-07-17
chore(copilot-reviews): Improve use-cases description.

ericfennis:
A small guide to copilot how to write use-caces.

We should also have a guide for humans, but working on better contribution guide for that.


jguddas (commented):
What about forbidding chaining, and requiring multiple use cases, no or or or?

So 

> - `calendar-sync` → ["Representing recurring events, promotions, or limited-time offers"]

Would become 

- `calendar-sync` → [
"Representing recurring events", "Representing promotions",
"Representing limited-time offers"
]

ericfennis:
@jguddas I get you point. For some time I thought the same. But there is not really a benefit here to it. It's not that context window gets wider or something. It's just makes it more impractical to my opinion.

See this example:
```json
"use-cases": [
    "Indicating a USB-C port or connector",
    "Representing data transfer and device connectivity",
    "Showing a charging or power delivery interface",
    "Marking USB-C inputs and outputs on devices",
    "Indicating USB-C charging compatibility on product listings",
    "Denoting a reversible connector for user setup instructions",
    "Signifying USB-C video output or alt-mode support",
    "Labeling ports on docking stations, hubs and adapters",
    "Representing USB-C cables and accessory listings"
  ],
```

If we require this and we take a list of use cases for example for `use-c` icon it get super long. And harder to review.

jguddas:
It will make it hard to match use cases.

`represent a or b` has quite a different embed form `represent a`.

I guess we can try to write a script to discombobulate them tho.

PR #4569, opened 2026-07-21
fix(packages/angular): pass through stroke and fill presentation attributes on icon nodes

ramtinsoltani:
Addresses https://github.com/lucide-icons/lucide/pull/4519#issuecomment-5028975970

## Description

#4519 moved child icon node rendering from an imperative effect to a declarative template, which only binds a fixed subset of attributes per SVG element type (mostly positional attributes, plus `fill` on `path` and `circle` only).

This is fine for the built-in icon set, since those icons always inherit stroke and fill from the root `<svg>`. It's a silent regression for consumers of `LucideIconData`, the [documented public API](https://lucide.dev/guide/angular/advanced/icon-provider#registering-custom-icons) for custom icons, who rely on per-node overrides like `stroke`, `stroke-width`, `stroke-linecap`, `stroke-linejoin`, `stroke-dasharray`, `stroke-dashoffset`, `fill-opacity`, and `opacity`. These attributes are now dropped with no warning, and `fill` is also missing entirely on `rect`, `ellipse`, `line`, `polygon`, and `polyline`.

This PR adds the standard SVG presentation attributes to every element case in the template, matching what custom multi-color and multi-stroke-width icons need, while keeping the declarative, SSR-safe rendering approach introduced in #4519. No changes to the built-in icon set's rendered output, since none of those icons use the newly-bound attributes.

### DIFF
--- a/packages/angular/src/lucide-dynamic-icon.spec.ts
+++ b/packages/angular/src/lucide-dynamic-icon.spec.ts
@@ -52,6 +52,38 @@ describe('LucideDynamicIcon', () => {
       ['rect', { x: 1, y: 2, width: 3, height: 4, rx: 5, ry: 6, key: 'rect-key' }],
     ],
   };
+  const fillAndStrokeAttrs = {
+    fill: 'red',
+    'fill-opacity': '0.5',
+    stroke: 'blue',
+    'stroke-width': '3',
+    'stroke-linecap': 'round',
+    'stroke-linejoin': 'bevel',
+    'stroke-dasharray': '4 2',
+    'stroke-dashoffset': '1',
+    opacity: '0.75',
+  };
+  const strokeOnlyAttrs = {
+    stroke: 'blue',
+    'stroke-width': '3',
+    'stroke-linecap': 'round',
+    'stroke-linejoin': 'bevel',
+    'stroke-dasharray': '4 2',
+    'stroke-dashoffset': '1',
+    opacity: '0.75',
+  };
+  const presentationAttrsIcon: LucideIconData = {
+    name: 'presentation-attrs',
+    node: [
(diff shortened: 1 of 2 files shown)


karsa-mistmere:
Hey @ramtinsoltani,

Thanks for raising this and putting together a fix! I've opened a separate PR in #4570 that takes a deliberately more limited approach, which will supersede this one.

Extending per-node attribute support is actually something I've wanted to do since the declarative rendering migration, but hadn't gotten around to yet — so thanks for giving me the kick I needed to finally get it done. 😄

I also agree that we should make the supported contract explicit. Type-restricting `LucideIconNode` to only allow specific SVG elements with the specific attributes we actually support would probably be a good way to achieve that. However, I think this should be done as part of, or after, #4409, so we can keep the API and devX consistent across packages rather than defining an Angular-specific contract in isolation.

That said, we definitely don't want to support arbitrary SVG presentation attributes here. In particular, attributes such as `stroke-width`, `stroke-linecap`, `stroke-linejoin`, `stroke-dasharray`, and `stroke-dashoffset` will not be supported, as they allow custom icons to fundamentally alter Lucide's established stroke characteristics and stray from its design language.

The new PR instead intentionally supports a limited subset:
- `class`
- `id`
- `opacity`
- `fill`
- `fill-opacity`
- `stroke`
- `stroke-opacity`

These cover the attributes that have reasonable use cases for custom solid, duotone, and per-node color/opacity styling without opening the door to arbitrary changes to Lucide's core visual conventions.

Thanks again for bringing attention to this!

karsa-mistmere:
Hey @ramtinsoltani,

I've not heard back from you, but #4570 has now been merged.

ramtinsoltani:
@karsa-mistmere Thank you for your response, and that makes sense.

PR #3386, opened 2025-07-02
fix(icons): redesigned `podcast` icon

jguddas:
resolves #3384

## What is the purpose of this pull request?
- [x] Other: Icon update

### Description
Redesign to resolve the following issues:
- Brand guideline violation (copy of Apple Podcasts logo)
- 2px gap rule violation

## Before Submitting
- [x] I've read the [Contribution Guidelines](https://github.com/lucide-icons/lucide/blob/main/CONTRIBUTING.md).
- [ ] I've checked if there was an existing PR that solves the same issue.

PR #4583, opened 2026-07-23
feat(icons): replaces `podcast` icon with new `mic-*` icons

karsa-mistmere:
closes #3386 

## Description

Added two new `mic-*` icons to replace `podcast`, which contains a brand-infringing icon design.

### Concept 

### Author, credits & license

- [x] The icons were originally created in #3386 by @jguddas & I

### Naming 

- [x] I've read and followed the [naming conventions](https://lucide.dev/contribute/icon-design-guide#naming-conventions)

### Design 

- [x] I've read and followed the [icon design guidelines](https://lucide.dev/contribute/icon-design-guide)

## Before Submitting 


jguddas:
Mic should be microphone imo

karsa-mistmere:
> Mic should be microphone imo

[`mic`](https://lucide.dev/icons/mic) is also just mic

I think it's okay, since [the word "mic" doesn't really have more everyday meanings](https://en.wiktionary.org/wiki/mic#English) and is a quite commonly used abbreviation

jguddas:
> > Mic should be microphone imo
> 
> [`mic`](https://lucide.dev/icons/mic) is also just mic
> 
> I think it's okay, since [the word "mic" doesn't really have more everyday meanings](https://en.wiktionary.org/wiki/mic#English) and is a quite commonly used abbreviation

https://en.wikipedia.org/wiki/Mic_(media_company)

I don't like it, also it's an abbreviation for military industrial complex.

I don't think we should abbreviate icon names, abbreviations are harder to understand for non natives and just not clear.

karsa-mistmere:
> https://en.wikipedia.org/wiki/Mic_(media_company)

Wow, they named a company after a commonplace word. Utterly scandalous! 😱

Jokes aside, `mic` is an _extremely_ common shorthand for `microphone`, no-one in their right mind is going to think of a military industrial complex.

By the same logic, we might as well rename `lock` to `padlock`, since that's technically what the icon depicts. 🤷‍♂️

jguddas:
> > https://en.wikipedia.org/wiki/Mic_(media_company)
> 
> Wow, they named a company after a commonplace word. Utterly scandalous! 😱
> 
> Jokes aside, `mic` is an _extremely_ common shorthand for `microphone`, no-one in their right mind is going to think of a military industrial complex.
> 
> By the same logic, we might as well rename `lock` to `padlock`, since that's technically what the icon depicts. 🤷‍♂️

Yes, we should use `padlock` instead of `lock`, that way we also don't block ourselves from adding other lock icons.

In general I think it's only you who thinks it's obvious, and being descriptive is really something that we should enforce.

> In their final study, which they ran online, Fang and his colleagues found that 80% of participants predicted that other people would be indifferent to texting abbreviations, and 4.2% believed that the response would be positive. Only 15.8% predicted what the team had consistently found — that the impact was negative. It seems, then, that the vast majority of people are unaware of the undesirable effects of this common practice.

https://www.bps.org.uk/research-digest/omg-rly-text-abbreviations-lower-perceived-sincerity

It's obvious to you, but that doesn't mean it's obvious to everyone, it definitely isn't to me.

karsa-mistmere:
> In general I think it's only you who thinks it's obvious, and being descriptive is really something that we should enforce.

> It's obvious to you, but that doesn't mean it's obvious to everyone, it definitely isn't to me.

I'd much appreciate if we could steer clear of resorting to arguments about who does or doesn't personally find something obvious, and focus on the naming itself.

I don't think the study you linked is particularly relevant to this discussion. It concerns informal texting abbreviations such as "omg", "idk", and "rly", and how their use affects _perceived sincerity in interpersonal communication_.

That's a false equivalence to "mic":
1. Mic is not an ad-hoc texting abbreviation or internet shorthand in the same league as omg or rly; it's a long-established, conventional shortened form of microphone that has been used as a word in its own right for [more than half a century](https://www.merriam-webster.com/dictionary/mic).
2. Whether people perceive texting abbreviations as less sincere in day-to-day conversations tells us very little about whether mic is an appropriate or sufficiently descriptive name for a microphone icon.

Looking at [Iconify's collection of icon sets](https://github.com/iconify/icon-sets/tree/master/json), 34 use `mic`, while 82 use `microphone`. So while `microphone` is admittedly more common, `mic` is clearly not some unusual or obscure naming choice. Amongst others, [Material Symbols](https://fonts.google.com/icons?icon.query=mic) also uses simply `mic`. I'd say if it's good enough for Google, it should be good enough for us.

So in my opinion, unless there's concrete data showing that `mic` actually causes meaningful confusion, or some other non-anecdotal evidence that it isn't sufficiently well-established or widely understood, I'd vote to stick with what we have – `mic` has served us perfectly well for the past five or so years, and I've yet to see any reason why we should be in a hurry to rename it now.

> Yes, we should use `padlock` instead of `lock`, that way we also don't block ourselves from adding other lock icons.

Hard disagree on this one. The numbers overwhelmingly favour `lock`: 104 icon sets use `lock`, compared with only 7 using `padlock`.

I don't think we should optimise names for hypothetical future icons at the expense of the terminology users overwhelmingly expect today. If we ever add other types of locks, those can be disambiguated when and if we actually need to.

jguddas:
I think we should follow our guidelines and focus on having descriptive icon names, even when others don't.
I also know that abbreviations are less comprehendible and science backs that up, no matter the context also outside of texting abbreviations are less understandable, let's keep the icon names descriptive and understandable.

karsa-mistmere:
> I think we should follow our guidelines and focus on having descriptive icon names, even when others don't. I also know that abbreviations are less comprehendible and science backs that up, no matter the context also outside of texting abbreviations are less understandable, let's keep the icon names descriptive and understandable.

I fully agree with the principle here.

I just think "mic" _is_ already descriptive and widely understood enough, and I've not seen any evidence to the contrary.

The word mic essentially falls into the same category as:
* [bike](https://lucide.dev/icons/bike) → bicycle
* [phone](https://lucide.dev/icons/phone) → telephone
* fridge → [refrigerator](https://lucide.dev/icons/refrigerator)
* [ad](https://lucide.dev/icons/ad) → advertisement
* [app](https://lucide.dev/icons/app-window-mac) → application
* [burger](https://lucide.dev/icons/lab/burger) → [hamburger](https://lucide.dev/icons/hamburger)

So "abbreviations are less comprehensible" feels like a pretty broad a generalisation – we have plenty of similar conventional words that are just as descriptive in practice.

karsa-mistmere:
Coming back to this, I think a nice solution would be for icons like these to support two names, but with a distinction between canonical names and synonym aliases.

These synonym aliases could be available as first-class exports, without being treated as backward-compatibility aliases – i.e. they wouldn't generate compatibility CSS classes or trigger any of the other machinery intended for deprecated/renamed icons.

For example, we could represent this in the metadata with something along these lines:

```json
"aliases": [
  {
    "name": "microphone",
    "deprecated": false,
    "isSynonym": true,
  },
]
```

or adding a new metadata prop:

```json
"synonyms": ["microphone"]
```

That way, we could keep one canonical name without deprecating it, so people could continue using `import { Lock, Bike, Mic } from 'lucide-react';` without getting deprecation warnings, while the more explicit/descriptive names would also be available as imports.

What I definitely wouldn't do is deprecate all the existing `mic` icons as part of this PR. That would result in users getting deprecation warnings for long-established names that are still perfectly valid, without there being any actual need for them to migrate away from those names.

ericfennis:
I think `mic` is a very acceptable shorter way of naming a microphone. According to [wikipedia](https://en.wikipedia.org/wiki/Microphone): `A microphone, colloquially called a mic`.  So not super convinced to rename it.
We could adding an alias but that can also maybe get confusing as well. For example in code autocompletion you will get 8 results instead of 4. So not sure

ericfennis (approved):

karsa-mistmere:
> For example in code autocompletion you will get 8 results instead of 4. So not sure

Yeah, that is also a very valid point.

Issue #3804, opened 2025-11-20
Optimization of regular expression icons(icons/regex.svg)?


nyable:
Hello, Lucide team!
The "regex" icon in the lucide-icons set currently renders a hollow/outlined dot followed by an asterisk (appearing like "。") rather than the commonly recognized solid dot used in regex notation ("."). This makes the icon less immediately recognizable as a regular-expression symbol.
Is it possible to optimize this icon?
Thanks!

<img width="787" height="280" alt="Image" src="https://github.com/user-attachments/assets/c55f81d1-e320-456b-9a12-27ce983f15ae" />

jguddas:
We do not support fills.

PR #4611, opened 2026-07-29
fix(icons): update `regex` icon to use dot instead of rectangle path

AeonSake:
## Description
Since there are several icons that use a dot (e.g., `dot`, `circle-dot`, `grip`, etc.), I think it would be a good idea to revisit #3804 and replace the current rectangle path in the `regex` icon with a dot (circle path). In my opinion, this more closely resembles how regex is represented in other icons/fonts.

| `regex` (old) | `regex` (new) |
| --- | --- |
| <a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=2.m12.67r0.5_8.66-5m12.67_5.5_8.66_5M17_3v10M9r7Bn-2-2H5Bn-o2v2Bn_o2h2Bnu-2v-2z&name=regex">regex<br/>Open lucide studio</a> | <a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=1.LYRgTAdAbA7ABCADBArHAHNKBaFpKxwqoZZECyI8AzAG5IDaDARAMYCWATqwDYCmzADQBvNgA9mALiiC2ATykh0szooC%2BAXQ1A&name=regex">icons<br/>Open lucide studio</a> |

## Before Submitting


AeonSake:
I also debated moving the dot further down into the corner, but I'm not sure if I'd like that more over the current one.

<a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=1.LYRgTAdAbA7ABCADBArHAHNKBaFpKxwqoZZECyI8AzAG5IDaDARAMYCWATqwDYCmzADQBvNgA9mALgAsgtgE8pYRHM5SQAXwC6WoA&name=regex">icons<br/>Open lucide studio</a>

karsa-mistmere:
The dot from `dot` is not our favourite elements because of design language compliancy (or lack thereof), so I'd prefer not to propagate it even further.

Some alternative ideas:

minimal viable fix:

<a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=1.LYRgDAdAbA7ABAFggVjgDmlOyCyJUBMAbuKAJwQDMliKAtBlFsgNosBEATgKYDGALuwA0Ab3YAPdgC4EQ9gE9p%2BOQHcAlgBN%2BAC2lQ527moDm2wVP1dJUkAF8AuvaA&name=parentheses&utm_source=lucide-studio&utm_medium=embed">icons<br/>Open lucide studio</a>


version consistent with typographic set:

<a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=1.LYRhDoFYGYDYAJL1uAnAJgBzwCwFkQloA3TUTcHWHRAWhQ2xwG1mAiAJwFMBjAFzYAaAN5sAHmwBckQWwCeUkLFkB3AJYATPgAspOWdq5qA5toGT9nCZJABfALr2gA&name=parentheses&utm_source=lucide-studio&utm_medium=embed">icons<br/>Open lucide studio</a>


`/.*/` version:

<a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=2.m1p8_6_3M13_6v7m16_8-6_3m2o4-3r6M5_4_o20M7r7h.y&utm_source=lucide-studio&utm_medium=embed">icons<br/>Open lucide studio</a>

`(.*)` version:

<a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=2.m1x8_6_3M14_6v7m17_8-6_3M18_3s4_3_4_9-4_9-4_9M6u1s-4-3-4-9_4-9_4-9M9r7h.y&utm_source=lucide-studio&utm_medium=embed">icons<br/>Open lucide studio</a>

AeonSake:
- The square one really doesn't symbolize regex to the user, at least not without double-checking. As an example, I recently built a website/tool that has a search filter with these 3 options. The regex button looks really out of place with the square in my opinion; though I'm not a designer so what do I _really_ know :)
  search
  As a comparison, this is how the regex filter looks in VS Code:
  vsc
  It's also using a more dense version of the star (4 lines instead of 3) but that would probably not fit very well with the lucide design language.

- `/.*/` is a language specific regex escape sequence, used e.g., in JavaScript. I'd avoid adding a language bias in the icon set.
- `(.*)` is also redundant to `.*`. Both would match _all_ symbols with the difference of the former to also create a group containing the match, which is the same as the initial match group. Therefore, it's kind of meaningless? Might be a bit pedantic though. But I do see it being used frequently by other fonts and tools so I would probably advocate for that version.

In the meantime I also thought about aligning the outer edge of the star with the grid, instead of the center line to give a more balanced bounding box. That means the center line is now no longer aligned to the grid though.

<a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=1.LYRhDoGYBYAIQAzgKywBzgGyYLTNBDLMiulpsQLIgmYDsskAbogNqsBEAxgJYBOXADYBTDgBoA3twAeHAFzIx3AJ7yATAiV95IAL4BdfUA&name=regex&utm_source=lucide-studio&utm_medium=embed">icons<br/>Open lucide studio</a>

AeonSake:
> `(.*)` version:
> 
> <a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=2.m1x8_6_3M14_6v7m17_8-6_3M18_3s4_3_4_9-4_9-4_9M6u1s-4-3-4-9_4-9_4-9M9r7h.y&utm_source=lucide-studio&utm_medium=embed">icons<br/>Open lucide studio</a>

This but with an evenly spaced/angled star:
<a title="Open lucide studio" target="_blank" href="https://studio.lucide.dev/edit?value=2.m10.969r1.25_6.062-3.5m10.969_7.75_6.06o3.5M14_6v7M18_3s4_3_4_9-4_9-4_9M6u1s-4-3-4-9_4-9_4-9M9r7h.y">regex<br/>Open lucide studio</a>
It's now overlapping with the right parenthesis so would need to be moved further in. How about reducing the size of the parenthesis? Reducing the arch is probably not an option since there are other icons using the same style of parenthesis already.

karsa-mistmere:
> * The square one really doesn't symbolize regex to the user, at least not without double-checking. As an example, I recently built a website/tool that has a search filter with these 3 options. The regex button looks really out of place with the square in my opinion; though I'm not a designer so what do I _really_ know :)
>   search

I think, it's actually the other two icons that look out of place, especially if you compare it you the VSCode screenshot. :(

AeonSake:
> I think, it's actually the other two icons that look out of place, especially if you compare it you the VSCode screenshot. :(

To be fair, VS Code matches the icon height of these filter icons specifically so they align nicer. The icons all try to fill out a similar bounding box, which leads to the glyphs to get really stretched for those. Lucide icons (at least for those icons) don't match the bounding box height but the glyph height within, so glyphs look more square-ish, but then the overall icon becomes a rectangle instead.

Though I still think the square instead of a dot is a reach for regex. Square is neither a real glyph nor a valid symbol for a regex. A dot is implied, but not clear enough to not make it stand out.

ericfennis:
Let's keep it as it is

PR #4613, opened 2026-07-29
fix(icons): changed `chess-queen` icon

karsa-mistmere:
- [x] Other: Icon update

### Description
Redrew top part to be less dense.

## Before Submitting


jguddas (changes requested):
nah, looks way less like a chess-queen piece now.

jguddas:
This version looks less pointy and a bit like frog feet.

image

PR #4619, opened 2026-07-30
ci(security): improve security with adding permissions

ericfennis:
Applies Githubs advisory to add permissions to every workflow in public repo.

### DIFF
--- a/.github/workflows/angular.yml
+++ b/.github/workflows/angular.yml
@@ -7,6 +7,9 @@ on:
       - tools/build-icons/**
       - pnpm-lock.yaml
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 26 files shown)


jguddas (commented):
It's better to put the permission field per job, that way you don't leak permissions down into other jobs.

Let's say I have two jobs:

1. lint
2. test

and I want to add `pull-request: write` to one of them, with top level permissions many will just add `pull-request: write` to the top level field, rather than add a new permissions field on the job.

ericfennis:
@jguddas Yeah understand your concern. 
But I don't think the fix is "always per job". It depends on the workflow.
Some workflows need same permissions per block like `linting-icons.yml`. Repeating that three times doesn't make anything safer, it just adds lines that can drift apart later. Same for the package CI files: one job, `contents: read`, nothing to leak.

I noticed the `ci.yml` should be more strictly, so there it makes sense. I've added `permissions: {}` top level to workflows that have different permissions per job.

PR #4625, opened 2026-08-02
docs(site): add Juno to showcase

MarshallBear1:
## Description

Adds [Juno](https://junocompanion.com/) to the official **Used by** showcase. Juno is a production React Native app with `lucide-react-native` as a direct dependency; Lucide icons ship in its voice-call controls, attachment flow, onboarding, accessibility controls, templates, and subscription-state UI.

The card uses a 128 px copy of Juno's production app icon and links to the current canonical product URL. The one-line `object-fit: contain` rule prevents square app marks from being distorted inside the existing 240 × 64 logo frame and leaves the existing marks contained at their current dimensions.

Disclosure: this submission is from the Juno team. It proposes factual product adoption only; no affiliation or endorsement is implied. This is not an icon contribution.

## Verification

- Full `@lucide/docs` production build passed on Node 24.14.0.
- Generated `showcase.html` contains the direct canonical Juno link and both image variants.
- `companiesData.json` parses successfully and the asset paths resolve.
- `git diff --check` passes.

## Before Submitting


karsa-mistmere:
Hey @MarshallBear1,

Thanks for the submission!

While we don’t have a formal checklist for the showcase, it’s a curated list that we maintain at our own discretion. In general, we reserve it for projects with broad recognition or an established presence in the ecosystem.

With that in mind, we’re going to pass on adding this one, but we appreciate your interest and wish you all the best with the project.

PR #4627, opened 2026-08-03
fix(site): layer sidebar scroll beneath brand header

Hsiii:
## Description

Paints the desktop sidebar's brand header as an opaque foreground layer using the existing sidebar background token. This keeps overscrolling sidebar content visually beneath the Lucide brand block instead of covering it.

Before and after:

Before and after

## Before Submitting 


docs/.vitepress/theme/style.css
```diff
@@ -158,6 +158,12 @@ html {
   border-radius: 2px;
 }
 
+@media (min-width: 960px) {
+  .VPNavBar.has-sidebar .container > .title {
+    background-color: var(--vp-sidebar-bg-color);
+  }
+}
+
```
karsa-mistmere, line 166: This doesn't fully solve the issue, as rubber banding will still be present if the sidebar is scrolled fast enough.

I think a more universal solution would be to disable overscroll altogether:


```suggestion
.VPSidebar {
  overscroll-behavior: none
}
```

karsa-mistmere (changes requested):

docs/.vitepress/theme/style.css
sago-cream: Thanks for the feedback!
My intent wasn’t to remove rubber-banding, but to prevent it from making the sidebar content appear to overflow.
The suggested change also solves the visual overflow issue while removing rubber-banding. If this better fits the feel of the site, I’m happy to do that instead. Would you prefer that approach?

karsa-mistmere: I understand, but rubber banding isn't fully prevented this way, and some areas of the scrollbar are hidden and non-interactive, so it's probably not the best solution.

I feel like this issue should be solved directly in VitePress, not on lucide.dev:

https://github.com/user-attachments/assets/0bfcf938-79fb-4d96-a357-3af01e7467ec

karsa-mistmere: Closing this, please submit a bug report / PR directly to https://github.com/vuejs/vitepress

PR #4633, opened 2026-08-03
feat(dependabot): added dependabot config, grouping PRs for each package

karsa-mistmere:
## Description

Added a dependabot config that groups dependency updates by pnpm workspace package, reducing PR noise and making routine updates — especially batches of security releases like #4632, #4631, #4630 and #4629 — much more manageable.

## Before Submitting 

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,182 @@
+version: 2
+
+updates:
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: weekly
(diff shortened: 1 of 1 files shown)


jguddas (approved):

ericfennis (approved):

ericfennis:
Hmm, this is not working as expected. 
We can't, unfortunately, group by package since the package lock file is not located in each directory.

Reverting the grouping in #4665

PR #4639, opened 2026-08-04
fix(icons): changed `barrel` icon

karsa-mistmere:
- [x] Other: Icon update

### Description
Upgraded barrel to proportions more in line with a classic French barrique.

## Before Submitting


jguddas (changes requested):
I don't think this looks like a barrel, also feels less lucid-y, the roundness matches square/circle more.

karsa-mistmere:
> I don't think this looks like a barrel, also feels less lucid-y, the roundness matches square/circle more.

Please Google barrique.

jguddas:
Might not be realistic proportions, but I still think it's super recognizable this way and I prefer the more cartoonish style we have right now.

image

PR #4640, opened 2026-08-05
ci(workflows): drop CVE-affected changed-files v41, pin actions, scope tokens and secrets

Totara-thib:
## Description

Drive-by CI hardening in five commits, one logical change each, found with the Plumber CLI (https://github.com/getplumber/plumber) and verified on main.

**Commit 1 moves `request-review.yml` off `tj-actions/changed-files@v41`** (bumped to v46.0.5, pinned by sha). v41 sits inside the range hit by the March 2025 supply chain compromise of that action (GHSA-mrrh-fwg8-r2c3 / CVE-2025-30066): the attacker re-pointed existing version tags at code that dumped runner memory, including the job token, into public logs. The action's file-matching API is unchanged for the patterns used here; the other two call sites were already on v46 and only get the pin.

**Commit 2 pins every remaining action to its commit sha**, versions kept as comments. A tag is a movable pointer, and commit 1 is the proof it gets abused in practice, against this very repo's dependency set. The paths that matter here: `release.yml` publishes the whole lucide npm family via OIDC trusted publishing, `ci.yml` cuts the GitHub releases, and the comment and review workflows hold pull-request write tokens. Every sha was resolved from the upstream repo and cross checked against its release tag (softprops v1 has no patch tag, its comment names the floating tag it replaced). The pins stay maintainable: the repo has no `dependabot.yml` yet, a config with the `github-actions` ecosystem keeps the pins bumped with the comments in sync, happy to add it here if you want.

**Commit 3 replaces the two `secrets: inherit` calls into `release.yml` with an explicit pass.** The only secret the release path reads is `BLOB_READ_WRITE_TOKEN` (build-font's blob upload); npm publishing is OIDC and needs no token secret. The workflow files now show exactly which secret flows into the release path, and nothing else does.

**Commit 4 declares least-privilege permissions on the 16 workflows without a block**, derived from actual token use: the fourteen package build/test workflows and the code linter drop to `contents: read`; `lint-pr-title.yml` gets the `pull-requests: read` its action documents; `comment-icon-preview.yml`, which runs on the privileged `workflow_run` trigger processing artifacts from fork PRs, gets exactly `actions: read` + `issues: write` + `pull-requests: write` + `contents: read`. The workflows that already declare scoped permissions (labeler, stale, linting-icons, release and friends, nicely done) are untouched.

**Commit 5 stops the release-creating job from restoring the shared pnpm cache.** Actions caches are writable from other workflow runs; the job that cuts releases should install from the registry, and the frozen lockfile still pins and verifies every dependency.

One heads up: if the org uses an Actions allowlist in settings, patterns written against tags (like `owner/action@v1`) stop matching once refs are shas and workflows refuse to start. Entries need to be `owner/action@*` in that case.

I will also open a PR which adds the tool to CI so this does not quietly drift back; that one is a bonus, this PR stands on its own.

## Before Submitting


ericfennis:
Fixed in #4619 and #4668

ericfennis:
But thanks for the explanation and for working on this!

PR #4674, opened 2026-08-06
fix(scripts): removed toBeRemovedInVersion from all scripts and tools

karsa-mistmere:
We no longer use this prop, but some scripts & tools still add it / depend upon it.

## Before Submitting 


ericfennis (approved):

ericfennis:
One lint error!

karsa-mistmere:
> One lint error!

@ericfennis Fixed already, just pushed it. :)

PR #4746, opened 2026-08-24
feat(docs): added better contribution guide

karsa-mistmere:
## Before Submitting 


jguddas:
### Looking at this, I didn't know what is what.

image

### Maybe do add a `Do` one for this:

image

### Maybe use 4 different icons here:

image

### That's not our barrel is it?

image

### I feel like this needs more clear visual representation of why it's bad:

image

### Would make the blue stuff green here + the end result could also be green.

image

### Don't feel like if I were to ask people "what does this mean", they would say "don't change the representation of base shapes".

image

#### Same for:

image

### I would like to see a section to the visual weight one added for why we put the `plus` in the bottom left for `file`, but not for most other icons.

image

karsa-mistmere:
> ### Looking at this, I didn't know what is what.
> image

That surprises me a bit, since during our last meeting you specifically asked for there to be not just a design language guide, but a separate specification as well. Is there something in particular about the current presentation that makes the distinction unclear?

Edit: I've now updated these descriptions to hopefully make the distinction clearer. 

> ### Maybe do add a `Do` one for this:
> image

Good idea. These “Don’t”s were carried over from the previous guide (and added by you, if I’m not mistaken), and I’m struggling to come up with matching “Do”s that aren’t simply redundant restatements of the rule. Do you maybe have concrete examples in mind?

> ### Maybe use 4 different icons here:
> image

Wouldn’t that somewhat defeat the purpose of asking users to compare these shapes specifically against `square` and `circle`? The repeated icons are intentional here so the difference in proportions is easier to compare directly.

> ### That's not our barrel is it?
> image

[It should be](https://github.com/lucide-icons/lucide/pull/4639). I used the proposed version here because it looks more recognisably like a barrel than the current squished one does.

> ### I feel like this needs more clear visual representation of why it's bad:
> image

Agreed that it could be clearer. Do you have a concrete idea for how you’d visualise the problem? I’m not immediately seeing a way to make the issue more explicit.

> ### Would make the blue stuff green here + the end result could also be green.
> image

Noted, but I definitely would not. I think that would draw too much attention to the intermediate elements and reduce the emphasis on the end result.

> ### Don't feel like if I were to ask people "what does this mean", they would say "don't change the representation of base shapes".

I disagree, and I think the current example communicates this more clearly than the alternative you proposed on Discord (which, again, with too much red in the rules would reduce emphasis on the end result).

That being said, I don’t think the visual needs to communicate the entire rule in isolation to begin this — this is why each example also has accompanying text explicitly stating what should and shouldn’t be done.

> ### I would like to see a section to the visual weight one added for why we put the `plus` in the bottom left for `file`, but not for most other icons.

I don't think this is really a visual-weight issue, but primarily a consequence of the geometry of `file`: there’s much more usable space in the bottom-left corner, which lets us place modifiers there consistently, whereas the bottom-right corner conflicts with the fold for some modifiers.

jguddas:
> Wouldn’t that somewhat defeat the purpose of asking users to compare these shapes specifically against square and circle? The repeated icons are intentional here so the difference in proportions is easier to compare directly.

Maybe only make it 3.

> Agreed that it could be clearer. Do you have a concrete idea for how you’d visualise the problem? I’m not immediately seeing a way to make the issue more explicit.

Good

image

Bad

image

> https://github.com/lucide-icons/lucide/pull/4639. I used the proposed version here because it looks more recognisably like a barrel than the current squished one does.

Don't like the barrel proposal, prefer the more round one.

image

>  Noted, but I definitely would not. I think that would draw too much attention to the intermediate elements and reduce the emphasis on the end result.

Maybe just use gray / white.

> I don't think this is really a visual-weight issue

Maybe use `folder` instead, so it's a more normal case.

I do think most variants fit in the bottom right, but they don't really look great there, due to the visual weight.

image

karsa-mistmere:
> > Wouldn’t that somewhat defeat the purpose of asking users to compare these shapes specifically against square and circle? The repeated icons are intentional here so the difference in proportions is easier to compare directly.
> 
> Maybe only make it .

?

> > Agreed that it could be clearer. Do you have a concrete idea for how you’d visualise the problem? I’m not immediately seeing a way to make the issue more explicit.
> 
> Good
> 
> image
> Bad
> 
> image

I'll try to do a riff on this for the subelement variant, but your initial comment was on the cloud one. 👀

> > #4639. I used the proposed version here because it looks more recognisably like a barrel than the current squished one does.
> 
> Don't like the barrel proposal, prefer the more round one.
> 
> image

Again, please Google French barrique, not Asterix and Obelix.

> > Noted, but I definitely would not. I think that would draw too much attention to the intermediate elements and reduce the emphasis on the end result.
> 
> Maybe just use gray / white.

Ok, I'll restore the original greyscale version.

jguddas:
> I'll try to do a riff on this for the subelement variant, but your initial comment was on the cloud one. 👀

You don't have to use the cloud tho, you can use other shapes where it's more clear when it is sharp and when it is not.

jguddas:
> Ok, I'll restore the original greyscale version.

You can still have it duotone, make the user gray + the lock white.

karsa-mistmere:
> > I'll try to do a riff on this for the subelement variant, but your initial comment was on the cloud one. 👀
> 
> You don't have to use the cloud tho, you can use other shapes where it's more clear when it is sharp and when it is not.

True, but even for one like that, I would be important to highlight common pitfalls. These include not just unaligned vertical and horizontal lines which are super easy to illustrate, but unaligned arcs and inconsistent pixel grid snapping.

karsa-mistmere:
> Maybe use `folder` instead, so it's a more normal case.

@jguddas: Which is exactly why we **shouldn’t** use `folder`, and why `file` is a perfect candidate. After all, the rule **_is_** about following the conventions of the icon set, and while the placement of modifiers in `file-*` icons may not be typical, that makes it an especially good example: people should follow the precedent set by existing icons, even if they don’t immediately understand why things are placed the way they are.

This is especially apparent given that even you and I don’t fully see eye to eye on why exactly `file` icons are the way they are.

---

I've now updated most rules and visuals we've talked about. Let me know if you have any more bones to pick.

jguddas:
> > Maybe use `folder` instead, so it's a more normal case.
> 
> @jguddas: Which is exactly why we **shouldn’t** use `folder`, and why `file` is a perfect candidate. After all, the rule **_is_** about following the conventions of the icon set, and while the placement of modifiers in `file-*` icons may not be typical, that makes it an especially good example: people should follow the precedent set by existing icons, even if they don’t immediately understand why things are placed the way they are.
> 
> This is especially apparent given that even you and I don’t fully see eye to eye on why exactly `file` icons are the way they are.
> 
> ---
> 
> I've now updated most rules and visuals we've talked about. Let me know if you have any more bones to pick.
> 

Usually I'm against this inter group consistency, and want more of a global consistency, for sub shape placement I do think it make sense to chose group consistency over global consistency.

Could be two rules or two don't cards or something.

jguddas (approved):

ericfennis (approved):
This is so lit @karsa-mistmere 🔥 . Thanks so much!

PR #4748, opened 2026-08-25
feat(docs): add list view with icon names to the icons page

luizgununes:
## Description

The icons page only renders a grid of glyphs. Icon names are visible one at a time, on hover or in the detail drawer, and aliases are not visible at all, two names for the same icon look like two unrelated results.

This adds a **Grid / List** toggle to the sticky bar, next to the sort select. In list view each row shows the icon, its canonical name, its aliases, and a marker for deprecated icons. The chosen view is persisted in `localStorage`.

It also adds a button that **copies the names of the current result set** to the clipboard, respecting the active search and sorting.

**Why this helps**

- Aliases become visible at a glance. `help-circle` / `circle-help`, `alert-circle` / `circle-alert` and `alert-triangle` / `triangle-alert` are the same icons under two names, and a codebase can accumulate both spellings for a long time without anyone noticing.
- Auditing which icons a project uses currently means reading names one by one.
- Getting names into other tools, such as an allowlist, a design doc, or a framework registration array, currently means manual transcription. Angular, for example, requires importing and registering each icon explicitly.

**Implementation notes**

- `IconList.vue` and `IconListItem.vue` are new and mirror the props of `IconGrid` / `IconItem`, so the overview switches between them with `<component :is>`.
- `useVirtualList` now receives `itemHeight` as a function instead of a fixed number, and the chunk size follows the active view. Both `totalHeight` and `offsetTop` are computed from that function, so switching views recalculates the scroll geometry. With the previous fixed number, list rows would have been measured with the grid's row height.
- Grid view is untouched and remains the default.

## Before Submitting

### DIFF
--- a/docs/.vitepress/theme/components/icons/IconList.vue
+++ b/docs/.vitepress/theme/components/icons/IconList.vue
@@ -1,0 +1,52 @@
+<script setup lang="ts">
+import type { IconEntity } from '../../types';
+import IconListItem from './IconListItem.vue';
+
+const emit = defineEmits(['setActiveIcon']);
+
+defineProps<{
+  icons: IconEntity[];
+  activeIcon?: string;
+  overlayMode?: boolean;
+}>();
+
+function setActiveIcon(name: string) {
+  emit('setActiveIcon', name);
+}
+</script>
+
+<template>
+  <div class="icon-list">
+    <div
+      class="icon-list-row"
+      v-for="icon in icons"
+      :key="icon.name"
+    >
+      <IconListItem
+        :iconNode="icon.iconNode"
+        :name="icon.name"
+        :aliases="icon.aliases"
+        :deprecated="icon.deprecated"
+        :externalLibrary="icon.externalLibrary"
+        :active="activeIcon === icon.name"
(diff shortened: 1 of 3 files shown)


jguddas:
Not sure this is needed

karsa-mistmere:
I agree with @jguddas here — I’m not convinced this adds enough value to justify the additional UI and complexity.

The copy button in particular feels like a hard no to me, especially since the names shown on the website aren’t necessarily directly usable in most scenarios that our users face (i.e. our React and Angular users, since our packages use `PascalCase` imports for icons).

I do think making aliases more apparent could be useful, but I’m not sure a completely separate list view is the right solution. If searching for help-circle currently makes circle-help look like an unrelated result, I’d rather solve that directly in the search/result UI — e.g. by indicating that the search term is an alias of the canonical icon.

The auditing use case also doesn’t quite convince me: the icons page doesn’t know which icons a project actually uses, so list view would only make an arbitrary result set easier to read rather than really providing an auditing workflow.

Overall, I’d prefer addressing alias discoverability independently rather than adding another view mode and export functionality.

PR #4760, opened 2026-08-28
fix(icons): correct "operater" typo to "operator" in equal-approximately

hovelopin:
## Description

This PR fixes a typo in the `tags` for the `equal-approximately` icon. The tag `"operater"` was corrected to `"operator"`.

Related math/operator icons (`equal`, `equal-not`, `divide`, `minus`, etc.) already use the `"operator"` tag. `"operater"` appears only in `equal-approximately.json`.

## Before Submitting

### DIFF
--- a/icons/equal-approximately.json
+++ b/icons/equal-approximately.json
@@ -8,7 +8,7 @@
     "about",
     "calculate",
     "math",
-    "operater"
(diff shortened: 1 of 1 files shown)


sevineleven:
so nice PR

PR #4812, opened 2026-09-07
fix(icons): changed `ruler` icon

jguddas:
- [x] Other: Icon update

### Description
flipped icon

## Before Submitting


karsa-mistmere (changes requested):
A hard no from me, since I want this to be consistent with 📏

Unicode 6.0 defined this as:

> A ruler, as used to draw straight lines and measure distances. Variously depicted as a metal, wood, or plastic straightedge with length markings at a 45° angle.
> 
> Commonly used for content concerning schooling, building, and measurement in various disciplines. May also represent various concepts of straightness and size.

PR #4815, opened 2026-09-07
fix(icons): changed `cookie` icon

karsa-mistmere:
- [x] Other: Icon update

### Description
Arcified cookie icon, snapped dots to grid & added crumbs

## Before Submitting


ericfennis (approved):
Ah you are this kind of person. I knew it 😛

jguddas (approved):
Better than before, but I think we can still make it even better, doesn't have to be in this PR tho.

Issue #3354, opened 2025-06-22
Add individual entry points.


TheColorRed:
### Package

- [ ] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [x] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] lucide-astro
- [ ] Figma plugin
- [ ] all JS packages
- [ ] site

### Description

Exporting all icons in the barrel file can slow down build times. They can negatively impact performance, especially in larger projects, by increasing build times. This is because bundlers need to parse and resolve all the modules exported by the barrel file, even if only a small portion is actually used. 

When a bundler encounters a barrel file, it needs to parse all the files it re-exports to determine their contents, including potential side effects. This process can become time-consuming, especially with many barrel files or large modules.

### Use cases

The barrel file exports 1000+ icons (which tree-shaking does occur with a good bundler), however build times could possibly be speed up if the files were imported individually as well as with a barrel file (for compatibility).

example:

```tsx
import { XIcon } from 'lucide-react/icons/x'

return <XIcon />;
```

### Checklist

- [x] I have searched the existing issues to make sure this bug has not already been reported.

TheColorRed:
I think a fix might be as simple as adding an exports to the package.json

```js
{
  "exports": {
    ".": "./dist/esm/lucide-react.js",
    "./icons/*": "./dist/esm/icons/*.js"
  }
}
```

jguddas:
We tried and failed to do this here:
https://github.com/lucide-icons/lucide/pull/2814/files#diff-87a1e80768356231d72187ad57dcac32eebebaf49098c9d4522d36c000dca5beL48-L54

ericfennis:
Yess, @jguddas said. We tried this.
Unfortunately, in worker environments, this caused weird file memory issues. Builds from sites in Vercel, Cloudflare and Netlify broke. See thread: https://github.com/lucide-icons/lucide/issues/2743

So for now, we can't do this. 

You can make this possible by configuring your bundler, for example, with Vite:
```ts
resolve: {
  alias: {
    "lucide-react/icons": fileURLToPath(
      new URL("./node_modules/lucide-react/dist/esm/icons", import.meta.url)
    ),
  },
},
```
Declaration File:
```
declare module "lucide-react/icons/*" {
[...]

Issue #2743, opened 2025-01-17
[react] 0.470.0+ crashing deployments


rub1e:
### Package

- [ ] lucide
- [ ] lucide-angular
- [ ] lucide-flutter
- [ ] lucide-preact
- [x] lucide-react
- [ ] lucide-react-native
- [ ] lucide-solid
- [ ] lucide-svelte
- [ ] lucide-vue
- [ ] lucide-vue-next
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

0.470.0+

### Can you reproduce this in the latest version?

- [x] Yes
- [ ] No

### Browser

- [ ] Chrome/Chromium
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [x] Other/not relevant

### Operating system

- [ ] Windows
- [ ] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [x] Other/not relevant

### Description

**NOTE**: if this is affecting anyone currently, downgrading to 0.469.0 solves it for me

From 0.470.0 using lucide-react icons breaks my `react-router v7` deployments.

No problems building locally and serving with `react-router serve`, and no problem running `react-router build` for the deployment - no errors or warnings. 

But visiting the page seems to crash the Netlify lambda itself (also tested on Vercel). Here is the error log from Netlify:

> Jan 17, 01:20:06 PM: ERROR  Uncaught Exception 	{"errorType":"Error","errorMessage":"EMFILE: too many open files, open '/var/task/node_modules/lucide-react/dist/esm/icons/message-circle.js'","code":"EMFILE","errno":-24,"syscall":"open","path":"/var/task/node_modules/lucide-react/dist/esm/icons/message-circle.js","stack":["Error: EMFILE: too many open files, open '/var/task/node_modules/lucide-react/dist/esm/icons/message-circle.js'"]}

Also note that this isn't a react v19 issue, it's happening on 18.3 as well

### Steps to reproduce

Here's a minimal repro: https://github.com/rub1e/lucide-react-repro

It's just the Netlify React Router starter with one page with a single lucide icon

Here it is crashing on netlify: https://venerable-sunshine-738964.netlify.app/

### Checklist

- [x] I have searched if someone has submitted a similar issue before and there weren't any. (Please make sure to also search closed issues, as this issue might already have been resolved.)

olivier-thatch:
[...]

PR #4817, opened 2026-09-08
fix(lucide-react): add types for per-icon imports

huuyafwww:
## Description

Thanks for maintaining Lucide — we use it in an SSR app and wanted to avoid loading the whole icon barrel when only a few icons are needed.

The existing per-icon files work at runtime, but TypeScript reports TS7016 for this import:

```tsx
import Camera from 'lucide-react/dist/esm/icons/camera.mjs';
```

This adds a wildcard `typesVersions` entry for those `.mjs` files and generates a shared default-icon declaration from `src/types.ts`. It also publishes a lightweight `lucide-react/dist/lucide-types` entry, with an explicit mapping for NodeNext, so importing `LucideProps` does not pull in the full icon declaration bundle. Both outputs use the existing Rollup declaration settings, including the shared types.

I saw the concerns in #3354 and the rollback in #2814. This only adds type declarations: `main`, `module`, existing entry points and the emitted JavaScript stay unchanged.

The shared declaration covers the default component export. It does not include per-icon deprecation metadata or the named `__iconNode` / `__iconData` exports. A misspelled icon path still typechecks with the wildcard; the bundler reports the missing file.

In a minimal packed-package consumer, the Lucide declarations loaded by TypeScript dropped from 2.20 MB to 2.15 KB. The Vite fixture covers a canonical icon and an alias; its typecheck fails before the change and passes here. I also checked strict Bundler/NodeNext consumers and confirmed that the existing 4,184 JavaScript and sourcemap files match byte-for-byte when built from the same generated icon inputs.

The package build and lint pass, as do build, typecheck and Chromium tests in all five integration fixtures. The 33 unit tests pass with two workers. An initial run alongside other builds timed out in a dynamic-import test, and React Router’s first browser run hit a connection timeout; both passed when rerun separately.

Would you be open to this approach? Happy to adjust the shape if you would prefer a different way to expose these types.

## Before Submitting

### DIFF
--- a/integrations/lucide-react/vite/src/IconShowcase.tsx
+++ b/integrations/lucide-react/vite/src/IconShowcase.tsx
@@ -1,4 +1,6 @@
-import { Camera, Droplet, Edit2, House, LucideProvider, Pen } from 'lucide-react';
+import { Camera, House, LucideProvider, Pen } from 'lucide-react';
+import Droplet from 'lucide-react/dist/esm/icons/droplet.mjs';
+import Edit2 from 'lucide-react/dist/esm/icons/edit-2.mjs';
 import { DynamicIcon } from 'lucide-react/dynamic';
 
 export default function IconShowcase() {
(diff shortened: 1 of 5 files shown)


huuyafwww:
Closing this in favor of a broader proposal that also adds an `exports` map with `./icons/*` subpaths while keeping Node's root resolution on the CJS bundle (so the #2743 failure mode cannot recur). I will open it as a separate PR.

PR #4825, opened 2026-09-08
feat(docs): fixed fuse js search

karsa-mistmere:
## Motivation

1. The search query "user pen" should return [`user-round-pen`](https://lucide.dev/icons/user-round-pen), but it doesn't:
   image

2. The "no results" placeholders are missing:
   image

## Description

1. Fixed Fuse.js search (moved from a single phrase search to matching all words)

2. Fixed "no results" behaviour by fixing the computed `isSearchMetadataLoading`

## Before Submitting 

### DIFF
--- a/docs/.vitepress/theme/composables/useSearch.ts
+++ b/docs/.vitepress/theme/composables/useSearch.ts
@@ -9,6 +9,7 @@ const useSearch = <T>(
   const index = shallowRef(
     new Fuse(collection.value, {
       threshold: 0.2,
+      useExtendedSearch: true,
       keys,
     }),
   );
@@ -17,7 +18,9 @@ const useSearch = <T>(
     index.value.setCollection(collection.value);
 
     if (query.value) {
-      return index.value.search(query.value).map((result) => result.item);
+      return index.value
+        .search({ $and: query.value.split(' ').filter((t) => !!t) })
+        .map((result) => result.item);
     }
 
     return collection.value;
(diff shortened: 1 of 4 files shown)


jguddas (approved):

ericfennis (dismissed):

docs/.vitepress/theme/composables/useSearch.ts
```diff
@@ -17,7 +18,9 @@ const useSearch = <T>(
     index.value.setCollection(collection.value);
 
     if (query.value) {
-      return index.value.search(query.value).map((result) => result.item);
+      return index.value
+        .search({ $and: query.value.split(' ').filter((t) => !!t) })
```
ericfennis, line 22: I think we should also split on dashes
Image

```suggestion
        .search({ $and: .split(/[\s-]+/).filter(Boolean) })
```

karsa-mistmere: I deliberately kept this as is so that we can search for specific exact matches.

karsa-mistmere: E.g. https://lucide-8oi77pfhk-lucide.vercel.app/icons/?focus=&search=arrow-up
 vs. https://lucide-8oi77pfhk-lucide.vercel.app/icons/?focus=&search=arrow+up

ericfennis: Hmm that makes sense

ericfennis (approved):

PR #4829, opened 2026-09-09
feat(docs): added "How to use Lucide icons" section to resources

karsa-mistmere:
## Before Submitting 


jguddas (approved):

ericfennis (changes requested):
I'm not sure about this part: 
image

It's confuses a bit. The "don't" visual in the first one is actually a "Do" situation.

And below we should have an extra "Do" and "Don't" what more support the text. Because the "Do" have icons are actually the icons that don't need a label. So I think we should split this in two situations:
- Use label if icons are not conventionally known
- Don't use labels if icons are conventionally known like: `House`, `settings`, `account`

But the for the rest. Nice work 🔥

karsa-mistmere:
> I'm not sure about this part ....

@ericfennis: changed this section to include some new "Caution" guideance items, I hope this is clearer.

ericfennis (approved):

PR #4846, opened 2026-09-11
fix(packages/svelte,solid): fix shared type imports in Solid and Svelte

### LINKED ISSUE #4842: lucide-solid and @lucide/svelte: published .d.ts imports from unpublished internal package @lucide/shared, breaking strict type-checking (Cannot find module '@lucide/shared/types')
### Package

- [ ] lucide
- [ ] lucide-react
- [ ] @lucide/vue
- [x] @lucide/svelte
- [ ] @lucide/angular
- [ ] @lucide/astro
- [ ] lucide-react-native
- [ ] lucide-preact
- [x] lucide-solid
- [ ] lucide-static
- [ ] @lucide/icons
- [ ] Figma plugin
- [ ] source/main
- [ ] other/not relevant

### Version

1.44.0 (also present in 1.42.0 and 1.43.0)

- [x] Yes
- [ ] No

### Browser

- [ ] Chrome/Chromium
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Opera
- [x] Other/not relevant

### Operating system

- [ ] Windows
- [ ] Linux
- [ ] macOS
- [ ] ChromeOS
- [ ] iOS
- [ ] Android
- [x] Other/not relevant

### Description

Since 1.42.0, the published type declarations for `lucide-solid` and `@lucide/svelte` contain an import from an internal monorepo-only package, `@lucide/shared`, which is never published to npm. Any consumer with strict module resolution (`skipLibCheck: false`) gets a `TS2307: Cannot find module` error.

`dist/types/types.d.ts` (lucide-solid) and `dist/types.d.ts` (@lucide/svelte) both contain:

    import type { LucideIconData as SharedLucideIconData, LucideIconNode as SharedLucideIconNode } from '@lucide/shared/types';

This file is re-exported from each package's public entry point:
- `lucide-solid`: `dist/types/lucide-solid.d.ts` → `export * from './types';`
- `@lucide/svelte`: `dist/lucide-svelte.d.ts` → `export * from './types.js';`

So the broken import is reachable from a plain `import { Check } from 'lucide-solid'` / `import { Check } from '@lucide/svelte'`.

`@lucide/shared` is not published on npm (`npm view @lucide/shared` → `404 Not Found`), and in both packages' `package.json` it only appears under `devDependencies` (an internal workspace package used at build time), never under `dependencies` or `peerDependencies`.

Comparing against 1.41.0 (last good release): neither package referenced `@lucide/shared` at all — both `types.d.ts` files were fully self-contained. The reference first appears in 1.42.0, the same release that introduced the `rollup-plugin-dts@^6.5.1` / `typescript@^6.0.3` bump referenced in #4806 (which also appears to be the root cause of #4837). It looks like the dts bundler is no longer inlining/rewriting cross-workspace-package type imports the way it did on 1.41.0.

**Why this is easy to miss:** `skipLibCheck: true` (the default in most Svelte/SolidJS project templates) suppresses errors inside `.d.ts` files entirely, so builds stay green while these types are effectively unresolved for any strict consumer.

Not affected: `@lucide/angular`, `@lucide/astro`, `lucide-preact`, `lucide-react`, `lucide-react-native` — none of their bundled declaration files reference `@lucide/shared`.

Suggested fix direction: either (1) have the dts build step inline/bundle the types from `@lucide/shared`, as was effectively happening pre-1.42.0, or (2) publish `@lucide/shared` to npm and move it to `dependencies`/`peerDependencies` in both packages.

### Steps to reproduce

1. Create a new project and install: npm i typescript@6.0.3 solid-js@1.9.15 lucide-solid@1.44.0 svelte@5 @lucide/svelte@1.44.0
2. Create probe-solid.ts: import { Check } from 'lucide-solid'; export const A = Check;
3. Create probe-svelte.ts: import { Check } from '@lucide/svelte'; export const B = Check;
4. Create tsconfig.json with: "strict": true, "module": "esnext", "moduleResolution": "bundler", "noEmit": true, "skipLibCheck": false, "include": ["*.ts"]
5. Run: npx tsc -p .

Observed:
node_modules/lucide-solid/dist/types/types.d.ts(2,101): error TS2307: Cannot find module '@lucide/shared/types' or its corresponding type declarations.
node_modules/@lucide/svelte/dist/types.d.ts(10,101): error TS2307: Cannot find module '@lucide/shared/types' or its corresponding type declarations.

Expected: no errors.

Pinning both packages to 1.41.0 and re-running the same repro produces no errors — types.d.ts in that version is fully self-contained with no reference to @lucide/shared.

### Checklist

karsa-mistmere:
closes #4842 

## Description

Fixes published `lucide-solid` and `@lucide/svelte` declarations so they no longer reference unpublished `@lucide/shared` types.

- Solid now bundles declaration types with `rollup-plugin-dts` similar to other packages.
- Svelte imports its packaged local utility types from `copy:utils`.

## Before Submitting 

### DIFF
--- a/packages/lucide-solid/package.json
+++ b/packages/lucide-solid/package.json
@@ -80,6 +80,7 @@
     "babel-preset-solid": "^1.9.15",
     "esbuild": "^0.28.2",
     "jest-serializer-html": "^7.1.0",
+    "rollup-plugin-dts": "^6.5.1",
     "rollup": "^4.63.1",
     "solid-js": "^1.9.15",
     "typescript": "^6.0.3",
--- a/packages/lucide-solid/rollup.config.mjs
+++ b/packages/lucide-solid/rollup.config.mjs
@@ -2,9 +2,14 @@ import path from 'path';
 import { babel } from '@rollup/plugin-babel';
 import esbuild from 'esbuild';
 import plugins from '@lucide/rollup-plugins';
-import ts from 'typescript';
+import dts from 'rollup-plugin-dts';
 
 import pkg from './package.json' with { type: 'json' };
+
+const dtsOptions = {
+  includeExternal: ['@lucide/shared/types'],
+  tsconfig: './tsconfig.json',
+};
 
 const packageName = 'LucideSolid';
 const outputFileName = 'lucide-solid';
@@ -92,21 +97,6 @@ const configs = bundles
                   ],
                   external: ['solid-js'],
                 });
-
-                // Generate types
-                const program = ts.createProgram([pkg.source], {
-                  target: ts.ScriptTarget.ESNext,
-                  module: ts.ModuleKind.ESNext,
-                  moduleResolution: ts.ModuleResolutionKind.NodeJs,
-                  jsx: ts.JsxEmit.Preserve,
-                  jsxImportSource: 'solid-js',
-                  allowSyntheticDefaultImports: true,
-                  esModuleInterop: true,
-                  declarationDir: `dist/types`,
-                  declaration: true,
-                  emitDeclarationOnly: true,
-                });
-                program.emit();
               },
             }
           : null,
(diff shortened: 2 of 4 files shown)


ericfennis (approved):
Looks good

PR #4848, opened 2026-09-11
fix(icons): remove path from save-off

HPRILLER:
## Description

Removed an invalid vector path from `save-off.svg`.

The path was located outside the icon's `24x24` viewBox and was clipped during rendering. Although it was not visible, it was invalid source data and should not be part of the icon.

## Before Submitting


HPRILLER:
The PR is ready for review. The lint-code workflow is currently waiting for maintainer approval.

karsa-mistmere (approved):
Nice catch, thanks!

PR #4855, opened 2026-09-13
feat(site): persist icon customizer settings across visits

### LINKED ISSUE #4358: Remember customizer settings on website

nusygako:
## What
Persist the icon customizer's size/strokeWidth/color/absoluteStrokeWidth settings across page reloads and visits.

## Why
Closes #4358. The customizer state was plain Vue `ref()`s that reset to `STYLE_DEFAULTS` on every load, forcing users to re-adjust settings each time before downloading an icon.

## How
Switched `iconStyleContext` in `useIconStyle.ts` to use `useStorage` from `@vueuse/core` instead of plain `ref()`. This matches the existing persistence pattern already used in `ButtonMenu.vue` for the same purpose, so no new dependency or custom pattern was introduced. `useStorage` is SSR-safe by design (falls back to the given default when `window`/`localStorage` isn't available, e.g. during VitePress's static build).

No other files needed changes — `SidebarIconCustomizer.vue`'s `resetStyle()` still works unmodified since assigning `.value` on a `useStorage` ref writes through to localStorage automatically.

## Checklist
- [x] No new dependency (`@vueuse/core` already used for this exact pattern)
- [x] Matches existing codebase convention rather than introducing a new one
- [x] No test added — the repo has no test infrastructure under `docs/` to match

🤖 Generated with [Claude Code](https://claude.com/claude-code)


karsa-mistmere:
Hey @nusygako,

I'm afraid this implementation suffers from the same issues as described in https://github.com/lucide-icons/lucide/pull/4393#pullrequestreview-4371064404

> 1. `absoluteStrokeWidth` does not appear to persist after refresh.
> 2. On page refresh, the default values are rendered first, then replaced by the `localStorage` values a moment later. This causes a very visible state jump/flicker, that I think we should try to avoid if possible.
> 
> https://github.com/user-attachments/assets/523b45e7-203d-422d-9e14-7f6adb265eb8

ericfennis:
@karsa-mistmere Good point. Closing this one

Issue #4733, opened 2026-08-22
`receipt-pakistani-rupee` and `badge-pakistani-rupee` icons


Muhammad-Aqib-Bashir:
### Icon name

`receipt-pakistani-rupee` `badge-pakistani-rupee` 

### Use cases

I would like to request Pakistani Rupee (PKR) variants for the existing `receipt-*` `and badge-*` currency icon families.

* **Point-of-sale (POS) systems:** I’m currently developing a POS application that uses Pakistani Rupees for product prices, receipts, invoices, checkout screens, transaction records, and payment summaries.

**Some Other:**
* **Retail and accounting applications:** Identify PKR amounts for sales, refunds, discounts, taxes, expenses, and cash management.
* **E-commerce platforms:** Clearly indicate Pakistani Rupee prices, order totals, invoices, and payment transactions for customers in Pakistan.
* **Financial dashboards:** Represent PKR revenue, balances, expenses, and other monetary metrics in a compact and recognizable way.
* **Multi-currency applications:** Provide a consistent Pakistani Rupee icon alongside existing currency icons such as `receipt-euro`, `receipt-indian-rupee`, `badge-euro`, and `badge-indian-rupee`.

### Design ideas

`receipt-pakistani-rupee` 

[...]

Issue #3518, opened 2025-08-07
Icon Request: badge-pakistani-rupee, receipt-pakistani-rupee


Muhammad-Aqib-Bashir:
### Icon name

receipt-pakistani-rupee, badge-pakistani-rupee

### Use cases

- Financial dashboards or leaderboards in Pakistan-based apps
- Coupons, vouchers, or promotional incentives displayed in PKR

### Design ideas

`receipt-pakistani-rupee,`:

[...]

PR #4874, opened 2026-09-16
feat(icons): add pakistani-rupee icon

hahmed-cpu:
## New icon: pakistani-rupee

Adds a currency icon for the Pakistani Rupee (PKR), symbol "Rs".

Lucide already has `indian-rupee` and `nepali-rupee`, both based on the Devanagari-derived rupee glyph. 
Pakistan's rupee symbol is written in Latin script as "Rs" (per ISO 4217 / common usage), so this icon is drawn as a capital R + lowercase s to match, keeping it visually distinct from the existing rupee icons.

- Tags: currency, money, pakistan, rupee, pkr
- Categories: finance, shopping, travel


karsa-mistmere:
Duplicate of #3518 and #4733

